PR #1, opened 2017-03-08
Aakarshit

aksh1618:
Kindly review and suggest improvements @kaustubh-karkare


solutions/aakarshit-uppal/assignment_1
```diff
@@ -0,0 +1,50 @@
...
+import argparse  # For ArgumentParser
+import json  # For dumps - converts to JSON String
+
+# Create Parser
+parser = argparse.ArgumentParser(description='Parse Command Line to JSON')
+
+# Add Arguments
+parser.add_argument('command', nargs='?', type=str, help='Main Command')
+parser.add_argument(
+    'subcommand', nargs='?', type=str, help='Secondary Command')
```
kaustubh-karkare: One argument per line makes things a lot easier to read.

kaustubh-karkare:
Thanks for pointing out the issues! I noticed some of these myself, and was about to push my own commit here, and so I fixed the mistakes I didn't know about. Your commits could have been separate pull requests, since I couldn't accept one without the other.

solutions/aakarshit-uppal/assignment_1
```diff
@@ -0,0 +1,50 @@
...
+
+Supports a command and a subcommand (both optional), a required integer
+argument 'key', an optional string argument 'name', a verbose argument and two
+mutually exclusive arguments, local and remote, which set 'local' to true or
+false, respectively.
+
+Any arguments not used are not included in JSON.
+"""
+
+import argparse  # For ArgumentParser
```
kaustubh-karkare: I updated the problem statement to say that you cannot use an existing library like this, since it defeats the purpose of the exercise. So you'll have to rewrite this. Sorry!

aksh1618: Agreed (I should have thought of that). I'll get on it.

solutions/aakarshit-uppal/assignment_1
aksh1618: Thanks for the tip!

kaustubh-karkare: The rule I follow is: entire function call on one line, or every argument in its own line.

aksh1618:

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,280 @@
...
+    solution.add_arg('--local', set_value=True)
+    solution.add_arg('--remote', set_value=True)
+    solution.make_args_exclusive('--local', '--remote')
+    solution.parse('alpha beta --keys 1 2 -V')
+    solution.print_json()
+Output::
+    {"command": "alpha", "subcommand": "beta", "keys": [1, 2], "verbose": true}
+"""
+import sys as _sys  # For argv (to read arguments from command line)
+import json as _json  # For dumps (to convert dict to JSON)
```
kaustubh-karkare: Why do you need to rename them like this?

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,280 @@
...
+_positional_args = []
+_optional_args = []
+_required_args = []
+_exclusive_groups = []
+_arg_values = _collections.OrderedDict()
+
+
+def _get_new_params_list():
+    # Returns initial params list
+    return {'alias': None,
```
kaustubh-karkare: Storing it as a map isn't particularly elegant. Use a class?

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,280 @@
...
+import sys as _sys  # For argv (to read arguments from command line)
+import json as _json  # For dumps (to convert dict to JSON)
+import collections as _collections  # For OrderedDict (For ordered values)
+
+_arg_params = {}
+_positional_args = []
+_optional_args = []
+_required_args = []
+_exclusive_groups = []
+_arg_values = _collections.OrderedDict()
```
kaustubh-karkare: Module-level variables like this aren't the best design. What if I want to parse 2 separate lists of arguments in 2 parallel threads?

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,280 @@
...
+              'argument is required, but missing from input.')
+    if type_of_error is 'exclusive':
+        print('Error:', str(arg_names)[1:-1], 'cannot be used together.')
+    if type_of_error is 'no_such_arg':
+        print('Error: Invalid argument', arg_names[0])
+    if type_of_error is 'too_few_values':
+        print('Error: Too few values supplied for', arg_names[0])
+    if type_of_error is 'invalid_type':
+        print('Error: Invalid value type(s) for', arg_names[0])
+    if type_of_error is 'too_many_posargs':
```
kaustubh-karkare: Using raw strings like this is generally not a good idea, since it is hard to notice typos. Use Enums instead. Or, just use separate exception types.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,280 @@
...
+        if arg not in cli_args:
+            _handle_error('missing_required', arg)
+
+
+def _check_exclusive(cli_args):
+    # Checks if two or more exclusive arguments were used
+    for arg_group in _exclusive_groups:
+        common_args = set.intersection(set(arg_group), set(cli_args))
+        common_args = list(common_args)
+        common_args.sort()
```
kaustubh-karkare: Why do you need to convert to list and sort?

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,280 @@
...
+                params['value'] = params['set_value']
+            elif params['nvals'] is 1:
+                i += 1
+                try:
+                    params['value'] = params['type'](cli_args[i])
+                except IndexError:
+                    _handle_error('too_few_values', curr_arg)
+                except ValueError:
+                    _handle_error('invalid_type', curr_arg)
+            else:
```
kaustubh-karkare: Why not just combine this else with the previous if?

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,280 @@
...
+                l = []
+                for j in range(params['nvals']):
+                    try:
+                        l.append(params['type'](cli_args[i]))
+                        i += 1
+                    except IndexError:
+                        _handle_error('too_few_values', curr_arg)
+                    except ValueError:
+                        _handle_error('invalid_type', curr_arg)
+                i -= 1
```
kaustubh-karkare: It is inelegant how you decrement here, and increment before the loop. get rid of both?

solutions/aakarshit-uppal/solution_1_2/README.md
```diff
@@ -0,0 +1,30 @@
...
+    - `set_value` (to specify value to be set on argument use)
+- Multiple values can be passed for argument
+- Values passed on call can be accessed as ordered or simple dictionary in:
+    - Long form (including none-valued)
+    - Short form (excluding none-valued)
+- Values can be printed in JSON format
+
+[**solution_test.py**](./solution_test.py) : Tests for solution module.
+
+Testing functions:
```
kaustubh-karkare: Not sure if this section is really needed. This forces the developer to keep this list in sync with the actual code, which is a pain. That is why everyone prefers avoiding comments.

solutions/aakarshit-uppal/solution_1_2/README.md
```diff
@@ -0,0 +1,30 @@
...
+- Arguments can be added with specific paramters:
+    - `alias` (to add short name for argument)
+    - `required` (to make argument compulsory)
+    - `type` (to specify type of value for argument)
+    - `nvals` (to specify number of values for argument)
+    - `set_value` (to specify value to be set on argument use)
+- Multiple values can be passed for argument
+- Values passed on call can be accessed as ordered or simple dictionary in:
+    - Long form (including none-valued)
+    - Short form (excluding none-valued)
```
kaustubh-karkare: Not sure I understand the last 3 lines. Clarify language? Ask a friend to read, confirm that they understand, if you wish. Technical writing is a whole different skill-set, which is also worth developing.

solutions/aakarshit-uppal/solution_1_2/solution_test.py
```diff
@@ -0,0 +1,89 @@
...
+import solution
+
+
+class TestAddition(unittest.TestCase):
+
+    def test_duplicate_addition(self):
+        solution.add_arg('--first', set_value=True)
+        solution._handle_error = unittest.mock.Mock()
+        solution.add_arg('--first', set_value=True)
+        solution._handle_error.assert_called_once_with(
```
kaustubh-karkare: Yeah, this is non-standard. You want to use `assertRaises`

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: Because they are for private use in the module, seemed like a convention. (Isn't it?)

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: Would putting everything in a class solve this?

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: Will do.

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: Will do.

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: The test wasn't working, as mock function needed exact arguments passed to the error handler. I'd replace this with an exception.

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: So that in case of a single value, the value itself is assigned, instead of a list with a single value. (I should have used a comment)

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: Will do.

solutions/aakarshit-uppal/solution_1_2/README.md
aksh1618: Agreed. Will do.

solutions/aakarshit-uppal/solution_1_2/README.md
aksh1618: Sure. Should this be replaced with a short summary of tests, or just the first one-line description is enough?

solutions/aakarshit-uppal/solution_1_2/solution_test.py
aksh1618: I had no idea about that. (That would actually reduce so much work!)

solutions/aakarshit-uppal/solution_1_2/README.md
kaustubh-karkare: No need for description of tests. It is generally assumed that a reasonable engineer will write tests anyway.

solutions/aakarshit-uppal/solution_1_2/solution.py
kaustubh-karkare: Ah, okay. There's still duplicate code though. Avoid that?

solutions/aakarshit-uppal/solution_1_2/solution.py
kaustubh-karkare: Yes.

aksh1618:

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+        # Handle optargs
+        excl_optargs_to_skip = []
+        for optarg_name in self._optional_args:
+
+            # Prevent duplication of exclusive optargs
+            if optarg_name in excl_optargs_to_skip:
+                continue
+
+            optarg = self._args[optarg_name]
+            if not optarg.required:
```
kaustubh-karkare: This is a contradiction in terms. Which means that there is a flaw in naming.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+                other_excl_args = [x for x in excl_args if x != optarg_name]
+                for arg_name in other_excl_args:
+                    usage += '|' + arg_name
+                    if self._args[arg_name].alias is not None:
+                        usage += '/' + self._args[arg_name].alias
+                    excl_optargs_to_skip.append(arg_name)
+
+            # Handle optargs having nval > 0
+            no_of_values = optarg.nvals
+            if no_of_values:
```
kaustubh-karkare: Unnecessary.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+            usage += ' '
+
+        print(usage)
+
+    def parse(self, *cli_args):
+        """Parse command line arguments.
+
+        cli_args -- string containing arguments to parse
+        """
+        if not len(cli_args):
```
kaustubh-karkare: Empty should always only mean empty array. Not "assume default value".

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+            cli_args = _sys.argv
```
kaustubh-karkare: Which means that this should be provided as input.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+            cli_args.remove(cli_args[0])
```
kaustubh-karkare: The fact the the first argv is the name of the file is an abstraction leak. The parser class has no business filtering the input.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def get_ordered_non_none_values(self):
+        """Return ordered dictionary containing args with non-none values."""
+        non_none_values = _collections.OrderedDict(self._arg_values)
+        for arg in self._arg_values:
+            if self._arg_values[arg] is None:
+                del non_none_values[arg]
+        return _collections.OrderedDict(non_none_values)
+
+    def print_json(self):
```
kaustubh-karkare: No reason for this to be in the class.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def get_non_none_values(self):
+        """Return dictionary containing args with non-none values."""
+        non_none_values = dict(self._arg_values)
+        for arg in self._arg_values:
+            if self._arg_values[arg] is None:
+                del non_none_values[arg]
+        return dict(non_none_values)
+
+    def get_ordered_non_none_values(self):
```
kaustubh-karkare: Looks like these methods exist for the unit test? I'm not sure that's a good enough reason.
In order to keep things simple, I would discouraging having ordered versions at all.

solutions/aakarshit-uppal/solution_1_2/solution_test.py
```diff
@@ -0,0 +1,83 @@
...
+    def test_required_arg_not_used(self):
+        self.parser.add_arg('--key', type=int, required=True)
+        with self.assertRaises(solution.RequiredArgError):
+            self.parser.parse('alpha beta')
+
+    def test_multiple_exclusive_args_used(self):
+        self.parser.add_arg('--none', setval=True)
+        self.parser.add_arg('--single', setval=True)
+        self.parser.add_arg('--many', setval=True)
+        self.parser.make_args_exclusive('--none', '--single', '--many')
```
kaustubh-karkare: Tbh, this is a bit inelegant. Having to specify the argument details at more than one place is an anti pattern.

solutions/aakarshit-uppal/solution_1_2/README.md
```diff
@@ -0,0 +1,18 @@
+## Solution Description
+
+[**solution.py**](./solution.py) : Module facilitating parsing of command line arguments.
+
+Supports positional and optional arguments, and more features, like:
```
aksh1618: Agreed, will fix.

solutions/aakarshit-uppal/solution_1_2/README.md
```diff
@@ -0,0 +1,18 @@
...
+- Arguments can be added with specific attributes:
+    - `alias` (to add short name for argument)
+    - `required` (to make argument compulsory)
+    - `type` (to specify type of value for argument)
+    - `nvals` (to specify number of values for argument)
+    - `set_value` (to specify value to be set on argument use)
+- Multiple values can be passed for argument
```
aksh1618: My bad, will fix.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+    {"command": "alpha", "subcommand": "beta", "keys": [1, 2], "verbose": true}
+"""
+import sys as _sys  # For argv (to read arguments from command line)
+import json as _json  # For dumps (to convert dict to JSON)
+import collections as _collections  # For OrderedDict (For ordered values)
+
+
+class DuplicateArgError(Exception):
+    def __init__(self, arg_name):
+        error_message = 'The argument \'' + arg_name + '\' already exists.'
```
aksh1618: Will do.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def __init__(self, test=False):
+        """Initialize lists and dictionaries."""
+        self._args = {}
+        self._positional_args = []
+        self._optional_args = []
+        self._required_args = []
+        self._exclusive_groups = []
+        self._arg_values = _collections.OrderedDict()
+        self.test = test
```
aksh1618: Agreed. Will fix.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+            return self._args[arg_name]
+        else:
+            for arg in self._args:
+                if self._args[arg].alias == arg_name:
+                    return self._args[arg]
+        self._show_usage_and_raise(UnknownArgError(arg_name))
+
+    def _show_usage_and_raise(self, exception):
+        # Calls print_usage and raises exception
+        if not self.test:
```
aksh1618: To prevent printing of usage in tests, makes the test output look ugly. Should that not be a concern?

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+            self.print_usage()
+        # Uncomment to print single - line errors for exceptions
+        # _sys.excepthook = lambda exctype, exc, traceback: \
+        #     print("{}: {}".format(exctype.__name__, exc))
+        raise exception
+
+    def _check_required(self, cli_args):
```
aksh1618: Will do. (Feels like that's exactly what i subconsciously wanted to name them :P )

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+        # Checks if all required arguments were used
+        for arg in self._required_args:
+            if arg not in cli_args:
+                if self._args[arg].alias not in cli_args:
```
aksh1618: Will do. (and was leading to >79 chars before refactoring, not anymore)

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+                    self._show_usage_and_raise(RequiredArgError(arg))
+
+    def _check_exclusive(self, cli_args):
+        # Checks if two or more exclusive arguments were used
+        for arg_group in self._exclusive_groups:
+            common_args = set.intersection(set(arg_group), set(cli_args))
+            common_args = list(common_args)
+            common_args.sort()
```
aksh1618: I don't, this crept in from previous implementation, will fix.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+            curr_arg = cli_args[i]
+
+            if curr_arg[0] is '-':
+                arg = self._get_arg(curr_arg)
+                if arg.nvals is 0:
+                    arg.value = arg.setval
+                else:
+                    try:
+                        l = []
+                        for j in range(arg.nvals):
```
aksh1618: Will do.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+    def _put_values(self):
+        # Populates self._arg_values with argument names and corresponding
+        # values
+        for posarg_name in self._positional_args:
+            self._arg_values[posarg_name] = self._args[posarg_name].value
+        for optarg_name in self._optional_args:
+            clean_arg = optarg_name[2:]
+            self._arg_values[clean_arg] = self._args[optarg_name].value
+
+    def add_arg(self, name, *alias,
```
aksh1618: I was at one point thinking about implementing multiple aliases, but a default argument would do for a single alias. Will fix.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+        attributes -- one or more of:
+            required=(boolean)
+            type=(python type)
+            nvals=(number of values)
+            setval=(value to set)
+        """
+        if name in self._positional_args or name in self._optional_args:
+            raise DuplicateArgError(name)
+
+        new_arg = _Argument()
```
aksh1618: Will do.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+
+        if len(alias):
+            new_arg.alias = alias[0]
+        new_arg.required = required
+        new_arg.type_ = type
+        new_arg.nvals = nvals
+        new_arg.setval = setval
+        if new_arg.setval is not None:
+            new_arg.nvals = 0
+            new_arg.type_ = __builtins__['type'](new_arg.setval)
```
aksh1618: No, as one of the attributes is type (Allows the user to be able to specify argument type as "type = str" while calling add_arg)

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: This facilitates printing the type of value required by the argument while parsing, useful especially when invalid value type error is raised.

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: Will fix.

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: Will do.

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: Agreed, will do.

solutions/aakarshit-uppal/solution_1_2/solution.py
aksh1618: Ordered versions used by tests anymore. Will remove.

solutions/aakarshit-uppal/solution_1_2/solution.py
kaustubh-karkare: Hmm. I think the right model is that you never print exceptions, only raise them. Then it is upto the calling method to figure out how to handle them (print / write to file / retry / etc).

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,347 @@
...
+
+        if name[0] is '-':
+            self._optional_args.append(name)
+        else:
+            self._positional_args.append(name)
+        if new_arg.required is True:
+            self._required_args.append(name)
+        self._args[name] = new_arg
```
kaustubh-karkare: Basically, you have `self._args`, `self._required_args`, `self._optional_args` and `self._positional_args`, which is unnecessary complexity.
Imagine if you just saved `self._args`, then derive the remaining by filtering the key-values pairs.

solutions/aakarshit-uppal/solution_1_2/solution.py
kaustubh-karkare: `if not optarg.required:` Translation: If this optional argument is not required ...
I would expect that the answer is always yes, since optional by definition is not required. And so the condition is intuitively equivalent to "if True". But that isn't the case - this condition isn't intuitive, as a consequence of improper naming.

solutions/aakarshit-uppal/solution_1_2/solution.py
kaustubh-karkare: You can run code using "./solution.py", "python solution.py" or even "time python solution.py" (not sure about the last one) or through unit tests, which can all lead to different argv. And the decision to remove 0 or 1 or 2 arguments is based on other factors, which are not related to parsing the remaining arguments. It isn't the responsibility of this class to do that. It is the responsibility of the caller.

solutions/aakarshit-uppal/solution_1_2/solution_test.py
kaustubh-karkare: Since you're already familiar with python's argparse, note how argument groups work. You only have to specify an argument once. In this case, however, you're forced to specify details about "--single" at more than one place (both in the `add_arg` method, and the `make_args_exclusive` method).

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,263 @@
...
+from solution_exceptions import *  # For Exceptions
+
+
+class _Argument:
+    # Defines attributes for arguments.
+    def __init__(self, alias,
+                 required=False,
+                 type=None,
+                 nvals=1,
+                 setval=None):
```
kaustubh-karkare: Stylistically, I prefer:
```
def __init__(
  self,
  alias,
  ...
):
```
If the entire function call/signature cant be on one line, each element should be in it's own line. Additionally, what I showed is how most other languages do it, and I prefer consistency with them, instead of treating python specially.

Also, it's funny how this argument isn't aware of its own name.

solutions/aakarshit-uppal/solution_1_2/solution.py
```diff
@@ -0,0 +1,263 @@
...
+        self.alias = alias
+        self.required = required
+        self.type_ = type
+        self.nvals = nvals
+        self.setval = setval
```
kaustubh-karkare: I disagree with the creators of argparse, and am of the opinion that this is bad naming. Can you fix it?

kaustubh-karkare:
Inactive for too long.

PR #4, opened 2017-03-12
Solution 1.2

K-ratos:
i tried to solve the problem 1.2
please review the code
usage is like ./arparse --key1=value1 --key2=value2 -V(and other options with one dash [currently hzk is in the code]) rest commands and subcommands(currently supporting only 2 additional arguments)


solutions/Abhishek/arparse
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+arguments=$@
+set -- $arguments
+
+#required
```
kaustubh-karkare: space after `#`. Applies everywhere.

solutions/Abhishek/arparse
```diff
@@ -0,0 +1,104 @@
...
+done
+
+help_statement="USAGE: ./arparse [--key=value] [--name=value] [-V] [-k value] [-h] [-z] other arguments ..."
+
+#can't be used together
+not_together="--local --remote: " #add : separated can't be used together options
+IFS_OLD=$IFS
+IFS=$:
+no_of_arguments=0
+for t in $not_together;do
```
kaustubh-karkare: Looks like there should be a space after the `;`.

solutions/Abhishek/arparse
```diff
@@ -0,0 +1,104 @@
...
+    echo "$r is required"
+    echo "use arparse -h to see help"
+    exit 0
+  fi
+done
+
+help_statement="USAGE: ./arparse [--key=value] [--name=value] [-V] [-k value] [-h] [-z] other arguments ..."
+
+#can't be used together
+not_together="--local --remote: " #add : separated can't be used together options
```
kaustubh-karkare: What if there are other sets of arguments that cant appear together?
Something like either `--local` or `--remote`, and one of `--bridged`, `--nat` or `--host`. We want something that is more flexible than this.

solutions/Abhishek/arparse
```diff
@@ -0,0 +1,104 @@
...
+done
+IFS=$IFS_OLD
+arguments=$arguments_copy
+set -- $arguments
+
+#options
+options=":Vhk:z" #add options as required by adding option letter and defining its function below
+while getopts $options opt;do
+  case $opt in
+    V) json_object[Verbose]="true";;
```
kaustubh-karkare: Looks like these options are hard coded, which isn't pretty. Make it flexible?

solutions/Abhishek/arparse
```diff
@@ -0,0 +1,104 @@
...
+  esac
+  shift
+done
+
+
+#comands and subcommands
+for (( i=1 ; i<=$# ; ++i ));do
+  case $i in
+    1) json_object[command]=$1;;
+    2) json_object[subcommand]=$2;;
```
kaustubh-karkare: Same as above. Hardcoded. What if I wanted the 2nd positional argument to be called something else?

solutions/Abhishek/arparse
```diff
@@ -0,0 +1,104 @@
...
+    *) echo "unknown no. of commands put"
+       echo "please use arparse -h to see help"
+       exit 1;;
+  esac
+done
+
+#final step -validation
+for i in "${!json_object[@]}"
+do
```
kaustubh-karkare: Why does the `do` appear on the next line after a `for` loop, but appears on the same line as `while`? If it's just a matter of preference, the most important standard is consistency.

solutions/Abhishek/arparse
```diff
@@ -0,0 +1,104 @@
...
+    local) json_object[$i]="true";;
+    remote) json_object[$i]="true";;
+  esac
+done
+
+#echoing all the results
+echo -n "{ "
+for i in "${!json_object[@]}"
+do
+  echo -n "$i : ${json_object[$i]}, "
```
kaustubh-karkare: So there will be a trailing comma after the last value?

solutions/Abhishek/arparse
```diff
@@ -0,0 +1,104 @@
...
+done
+IFS=$IFS_OLD
+
+#declarence of array to hold key value pair
+declare -A json_object
+
+#taking the --key=value pairs
+arguments_copy=$arguments
+for each in $arguments;do
+  if [[ $each == --* ]];then
```
kaustubh-karkare: Space after `;` ?

solutions/Abhishek/arparse
K-ratos: will fix that

solutions/Abhishek/arparse
K-ratos: you can put another set of not_together by separating with colon like => 
not_together="--local --remote: --bridged --nat --host: "

solutions/Abhishek/arparse
K-ratos: it works like that as well .. but i will fix that

solutions/Abhishek/arparse
K-ratos: yeah same, i will fix that too

solutions/Abhishek/arparse
K-ratos: how do you wan't it .. i don't understand ... you can add options there like =>
options=":Vhk:zlm:n"
and in case add whatever is it you want with lmn like=>
l) json_object[whatever_key1]="whatever_value1";;
m) json_object[whatever_key2]=$OPTARG;; # if you want user input as value for given option
n) json_object[whatever_key3]="whatever_value2";;

solutions/Abhishek/arparse
K-ratos: yes.. i will fix that

solutions/Abhishek/arparse
K-ratos: yeah its just printing the values, actually the json_obect is associative array or map sort of so key value pair are stored in that.. so i think that shouldn't matter.. but if it does i will fix that ..

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
+import sys
+
+
+class optional_arg_exception(Exception):
+
+    def __init__(self, message):
+        self.message = message
```
kaustubh-karkare: Don't think you need this. The default behavior should be the same.

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+
+class parser():
+
+    def __init__(self):
+        # positional arguments *necessary arguments*
+        self.positional_arguments = []
+        # optional arguments like --help
+        self.optional_arguments = [{'name': '--help', 'shortname': '-h',
+                                    'arg_type': 'string', 'required': False,
+                                    'help': 'prints this help statement'}]
```
kaustubh-karkare: Instead of this, call `add_optional_arguments` or something at the end of `__init__`.

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+        # arguments that can't be used together
+        self.mutually_exclusive_group = []
+        self.mutually_exclusive_group_shortname = []
+        # help statement
+        self.usage_statement = ""
+        self.help_statement = ""
+
+    # add positional necessary arguments
+    def add_argument(self, name, help='', arg_type='string'):
```
kaustubh-karkare: This function is called `add_argument`. But it is used for `positional_arguments`.

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+        if "--" not in name:
+            raise optional_arg_exception("'--' missing for optional_arguments")
+        else:
+            self.optional_arguments.append(
+                {'name': name, 'shortname': shortname, 'help': help, 'arg_type':
+                 arg_type, 'required': required})
+
+    # find index of dictionary from key value pair
+    def find_index(self, dicts, key, value):
+        class Null:
```
kaustubh-karkare: What is going on here?

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+
+    # add positional necessary arguments
+    def add_argument(self, name, help='', arg_type='string'):
+        self.positional_arguments.append(
+            {'name': name, 'help': help, 'arg_type': arg_type})
+
+    # add optional arguments
+    def add_optional_arguments(self, name, shortname='', help="",
+                               arg_type='string', required=False):
+        if "--" not in name:
```
kaustubh-karkare: Why is this a restriction?

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+        self.usage_statement = ""
+        self.help_statement = ""
+
+    # add positional necessary arguments
+    def add_argument(self, name, help='', arg_type='string'):
+        self.positional_arguments.append(
+            {'name': name, 'help': help, 'arg_type': arg_type})
+
+    # add optional arguments
+    def add_optional_arguments(self, name, shortname='', help="",
```
kaustubh-karkare: Empty strings are very dangerous. I guarantee you're going to spend hours due to bugs caused by using empty strings like this :P

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+        else:
+            self.optional_arguments.append(
+                {'name': name, 'shortname': shortname, 'help': help, 'arg_type':
+                 arg_type, 'required': required})
+
+    # find index of dictionary from key value pair
+    def find_index(self, dicts, key, value):
+        class Null:
+            pass
+        for i, d in enumerate(dicts):
```
kaustubh-karkare: Make it a habit to use proper descriptive names, so that you don't have to think about it later. Even if the loop is tiny like this.

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+            h = each['name'] + '            ' + each['help'] + '\n'
+            self.help_statement += h
+            h = ' <' + each['name'] + '>'
+            self.usage_statement += h
+        for each in self.optional_arguments:
+            h = each['name'] + '/' + each['shortname'] + \
+                '              ' + each['help'] + '\n'
+            self.help_statement += h
+            h = ' [' + each['name'] + '/' + each['shortname'] + ']'
+            self.usage_statement += h
```
kaustubh-karkare: Instead of regenerating this every time you call parse, why not do it only once?

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+                        raise optional_arg_exception(
+                            'please eanter no for ' + argument['name'])
+
+        final_json = {}
+        name_list_of_arguments = []
+        for each in self.positional_arguments:
+            name_list_of_arguments.append(each['name'])
+        for each in self.optional_arguments:
+            name_list_of_arguments.append(each['name'])
+
```
kaustubh-karkare: Shouldn't it be possible to have an argument that is not optional, but also not positional?

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+        for each in local_argument:
+            final_json[each] = local_argument[each]
+        for each in local_optional_argument:
```
kaustubh-karkare: In all these loops, please use a better name than `each`.

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+                local_argument[self.positional_arguments[
+                    count_of_arguments - 1]['name']] = argument
+            if argument == "--help" or argument == "-h":  # if argument has -h
+                                                        # or --help in it
+                print(self.usage_statement + "\n\n" + self.help_statement)
+                exit()
+            if '--' in argument and '=' in argument:
+                x = argument.split('=')
+                if (self.optional_arguments[self.find_index(
+                        self.optional_arguments, 'name', x[0])]['shortname']
```
kaustubh-karkare: I'm seeing this pattern of using find_index way too many times. Why not use a dict or something.

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+        if len(self.positional_arguments) != count_of_arguments:
+            raise arguments_error(
+                "please enter valid positional arguments or use python3 " +
+                argv[0] + " -h for help")
+
+        # checking mutually exclusive arguments
+        for list_of_grouped_arguments in self.mutually_exclusive_group:
+            count_of_optional_arguments = 0
+            for each in list_of_grouped_arguments:
+                sn = self.optional_arguments[self.find_index(
```
kaustubh-karkare: `sn`?

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+                if (each in local_optional_argument
+                        or sn in local_optional_argument):
+                    count_of_optional_arguments += 1
+                    if count_of_optional_arguments >= 2:
+                        raise optional_arg_exception(
+                            'mutually exclusive arguments used simultaneously' +
+                            str(list_of_grouped_arguments))
+
+        # checking for required arguments in optional_arguments and if boolean
+        # is having some value other than true false
```
kaustubh-karkare: Shouldn't this validation logic be in a separate method?

This one is already pretty long ... lets not complicate it.

solutions/abhishek-kujur/omega1.2/argparse.py
```diff
@@ -0,0 +1,256 @@
...
+        '--remote', '-r', 'remote thing whatever', 'boolean')
+    parser1.add_optional_arguments(
+        '--nat', '', 'remote thing whatever', 'boolean')
+    parser1.add_optional_arguments(
+        '--bridged', '-b', 'remote thing whatever', 'boolean')
+    parser1.add_optional_arguments(
+        '--host', '', 'remote thing whatever', 'boolean')
+    parser1.add_to_group('--local', '--remote')
+    parser1.add_to_group('--nat', '--bridged', '--host')
+    x = parser1.parse(args)
```
kaustubh-karkare: `parsed_arguments`?

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: okay

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: yeah .. will fix that

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: okay i'll use None instead

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: i thought optional argumnts have '--' in their name so i just added a measure to check

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: yes

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: i'll make the change

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: okay

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: i though such a thing  won't matter . sigh

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: okay ...

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: okay

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: yes .  it is in main function i thought it won't matter a lot

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: i don't understand , you mean optional argument that is required . if it is that . then you can turn on the boolean required variable in add_optional_argument method

solutions/abhishek-kujur/omega1.2/argparse.py
K-ratos: i don't understand

kaustubh-karkare: Basically, the `Exception` baseclass already does this, and so this is redundant, and therefore unnecessary.

solutions/abhishek-kujur/omega1.2/argparse.py
kaustubh-karkare: I don't think that is a restriction. The "Error: The '--key' argument is required, but missing from input." example in the problem is a clear counterexample.

solutions/abhishek-kujur/omega1.2/argparse.py
kaustubh-karkare: Haha. I'd rather you make the habit of using good names in all places, instead of deciding that based on how it is used. Imagine if this one line was part of a stack trace: a good name could help you figure out things faster than if you were forced to explore the code to find the original of this data.

solutions/abhishek-kujur/omega1.2/argparse.py
kaustubh-karkare: I try to review code thoroughly, and like to ensure consistency :)

PR #6, opened 2017-03-13
Koushik

koushikmln:
Made changes to 2.2

### DIFF
--- a/solutions/koushik/assignment-1/assignment-1.2.js
+++ b/solutions/koushik/assignment-1/assignment-1.2.js
@@ -1,0 +1,32 @@
+#!/usr/bin/nodejs
+var parser = require('./parser');
+
+//Add Random Options
+parser.addArgumentWithShortname(
+	'--key',
+	'-k',
+	"Integer Argument",
+	{ isDataVar: true, dataType: 1 , isRequired: true}
+);
+parser.addArgument('--value',"String Argument",{isDataVar: true});
+parser.addArgument('--local',"Random Exclusive Argument",{group: 1});
+parser.addArgument('--remote',"Random Exclusive Argument",{group: 2});
+//Parse the Argv
+parser.parse(process.argv);
+
+//Print Key Value JSON
+if(parser.options.key && parser.options.key.flag){
+	var json = {}
+	json.key = parser.options.key.value;
+	json.value = "";
+	if(parser.options.value && parser.options.value.flag)
+		json.value = parser.options.value.value;
+	console.log(json);
+}
+//Print Commands
+if(parser.commands.length > 0){
+	var json = {};
+	json.command = parser.commands[0];
+	json.subCommands = parser.commands.slice(1).join(" ");
+	console.log(json);
+}
--- a/solutions/koushik/assignment-1/parser.js
+++ b/solutions/koushik/assignment-1/parser.js
@@ -1,0 +1,99 @@
+/**
+ * Parser Object to Parse Argv
+ */
+function Parser() {
+	this.arguments = [];
+	this.options = [];
+	this.commands = [];
+	this._group = 0;
+	this._groupCommand = "";
+}
+
+/**
+ * Add Argument to Parser. Default Datatype is String(0), Int is 1
+ * Required is default false.
+ */
+Parser.prototype.addArgumentWithShortname = function(name,shortname,description,argOpts) {
+	var argument = {};
+	argument.name = name;
+	argument.shortname = shortname;
+	argument.description = description||'';
+	argument.dataFlag = (typeof argOpts.isDataVar === 'undefined') ? false : argOpts.isDataVar;
+	argument.dataType = (typeof argOpts.dataType === 'undefined') ? 0 : argOpts.dataType;
+	argument.required = (typeof argOpts.isRequired === 'undefined') ? false : argOpts.isRequired;
+	argument.group = (typeof argOpts.group === 'undefined') ? 0 : argOpts.group;
+	this.arguments.push(argument);
+};
+
+//Add Argument without Shortname
+Parser.prototype.addArgument = function(name,description,argOpts) {
+	this.addArgumentWithShortname(name,null,description,argOpts)
+};
+
+Parser.prototype.printHelp = function() {
+	console.log("Usage Information: program [options]"  );
+	console.log("Options:");
+	this.arguments.forEach(function(argument){
+		console.log(argument.name + ": " + argument.description);
+	});
+};
+
+Parser.prototype.parse = function(argv) {
+	//Split into arg and value if "=" exists
+	var self = this;
+	var args = [];
+	var values = [];
+	argv.forEach(function(arg){
+		var data = arg.split("=");
+		args.push(data[0]);
+		values.push(data[1]);
+	});
+	//Parse the argv while checking for each Option
+	this.arguments.forEach(function(argument){
+		if (
+			args.indexOf(argument.name)!=-1 || 
+			(
+				argument.shortname!=null && 
+				args.indexOf(argument.shortname)!=-1
+			)
+		) {
+			//Check for mutually Exclusive groups
+			if (argument.group != 0 && self._group == 0) {
+				self._group = argument.group;
+				self._groupCommand = argument.name;
+			}else if (argument.group != self._group) {
+				console.error("Error: Options " + argument.name + " & " + self._groupCommand + " cannot be used together.");
+				process.exit(1);
+			}
+			//Check and assign value
+			var index = args.indexOf(argument.name);
+			if (index == -1) {
+				index = args.indexOf(argument.shortname);
+			}
+			var value = values[index];
+			//Check for Integer Datatype
+			if (argument.dataFlag && argument.dataType == 1) {
+				value = parseInt(value);
+				//Throw Error and Exit if NaN
+				if (isNaN(value)) {
+					console.error("Error: Argument " + argument.name + " must have an Integer Value");
+					process.exit(1);
+				}
+			}
+			self.options[argument.name.replace(/-/g,"")] = {};
+			self.options[argument.name.replace(/-/g,"")].flag = true;
+			self.options[argument.name.replace(/-/g,"")].value = value;
+		}else if (argument.required == true) {
+			console.error("Error: Argument " + argument.name + " Missing");
+			process.exit(1);
+		}
+	});
+	//Check for Commands
+	args.forEach(function(arg){
+		if (!arg.startsWith("-") && !arg.startsWith("/")) {
+			self.commands.push(arg);
+		}
+	});
+};
+
+module.exports = new Parser();
--- a/solutions/koushik/assignment-2/assignment-2.2.py
+++ b/solutions/koushik/assignment-2/assignment-2.2.py
@@ -4,10 +4,10 @@ import logging
 import argparse
 import threading
 import time
-from urlparse import urlparse
 from mimetools import Message
 from StringIO import StringIO
 import os
+import re
 LEVELS = {'DEBUG': logging.DEBUG,
           'INFO': logging.INFO,
           'WARNING': logging.WARNING,
@@ -15,7 +15,28 @@ LEVELS = {'DEBUG': logging.DEBUG,
           'CRITICAL': logging.CRITICAL}
 THREAD_COUNT = 4
 
-class Client (threading.Thread):
+class SocketConnection():
+	def __init__(self, hostname, port):
+		self.hostname = hostname
+		self.port = port
+
+	def __enter__(self):
+		self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+		try:
+			self.sock.connect((self.hostname, self.port))
+		except socket.error, msg:
+			sys.stderr.write("[ERROR] %s\n" % msg[1])
+			sys.exit(1)
+		return self.sock
+
+	def __exit__(self, exc_type, exc_value, traceback):
+		if exc_type is not None:
+			print exc_type, exc_value, traceback
+		self.sock.close()
+		return self
+
+
+class DownloadThread (threading.Thread):
     def __init__(self, thread_id, hostname, 
     			 port, path, start_range, end_range):
         threading.Thread.__init__(self)
@@ -29,50 +50,44 @@ class Client (threading.Thread):
     def run(self):
 		logging.info("[Start: %s]", self.start_range)
 		logging.info("[End: %s]", self.end_range)
-		sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-		try:
-			sock.connect((self.hostname, self.port))
-		except socket.error, msg:
-			sys.stderr.write("[ERROR] %s\n" % msg[1])
-			sys.exit(1)
-		# Send HEAD Request to fetch Headers  
-		sock.send("GET %s HTTP/1.1\r\nHost: %s\r\nRange: bytes=%s-%s\r\nConnection: close\r\n\r\n"
-			% (self.path, self.hostname, self.start_range, self.end_range))
-		data = sock.recv(512)
-		recv = ""
-		while len(data):
-			recv = recv + data
+		# Establish Socket Connection
+		with SocketConnection(self.hostname, self.port) as sock:
+			# Send HEAD Request to fetch Headers  
+			sock.send("GET %s HTTP/1.1\r\nHost: %s\r\nRange: bytes=%s-%s\r\nConnection: close\r\n\r\n"
+				% (self.path, self.hostname, self.start_range, self.end_range))
 			data = sock.recv(512)
+			recv = ""
+			while len(data):
+				recv = recv + data
+				data = sock.recv(512)
 		# Write to Partfile after removing Headers
-		with open('partfile'+str(self.thread_id),'w') as f:
+		with open('partfile' + str(self.thread_id),'w') as f:
 			f.write("".join(recv.split("\r\n\r\n")[1:]))
-		sock.close()
 
 
+def parse_url(url):
+	url_pattern = re.compile("(https?)?(\:\/\/)?([a-zA-Z0-9.]+)\:?(\d+)?(.*)")
+	url_group = url_pattern.match(url)
+	return url_group
 
 """ GET contents from a URL in multiple threads. 
 Accepts URL as command line argument. 
 """
 def get(url):
-	url = urlparse(url)
-	HOSTNAME = url.hostname
-	PORT = url.port or 80
-	PATH = url.path
-	logging.info("[Host: %s Port: %s Path: %s]", HOSTNAME, PORT, PATH)
-	sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-	try:
-		sock.connect((HOSTNAME, PORT))
-	except socket.error, msg:
-		sys.stderr.write("[ERROR] %s\n" % msg[1])
-		sys.exit(1)
-	# Send HEAD Request to fetch Headers 
-	sock.send("HEAD %s HTTP/1.0\r\nHost: %s\r\n\r\n" % (PATH, HOSTNAME))
-	data = sock.recv(1024)
-	recv = ""
-	while len(data):
-		recv = recv + data
+	url = parse_url(url)
+	hostname = url.group(3)
+	port = url.group(4) or 80
+	path = url.group(5)
+	logging.info("[Host: %s port: %s path: %s]", hostname, port, path)
+	# Establish Socket Connection
+	with SocketConnection(hostname, port) as sock:
+		# Send HEAD Request to fetch Headers 
+		sock.send("HEAD %s HTTP/1.0\r\nHost: %s\r\n\r\n" % (path, hostname))
 		data = sock.recv(1024)
-	sock.close()
+		recv = ""
+		while len(data):
+			recv = recv + data
+			data = sock.recv(1024)
 	# Parse Headers using MIMETools to Fetch Content-Length
 	request_line, headers_alone = recv.split('\r\n', 1)
 	headers = Message(StringIO(headers_alone))
@@ -81,27 +96,32 @@ def get(url):
 	# initialize threads for download
 	thread_count = THREAD_COUNT;
 	bytes_per_thread = int(content_length)/thread_count;
-	clients = []
+	download_threads = []
 	for x in xrange(0,thread_count):
-		start_range = (x)*bytes_per_thread
+		start_range = (x) * bytes_per_thread
 		if x is not 0:
 			start_range += 1
-		end_range = (x+1)*bytes_per_thread
+		end_range = (x+1) * bytes_per_thread
 		if x == (thread_count-1):
 			end_range = int(content_length) - 1
-		client = Client(x,HOSTNAME,PORT,PATH,start_range,end_range)
-		clients.append(client)
+		download_thread = DownloadThread(x,hostname,port,path,start_range,end_range)
+		download_threads.append(download_thread)
 	for x in xrange(0,thread_count):
-		clients[x].start()
+		download_threads[x].start()
 	while threading.active_count() > 1:
 		time.sleep(1)
 		logging.info("Download Thread Running")
 	logging.info("Joining Files")
-	# Joining partfiles using cat
-	file_name = os.path.basename(PATH)
-	os.system("cat partfile? > "+os.path.basename(file_name))
+	# Joining partfiles
+	file_name = os.path.basename(os.path.basename(path))
+	with open(file_name, 'w') as outfile:
+	    for x in xrange(0,thread_count):
+	        with open("partfile" + str(x)) as infile:
+	            for line in infile:
+	                outfile.write(line)
 	# Removing Partfiles
-	os.system("rm partfile?")
+	for x in xrange(0,thread_count):
+		os.remove("partfile" + str(x))
 	logging.info("Exiting")
 
 if __name__ == '__main__':
--- a/solutions/koushik/assignment-2/assignment-2.2.py
+++ b/solutions/koushik/assignment-2/assignment-2.2.py
@@ -68,16 +68,21 @@ class DownloadThread (threading.Thread):
 def parse_url(url):
 	url_pattern = re.compile("(https?)?(\:\/\/)?([a-zA-Z0-9.]+)\:?(\d+)?(.*)")
 	url_group = url_pattern.match(url)
-	return url_group
+	url_object = {}
+	url_object["scheme"] = url_group.group(0)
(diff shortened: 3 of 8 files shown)


solutions/kmln/assignment-1/parser.js
```diff
@@ -0,0 +1,93 @@
+/**
+* Parser Object to Parse Argv
+*/
+function Parser() {
+	this.arguments = [];
```
kaustubh-karkare: According to the style guide, use 2 spaces for indenting, not tab.

solutions/kmln/assignment-1/parser.js
```diff
@@ -0,0 +1,93 @@
...
+	this.options = [];
+	this.commands = [];
+	this._group = 0;
+	this.__groupCommand = "";
```
kaustubh-karkare: What is the difference between single vs double underscore above?

solutions/kmln/assignment-1/parser.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+* Add Argument to Parser. Default Datatype is String(0), Int is 1
+* Required is default false.
+*/
+Parser.prototype.addArgumentWithShortname = function(name,shortname,description,dataFlag,dataType,required,group) {
```
kaustubh-karkare: Exceeds 80 char limit for lines.

solutions/kmln/assignment-1/assignment-1.2.js
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/nodejs
+var parser = require('./parser');
+
+//Add Random Options
+parser.addArgumentWithShortname('--key','-k',"Integer Argument",true,1,true);
+parser.addArgument('--value',"String Argument",true,0,false);
+parser.addArgument('--local',"Random Exclusive Argument",true,0,false,1);
```
kaustubh-karkare: Good API design ensures that it obvious what is happening. In this case, I have no idea what these integer and boolean arguments represent, so reading the code doesn't help me. You shouldn't force the user to read your parser implementation. Can you figure out a better way to do this?

Maybe https://en.wikipedia.org/wiki/Builder_pattern

solutions/kmln/assignment-1/assignment-1.2.js
```diff
@@ -0,0 +1,27 @@
...
+		json.value = parser.options.value.value;
+	console.log(json);
+}
+//Print Commands
+if(parser.commands.length > 0){
+	var json = {};
+	json.command = parser.commands[0];
+	json.subCommands = parser.commands.slice(1).join(" ");
+	console.log(json);
+}
```
kaustubh-karkare: There should be a newline at the end of the file.

solutions/kmln/assignment-1/parser.js
```diff
@@ -0,0 +1,93 @@
+/**
+* Parser Object to Parse Argv
```
kaustubh-karkare: Align the first star:
```
/**
 * ...
 */
```

solutions/kmln/assignment-1/assignment-1.2.js
```diff
@@ -0,0 +1,27 @@
...
+//Add Random Options
+parser.addArgumentWithShortname('--key','-k',"Integer Argument",true,1,true);
+parser.addArgument('--value',"String Argument",true,0,false);
+parser.addArgument('--local',"Random Exclusive Argument",true,0,false,1);
+parser.addArgument('--remote',"Random Exclusive Argument",true,0,false,2);
+//Parse the Argv
+parser.parse(process.argv);
+
+//Print Key Value JSON
+if(parser.options.key && parser.options.key.flag){
```
kaustubh-karkare: Space after `if` and before `{`

solutions/kmln/assignment-1/parser.js
```diff
@@ -0,0 +1,93 @@
...
+	});
+};
+
+Parser.prototype.parse = function(argv) {
+	//Split into arg and value if "=" exists
+	var self = this;
+	var args = [];
+	var values = [];
+	argv.forEach(function(arg){
+		var data = arg.split("=");
```
kaustubh-karkare: Can you rely that the `=` will always be there?

solutions/kmln/assignment-1/parser.js
```diff
@@ -0,0 +1,93 @@
...
+		args.push(data[0]);
+		values.push(data[1]);
+	});
+	//Parse the argv while checking for each Option
+	this.arguments.forEach(function(argument){
+		if(args.indexOf(argument.name)!=-1 ||(argument.shortname!=null && args.indexOf(argument.shortname)!=-1)){
```
kaustubh-karkare: Space after `||`. Exceeds 80 chars.

solutions/kmln/assignment-1/parser.js
koushikmln: i assumed it for this scenario. so we might have to handle cases like --key20 or --key 20 which can be added if required.

solutions/kmln/assignment-1/parser.js
koushikmln: nothing. it was a typo.

solutions/koushik/assignment-1/parser.js
```diff
@@ -0,0 +1,99 @@
...
+	this.arguments.push(argument);
+};
+
+//Add Argument without Shortname
+Parser.prototype.addArgument = function(name,description,argOpts) {
+	this.addArgumentWithShortname(name,null,description,argOpts)
+};
+
+Parser.prototype.printHelp = function() {
+	console.log("Usage Information: program [options]"  );
```
kaustubh-karkare: Unnecessary spaces before `)`

solutions/koushik/assignment-1/parser.js
```diff
@@ -0,0 +1,99 @@
...
+	console.log("Options:");
+	this.arguments.forEach(function(argument){
+		console.log(argument.name + ": " + argument.description);
+	});
+};
+
+Parser.prototype.parse = function(argv) {
+	//Split into arg and value if "=" exists
```
kaustubh-karkare: Space after `//`

solutions/koushik/assignment-1/parser.js
```diff
@@ -0,0 +1,99 @@
...
+	var self = this;
```
kaustubh-karkare: Use arrow functions so that you don't need to do this.

solutions/koushik/assignment-1/parser.js
```diff
@@ -0,0 +1,99 @@
...
+	var values = [];
+	argv.forEach(function(arg){
+		var data = arg.split("=");
+		args.push(data[0]);
+		values.push(data[1]);
+	});
+	//Parse the argv while checking for each Option
+	this.arguments.forEach(function(argument){
+		if (
+			args.indexOf(argument.name)!=-1 ||
```
kaustubh-karkare: Prefer `!==`. This applies to all comparisons, until you intentionally decide not to do that.
Also, style requires spaces around operators.

solutions/koushik/assignment-1/parser.js
```diff
@@ -0,0 +1,99 @@
...
+			(
+				argument.shortname!=null && 
+				args.indexOf(argument.shortname)!=-1
+			)
+		) {
+			//Check for mutually Exclusive groups
+			if (argument.group != 0 && self._group == 0) {
+				self._group = argument.group;
+				self._groupCommand = argument.name;
+			}else if (argument.group != self._group) {
```
kaustubh-karkare: Space after `}`

solutions/koushik/assignment-1/parser.js
```diff
@@ -0,0 +1,99 @@
+/**
+ * Parser Object to Parse Argv
+ */
+function Parser() {
+	this.arguments = [];
```
kaustubh-karkare: In JS, this already has a special meaning, so I'd avoid reusing that name (even if it works), to prevent confusion.

solutions/koushik/assignment-1/parser.js
```diff
@@ -0,0 +1,99 @@
...
+	this.options = [];
+	this.commands = [];
+	this._group = 0;
+	this._groupCommand = "";
```
kaustubh-karkare: Why do some of these have underscores at the beginning, but others dont? You want to be consistent.
Additionally, there names aren't particularly useful, because as I'm reading this, I have no idea what the different between "arguments" and "options" is, or what "group command" is supposed to mean. Could you find names that make it super obvious?

solutions/koushik/assignment-1/parser.js
```diff
@@ -0,0 +1,99 @@
...
+	});
+	//Parse the argv while checking for each Option
+	this.arguments.forEach(function(argument){
+		if (
+			args.indexOf(argument.name)!=-1 || 
+			(
+				argument.shortname!=null && 
+				args.indexOf(argument.shortname)!=-1
+			)
+		) {
```
kaustubh-karkare: Instead of add nesting, prefer early return/continue.
```
if (!condition) continue;
// stuff

File: solutions/koushik/assignment-1/parser.js [hunk 18]
```diff
@@ -0,0 +1,99 @@
...
+				process.exit(1);
+			}
+			//Check and assign value
+			var index = args.indexOf(argument.name);
+			if (index == -1) {
+				index = args.indexOf(argument.shortname);
+			}
+			var value = values[index];
+			//Check for Integer Datatype
+			if (argument.dataFlag && argument.dataType == 1) {
```
kaustubh-karkare: You shouldn't assume that certain values of datatypes mean things. Pass in a function or something, which validates input for maximum flexibility.

solutions/koushik/assignment-1/parser.js
```diff
@@ -0,0 +1,99 @@
...
+				value = parseInt(value);
+				//Throw Error and Exit if NaN
+				if (isNaN(value)) {
+					console.error("Error: Argument " + argument.name + " must have an Integer Value");
+					process.exit(1);
```
kaustubh-karkare: Prefer to throw exceptions and let the caller figure out what they want (ask the user to fix the 1 mistake out of 20 inputs? or assume default and just continue) instead of assuming that we want to end the program.

solutions/koushik/assignment-1/parser.js
```diff
@@ -0,0 +1,99 @@
...
+				}
+			}
+			self.options[argument.name.replace(/-/g,"")] = {};
+			self.options[argument.name.replace(/-/g,"")].flag = true;
+			self.options[argument.name.replace(/-/g,"")].value = value;
```
kaustubh-karkare: Prefer not to store unnecessary state, since you can return the result here. What if I'm running this in a unittest, with 5 parse functions running parallel. There's no reason why they should conflict.

solutions/koushik/assignment-1/assignment-1.2.js
```diff
@@ -0,0 +1,32 @@
...
+);
+parser.addArgument('--value',"String Argument",{isDataVar: true});
+parser.addArgument('--local',"Random Exclusive Argument",{group: 1});
+parser.addArgument('--remote',"Random Exclusive Argument",{group: 2});
+//Parse the Argv
+parser.parse(process.argv);
+
+//Print Key Value JSON
+if(parser.options.key && parser.options.key.flag){
+	var json = {}
```
kaustubh-karkare: This isn't "json". This is an object.

solutions/koushik/assignment-1/assignment-1.2.js
```diff
@@ -0,0 +1,32 @@
...
+	json.key = parser.options.key.value;
+	json.value = "";
+	if(parser.options.value && parser.options.value.flag)
+		json.value = parser.options.value.value;
+	console.log(json);
+}
+//Print Commands
+if(parser.commands.length > 0){
```
kaustubh-karkare: I think you misunderstood the problem a bit. I expect something like:
```
parser.addPositionalArgument('command', "blah")
parser.addPositionalArgument('subcommand', "blah")
parser.addPositionalArgument('subsubcommand', "blah")
```
Ie - this is flexible too. Cant assume the names.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+""" GET contents from a URL in multiple threads. 
+Accepts URL as command line argument. 
+"""
+def get(url):
+	url = urlparse(url)
+	HOSTNAME = url.hostname
+	PORT = url.port or 80
+	PATH = url.path
```
kaustubh-karkare: Umm ... this kinda of capital case naming is usually reserved for global constants, not local variables.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,119 @@
+import socket
+import sys
+import logging
+import argparse
+import threading
+import time
+from urlparse import urlparse
+from mimetools import Message
+from StringIO import StringIO
```
kaustubh-karkare: Lets not use these 3 modules (urlparse, mimetools, stringio)

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,119 @@
...
+	logging.info("[Content-Length: %s]",content_length)
+	# initialize threads for download
+	thread_count = THREAD_COUNT;
+	bytes_per_thread = int(content_length)/thread_count;
+	clients = []
+	for x in xrange(0,thread_count):
+		start_range = (x)*bytes_per_thread
+		if x is not 0:
+			start_range += 1
+		end_range = (x+1)*bytes_per_thread
```
kaustubh-karkare: spaces around operators

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,119 @@
...
+		if x == (thread_count-1):
+			end_range = int(content_length) - 1
+		client = Client(x,HOSTNAME,PORT,PATH,start_range,end_range)
```
kaustubh-karkare: Not sure Client is the best name for that you're doing here.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,119 @@
...
+		clients.append(client)
+	for x in xrange(0,thread_count):
+		clients[x].start()
+	while threading.active_count() > 1:
+		time.sleep(1)
```
kaustubh-karkare: Isn't this equivalant to thread.join()?

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,119 @@
...
+		logging.info("Download Thread Running")
+	logging.info("Joining Files")
+	# Joining partfiles using cat
+	file_name = os.path.basename(PATH)
+	os.system("cat partfile? > "+os.path.basename(file_name))
```
kaustubh-karkare: Instead of using raw bash commands, use the proper API, so that it works on all platforms.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,119 @@
...
+	parser.add_argument("url", help="Enter the URL to Fetch Content")
+	parser.add_argument("--log", type=str,
+                    help="Set Log Level(DEBUG INFO WARNING CRITICAL)")
+	args = parser.parse_args()
+	# Fetching and initializing Log Level
+	if args.log is not None:
+		level = LEVELS.get(args.log, logging.NOTSET)
+		logging.basicConfig(level=level)
+	# Initialize GET Request
+	get(args.url)
```
kaustubh-karkare: Add an EOL before EOF.

solutions/koushik/assignment-2/assignment-2.2.py
koushikmln: i used mimetools for getting content-length header.. its just normal string split and parsing alternative.. removed urlparse

solutions/koushik/assignment-2/assignment-2.2.py
koushikmln: i let it be for now to just to use logger. can remove it if its affecting anything.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+import sys
+import logging
+import argparse
+import threading
+import time
+from mimetools import Message
+from StringIO import StringIO
+import os
+import re
+LEVELS = {'DEBUG': logging.DEBUG,
```
kaustubh-karkare: There should be an empty line after the import statements.
Additionally, I had feedback about the formatting here last time, which was not addressed.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+          'INFO': logging.INFO,
+          'WARNING': logging.WARNING,
+          'ERROR': logging.ERROR,
+          'CRITICAL': logging.CRITICAL}
+THREAD_COUNT = 4
```
kaustubh-karkare: This should be input, not constant.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+
+class SocketConnection():
+	def __init__(self, hostname, port):
+		self.hostname = hostname
+		self.port = port
+
+	def __enter__(self):
+		self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
kaustubh-karkare: Avoid abbreviations like "sock". 2 extra letters doesn't make more than a few seconds, but makes the code much easier to read. See my comments on the group, where I emphasize the importance of this.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+		try:
+			self.sock.connect((self.hostname, self.port))
+		except socket.error, msg:
+			sys.stderr.write("[ERROR] %s\n" % msg[1])
```
kaustubh-karkare: Using `logging`

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+		return self.sock
+
+	def __exit__(self, exc_type, exc_value, traceback):
+		if exc_type is not None:
+			print exc_type, exc_value, traceback
+		self.sock.close()
+		return self
+
+
+class DownloadThread (threading.Thread):
```
kaustubh-karkare: No space before `(`

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+    def __init__(self, thread_id, hostname, 
+    			 port, path, start_range, end_range):
```
kaustubh-karkare: There's a convention I follow: If you cant fit the entire function signature/call on one line, you should have each element on its own line.
```
def __init__(
  self,
  thread_id,
  hostname,
  ...
):
  ...

File: solutions/koushik/assignment-2/assignment-2.2.py [hunk 36]
```diff
@@ -0,0 +1,144 @@
...
+        self.start_range = start_range
+        self.end_range = end_range
+
+    def run(self):
+		logging.info("[Start: %s]", self.start_range)
+		logging.info("[End: %s]", self.end_range)
+		# Establish Socket Connection
+		with SocketConnection(self.hostname, self.port) as sock:
+			# Send HEAD Request to fetch Headers  
+			sock.send("GET %s HTTP/1.1\r\nHost: %s\r\nRange: bytes=%s-%s\r\nConnection: close\r\n\r\n"
```
kaustubh-karkare: Use multiline strings to make this easier to read.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+				% (self.path, self.hostname, self.start_range, self.end_range))
+			data = sock.recv(512)
```
kaustubh-karkare: Dont assume constants like this in code. Put this in a global variable defined at the top.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+			recv = ""
+			while len(data):
+				recv = recv + data
+				data = sock.recv(512)
+		# Write to Partfile after removing Headers
+		with open('partfile' + str(self.thread_id),'w') as f:
```
kaustubh-karkare: Use `tempfile` module. Also, space after `,`. Also, `f` is a bad name.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+			f.write("".join(recv.split("\r\n\r\n")[1:]))
+
+
+def parse_url(url):
+	url_pattern = re.compile("(https?)?(\:\/\/)?([a-zA-Z0-9.]+)\:?(\d+)?(.*)")
```
kaustubh-karkare: It looks like `:kaustubh.io:/:::::` is valid, but `kaustubh-karkare.blogspot.com` is not.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+	url_group = url_pattern.match(url)
+	url_object = {}
+	url_object["scheme"] = url_group.group(0)
+	url_object["hostname"] = url_group.group(3)
+	url_object["port"] = url_group.group(4)
+	url_object["path"] = url_group.group(5).split('#')[0].split('?')[0]
+	return url_object
```
kaustubh-karkare: In general, dictionaries with a fixed set of keys like this not a good idea, because people can make typing mistakes or something. Prefer `namedtuple`.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+
+""" GET contents from a URL in multiple threads.
```
kaustubh-karkare: Shouldn't the sentence start from the next line?

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+		# Establish Socket Connection
+		with SocketConnection(self.hostname, self.port) as sock:
+			# Send HEAD Request to fetch Headers  
+			sock.send("GET %s HTTP/1.1\r\nHost: %s\r\nRange: bytes=%s-%s\r\nConnection: close\r\n\r\n"
+				% (self.path, self.hostname, self.start_range, self.end_range))
+			data = sock.recv(512)
+			recv = ""
+			while len(data):
+				recv = recv + data
+				data = sock.recv(512)
```
kaustubh-karkare: What if the size of the file was so large that keeping it all in memory would cause OOMs?

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+	logging.info("[Host: %s port: %s path: %s]", hostname, port, path)
+	# Establish Socket Connection
+	with SocketConnection(hostname, port) as sock:
+		# Send HEAD Request to fetch Headers 
+		sock.send("HEAD %s HTTP/1.0\r\nHost: %s\r\n\r\n" % (path, hostname))
+		data = sock.recv(1024)
+		recv = ""
+		while len(data):
+			recv = recv + data
+			data = sock.recv(1024)
```
kaustubh-karkare: This part looks duplicated. I'm not fond of duplication.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+	# Parse Headers using MIMETools to Fetch Content-Length
+	request_line, headers_alone = recv.split('\r\n', 1)
+	headers = Message(StringIO(headers_alone))
+	content_length = headers['content-length']
+	logging.info("[Content-Length: %s]",content_length)
+	# initialize threads for download
+	thread_count = THREAD_COUNT;
+	bytes_per_thread = int(content_length)/thread_count;
```
kaustubh-karkare: Spaces around operators.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+	download_threads = []
+	for x in xrange(0,thread_count):
+		start_range = (x) * bytes_per_thread
+		if x is not 0:
+			start_range += 1
```
kaustubh-karkare: I'm not fond of special casing. Isn't the range exclusive?

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+		end_range = (x+1) * bytes_per_thread
+		if x == (thread_count-1):
+			end_range = int(content_length) - 1
+		download_thread = DownloadThread(x,hostname,port,path,start_range,end_range)
+		download_threads.append(download_thread)
+	for x in xrange(0,thread_count):
+		download_threads[x].start()
+	while threading.active_count() > 1:
```
kaustubh-karkare: You're still using this. It assumes that only this class is allowed to create threads, which renders this module unusable for a large class of applications. Prefer join instead.

solutions/koushik/assignment-2/assignment-2.2.py
```diff
@@ -0,0 +1,144 @@
...
+	logging.info("Joining Files")
+	# Joining partfiles
+	file_name = os.path.basename(os.path.basename(path))
+	with open(file_name, 'w') as outfile:
+	    for x in xrange(0,thread_count):
+	        with open("partfile" + str(x)) as infile:
+	            for line in infile:
+	                outfile.write(line)
+	# Removing Partfiles
+	for x in xrange(0,thread_count):
```
kaustubh-karkare: Instead of a separate loop, delete after reading.

solutions/koushik/assignment-2/assignment-2.2.py
koushikmln: the range was inclusive in the documentation.. 0-0 downloaded 1 byte.

solutions/koushik/assignment-2/assignment-2.2.py
koushikmln: oh i got it. i had left it for logging the string. il use join.

kaustubh-karkare:
Inactive for too long.

PR #7, opened 2017-03-14
Updated problem 2 for week 1 - Paritosh Sharma

pari08tosh:
Submitting the 10th update to problem 1.2 according to suggestions in previous PR. Have made the following changes:-
1. Improved variable names and type to reduce confusion.


solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,127 @@
...
+      key: 'remote',
+      arg: '--remote',
+      type: 'Boolean',
+      required: false,
+      position: null
+   }
+];
+
+var parsed_json = {};
+var err = [],flag,j;
```
kaustubh-karkare: Spaces after `,`

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,127 @@
+
+// add new argument entries here
+
+var command_args  = [
```
kaustubh-karkare: You have an extra space before `=`.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,127 @@
...
+      position: null
+   },
+   {
+      key: 'remote',
+      arg: '--remote',
+      type: 'Boolean',
+      required: false,
+      position: null
+   }
+];
```
kaustubh-karkare: Is this the best way to provide input? How about the builder pattern?

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,127 @@
...
+
+var parsed_json = {};
+var err = [],flag,j;
+var entered_args = process.argv.slice(2);
+
+
+function print_json(err){
+   if(err)
+   {
```
kaustubh-karkare: You are inconsistently using these braces. In case of functions, you put them on the same line. In case of `if` and the next `for`, you put them on the next line. Please fix that. I prefer same line. Also, space before `{`

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,127 @@
...
+      console.log('ERROR: ' + err);
+      return;
+   }
+   for(i=0;i<command_args.length;i++)
```
kaustubh-karkare: Proper spacing:
```
for (i = 0; i < command_args.length; i++) {
```

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,127 @@
...
+   {
+      if(command_args[i].required === true && !(command_args[i].key in parsed_json)){
+         console.log( "ERROR: Value of " + command_args[i].arg + " required");
```
kaustubh-karkare: If this function is called "print_json", it is not allowed to do validation. Please move this code out.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,127 @@
...
+         return;
+      }
+   }
+   console.log(parsed_json);
+   return;
+}
+
+arg_parser(entered_args, print_json);
```
kaustubh-karkare: I had to search for where `entered_args` is defined. Which meant that my flow of reading was broken. Let's try and avoid that. Move all statements outside functions to the bottom if possible.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,127 @@
...
+
+function arg_parser(entered_args, callback){
+   for(var i=0; i<entered_args.length; i++){
+      var str = entered_args[i];
+      flag = 0;
+      if(str.search('-') === -1){
```
kaustubh-karkare: Why shouldn't it appear inside the string? Like `init-clean`

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,127 @@
...
+         return;
+      }
+   }
+   console.log(parsed_json);
+   return;
+}
+
+arg_parser(entered_args, print_json);
+
+function arg_parser(entered_args, callback){
```
kaustubh-karkare: Most of your variable / function names don't meet the conventions mentioned in the style guide.
Also, why not just return the result. Why do you need a callback? This isn't async.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,127 @@
...
+            return;
+         }
+      }
+      else{
+         str = str.split('=');
+         for(j=0; j<command_args.length; j++)
+         {
+            if(command_args[j].arg === str[0]){
+               flag=1;
+               if(command_args[j].type === 'String'){
```
kaustubh-karkare: Avoid raw strings like this. Prefer enums or constants, which will result in an error if mistyped.

pari08tosh:
I have updated the solution to the problem according to your suggestions. Please review it again.

pari08tosh:
Updated problem 1.2 according to the changes requested in the previous pull request. Have made the following updates -

    1. Updated input format to a more user friendly one.
    2. Edited and updated code according to the style guide.
    3 . Removed Bugs

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+      }
+   }
+
+   // output parsedJson
+
+   console.log(parsedJson);
+   return;
+}
+
+// add new arguments here in the format argBuilder('key', 'argumentName', 'type', 'required(Boolean)', position);
```
kaustubh-karkare: This separation of core logic and input/output is so much cleaner :)

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+
+argBuilder('command', null, 'String', false, 0);
+argBuilder('subcommand', null, 'String', false, 1);
+argBuilder('key', '--key', '+Int', true, null);
+argBuilder('name', '--name', 'String', false, null);
+argBuilder('verbrose', '-v', 'Boolean', false, null);
+argBuilder('local', '--local', 'Boolean', false, null);
+argBuilder('remote', '--remote', 'Boolean', false, null);
```
kaustubh-karkare: Next challenge: Notice how I'm forced to depend on your comment to understand the meanings of the last 2 arguments. It isn't intuitive enough from the code itself. This is why, in general, boolean arguments are discouraged. Can you modify this builder so that the meanings of all inputs are obvious?

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+
+var parsedJson = {};
+var err = [], flag, j;
```
kaustubh-karkare: I know this works, but it isn't intuitive how you declare a variable here, but use it before. Unfortunately, while `let` and `const` are slightly better, they have the same behavior in this regard. Why not just make these local variables.

In general, you want to minimize the amount of global state.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+      }
+   }
+
+   // validating. Add aditional validations here
+
+   if(err.length > 0) {
+      console.log('ERROR: ' + err);
+      return;
+   }
+   for(i = 0; i < argList.length; i++) {
```
kaustubh-karkare: Space after `if`, `for`, etc. Applies everywhere.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+         }
+         if(flag === 0) {
+            console.log(`ERROR: Invalid argument '${str}' entered`);
+            return;
+         }
+      }
+      else {
+         str = str.split('=');
+         for(j = 0; j < argList.length; j++)
+         {
```
kaustubh-karkare: You're starting the block from the next line. But in other places (like line 38), you're starting it from the same line. You need to be consistent.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+   tempObj.arg = arg;
+   tempObj.type = type;
+   tempObj.position = position;
+   tempObj.required = required;
+   argList.push(tempObj);
+}
+
+function argParser(inputArgs) {
+
+   for(var i = 0; i < inputArgs.length; i++) {
```
kaustubh-karkare: Tip: This isn't something you'll find in a style guide, but lots of people do this. Use `ii` instead of `i`, `jj` instead of `j`, etc. This makes those variables much easier to find. And if you use multiple cursors, you will immediately see the benefits.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+
+argTypeEnum = {
+   'String': 0,
+   'Boolean': 1,
+   'Int': 2,
+   '+Int': 3, // represents + integers
+};
+
+function argBuilder(key, arg, type, required, position) {
+   if(key === null){
```
kaustubh-karkare: Space before `{`. Applies everywhere.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+               }
+               else if(argTypeEnum[argList[j].type] === 3) {
+                  if( isNaN(str[1]) || str[1] * 1 < 0) {
+                     console.log(`ERROR: Invalid value of argument '${str[0]}'. Must be a positive integer`);
+                     return;
+                  }
+                  else {
+                     parsedJson[argList[j].key] = str[1];
+                  }
+               }
```
kaustubh-karkare: This section could be improved. Note how your validation is separate from your type definition. This means that if I want to add a new type, I'll need to update 2 places which is inconvenient. You want to simplify it to 1 place.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+                  if(isNaN(str[1])) {
+                     console.log(`ERROR: Invalid value of argument '${str[0]}'. Must be a positive integer`);
+                     return;
+                  }
+                  else {
+                     parsedJson[argList[j].key] = str[1];
+                  }
+               }
+               else if(argTypeEnum[argList[j].type] === 3) {
+                  if( isNaN(str[1]) || str[1] * 1 < 0) {
```
kaustubh-karkare: Whitespace isn't right. Should be `if (isNan`

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+                     console.log(`ERROR: Invalid value of argument '${str[0]}'. Must be a positive integer`);
+                     return;
+                  }
+                  else {
```
kaustubh-karkare: While consistency isn't a problem in this case, how about `} else {`?

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+         if(flag === 0) {
+            console.log(`ERROR: Invalid argument '${str[0]}' entered`);
+            return;
+         }
+      }
+   }
+
+   // validating. Add aditional validations here
+
+   if(err.length > 0) {
```
kaustubh-karkare: `err` isn't set anywhere, as far as I see.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,121 @@
...
+argBuilder('name', '--name', 'String', false, null);
+argBuilder('verbrose', '-v', 'Boolean', false, null);
+argBuilder('local', '--local', 'Boolean', false, null);
+argBuilder('remote', '--remote', 'Boolean', false, null);
+
+var parsedJson = {};
+var err = [], flag, j;
+var inputArgs = process.argv.slice(2);
+
+argParser(inputArgs);
```
kaustubh-karkare: Console log the result, instead of print from inside the method. That can make the method unused for other purposes.

pari08tosh:
Third update of problem 1.2. Modified code according to changes requested in last PR. Have made the following changes :-
1. Modified input method to improve user friendliness.
2. Modified method to add new argument types.
3. Simplified main parser method.
4. Restructured functions to improve code readability and usability.
5. Removed style errors and other bugs.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,112 @@
...
+var parsedJson = {};
+
+// add aditional argument types here e.g 'Type': 'Conditions on argValue'.
+// Enter 'true' if no conditions
+
+argTypeConditions = {
+   'String': 'true',
+   'Boolean': 'true',
+   'Int': '!isNaN(argValue)',
+   '+Int': '!isNaN(argValue) && argValue * 1 >= 0' // represents + integers.
```
kaustubh-karkare: Interesting. This works, but what if you use functions instead? `eval`ing string is generally not a good idea, due to performance/security issues.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,112 @@
...
+
+// add new arguments here in the format argBuilder('key', 'argumentName', 'type', 'required/notRequired', position(if necessary));
+
+argBuilder('command', null, 'String', 'notRequired', 0);
+argBuilder('subcommand', null, 'String', 'notRequired', 1);
+argBuilder('key', '--key', '+Int', 'required');
+argBuilder('name', '--name', 'String', 'notRequired');
+argBuilder('verbrose', '-v', 'Boolean', 'notRequired');
+argBuilder('local', '--local', 'Boolean', 'notRequired');
+argBuilder('remote', '--remote', 'Boolean', 'notRequired');
```
kaustubh-karkare: Much better. You got read of all the booleans, but we have different problems with strings. Basically, it is really easy to make typing mistakes that no-one notices, and so you might type in `notReqired` and your code does unexpected things. As a result, it is better to use Enums.
```
const IS_REQUIRED = {
  YES: true,
  NO: false,
};
```
This makes things clearer, and safe too. Another approach is:
```
argBuilder({
  name: 'key',
  isRequired: true,
  ...
})
```

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,112 @@
...
+
+function argValidator(printCheck) {
+   for (i = 0; i < argList.length; i++) {
+      if (argList[i].required === true && !(argList[i].key in parsedJson)) {
+         console.log(`ERROR: Value of '${argList[i].arg}'  required`);
+         return;
+      }
+   }
+   if (printCheck === true)
+      printParsedArgs();
```
kaustubh-karkare: As mentioned in the post yesterday, a method called `argValidator` has no business print stuff. Remove this.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,112 @@
...
+      console.log(`ERROR: Invalid require parameter for argumnet '${arg}'`);
+      process.exit(1);
+   }
+   argList.push(tempObj);
+}
+
+//argument validator function. Add additional validations here.
+//Pass printCheck = true to print parsed arguments.
+
+function argValidator(printCheck) {
```
kaustubh-karkare: Since all you do is to ensure that required arguments are prevent, why not call it that. This name makes it sound like you're going to validate the individual values.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+function printParsedArgs() {
+   console.log(parsedJson);
+}
+
+function argParser(inputArgs) {
+   var  ii, jj, flag, argValue;
+   for (ii = 0; ii < inputArgs.length; ii++) {
+      var str = inputArgs[ii];
```
kaustubh-karkare: `str` tells me what type it is, but not what it contains. Better name?

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,112 @@
...
+      flag = 0;
+      if (str.search('-') !== 0) {
```
kaustubh-karkare: `indexOf` is clearer, with the same results.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,112 @@
...
+         }
+      }
+   }
+   argValidator(true);
+}
+
+// add new arguments here in the format argBuilder('key', 'argumentName', 'type', 'required/notRequired', position(if necessary));
+
+argBuilder('command', null, 'String', 'notRequired', 0);
+argBuilder('subcommand', null, 'String', 'notRequired', 1);
```
kaustubh-karkare: The position argument is still not obvious, unfortunately.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,112 @@
...
+
+function argParser(inputArgs) {
+   var  ii, jj, flag, argValue;
+   for (ii = 0; ii < inputArgs.length; ii++) {
+      var str = inputArgs[ii];
+      flag = 0;
+      if (str.search('-') !== 0) {
+         for (jj = 0; jj < argList.length; jj++) {
+            if (argList[jj].position === ii) {
+               flag = 1;
```
kaustubh-karkare: Instead of using a flag, you could just continue after the next line. And show a common error message at the bottom.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,112 @@
...
+            console.log(`ERROR: Invalid argument '${str}' entered`);
+            return;
+         }
+      } else {
+         str = str.split('=');
+         for (jj = 0; jj < argList.length; jj++) {
+            if (argList[jj].arg === str[0]) {
+               flag = 1;
+               argValue = str[1];
+               if (eval(argTypeConditions[argList[jj].type]))  {
```
kaustubh-karkare: As mentioned above, prefer not use `eval`.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,112 @@
...
+                  if (argList[jj].type === 'Boolean')
+                     parsedJson[argList[jj].key] = true;
+                  else
+                     parsedJson[argList[jj].key] = argValue;
```
kaustubh-karkare: So even an Int will be stored as a string? That's not elegant. How about your validator function returns the value, cast in the proper type. That way, you dont have to special case Boolean.

pari08tosh:
Fourth update of problem 1.2. Modified code according to changes requested in last PR. Have made the following changes :-
1. Modified input method to improve user friendliness.
2. Modified method to add new argument types.
3. Simplified main parser method.
4. Restructured functions to improve code readability and usability.
5. Improved data type handling and error handling.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,117 @@
...
+argBuilder({ key: 'name', arg: '--name', type: 'String', isRequired: false, position: null });
+argBuilder({ key: 'verbrose', arg: '-v', type: 'Boolean', isRequired: false, position: null });
+argBuilder({ key: 'local', arg: '--local', type: 'Boolean', isRequired: false, position: null });
+argBuilder({ key: 'remote', arg: '--remote', type: 'Boolean', isRequired: false, position: null });
+
+var inputArgs = process.argv.slice(2);
+
+var parsedObj = argParser(inputArgs);
+
+if (parsedObj instanceof Error) {
```
kaustubh-karkare: You shouldn't be returning Error objects. You should throw them, and use try-catch.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,117 @@
+
+var argList = [];
+
+// add aditional argument types here e.g 'Type': function(argValue) [{ return condition; }, 'Data Type'.
+// Enter 'true' if no conditions
+
+argTypeConditions = {
+  'String': [function(argValue) { return true; }, 'String'],
```
kaustubh-karkare: Prefer objects instead of arrays like this, since it gives you an opportunity to specify intent using keys.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,117 @@
...
+      return true;
+    case 2:
+      return Number(argValue);
+  }
+}
+
+function argBuilder(argObj) {
+  if (argObj.key === null) {
+    console.log(`Error: Key of argument '${arg}' cannot be null`);
+    process.exit(1);
```
kaustubh-karkare: Throw errors. Don't exit early like this.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,117 @@
...
+  argList.push(argObj);
+}
+
+// argument validator function. Add additional validations here.
+
+function parsedObjValidator(tempObj) {
+
+  for (i = 0; i < argList.length; i++) {
+    if (argList[i].isRequired === true && !(argList[i].key in tempObj)) {
+      return new Error(`Value of '${argList[i].arg}'  required`);
```
kaustubh-karkare: throw new Error ...

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  if (tempObj.remote === true && tempObj.local === true) {
+    return new Error(`Remote and Local arguments cannot be true at the same time`);
+  }
+  return true;
+}
+
+function printParsedArgs(parsedObj) {
```
kaustubh-karkare: Eliminate this. Use console.log directly.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,108 @@
...
+function addArgType(argTypeObj) {
+  argTypeConditions[argTypeObj.name] = argTypeObj.condition;
+  argDataType[argTypeObj.name] = argTypeObj.dataType;
+  argTypeCaster[argTypeObj.dataType] = argTypeObj.castFunction;
+}
+
+
+function argGroupBuilder(argObj) {
+  var ii;
+  for(ii = 0; ii < argObj.argList.length; ii++) {
```
kaustubh-karkare: According to style guide, there needs to be a space after `for`.
Also, not sure why you didn't just do `var ii = 0;`, instead of declaring it separately.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,108 @@
...
+addArgType({ name: '+Int', condition: function(argValue) { return (!isNaN(argValue) && argValue * 1 >= 0); }, dataType: 'Number', castFunction: function(argValue) { return Number(argValue); } });
+
+// add new arguments here in the format argBuilder({key: 'argKey', arg: argName, type: 'argType', isRequired: true/false, position: index});
+
+argGroupBuilder({ groupId: 1, isRequired: false, argList: [ { key: 'command', arg: null, type: 'String', position: 0 } ] });
+argGroupBuilder({ groupId: 2, isRequired: false, argList: [ { key: 'subcommand', arg: null, type: 'String', position: 1 } ] });
+argGroupBuilder({ groupId: 3, isRequired: true, argList: [ { key: 'key', arg: '--key', type: '+Int', position: null } ] });
+argGroupBuilder({ groupId: 4, isRequired: false, argList: [ { key: 'name', arg: '--name', type: 'String', position: null } ] });
+argGroupBuilder({ groupId: 5, isRequired: false, argList: [ { key: 'verbrose', arg: '-v', type: 'Boolean', position: null } ] });
+argGroupBuilder({ groupId: 6, isRequired: false, argList: [ { key: 'local', arg: '--local', type: 'Boolean', position: null }, { key: 'remote', arg: '--remote', type: 'Boolean', position: null } ] });
```
kaustubh-karkare: Uh ... this API is rather inconvenient to use. The user has to provide these complex objects with multiple keys, and they wouldn't know if they made any mistakes until they actually run your `argParser` method, and maybe not even then. There is a "fail fast" philosophy, suggesting that you want errors to be generated ASAP.

At this point, I'd suggest looking at the API provided by https://www.npmjs.com/package/commander (or other examples) to see how simple they are to use, and compare what you have with them. After that, can you improve your API?

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,108 @@
...
+      }
+    }
+    if (count > 1) {
+      throw new Error(`Multiple arguments of group same as argument '${groupList[ii].argList[0].arg}' present.`);
+    }
+  }
+return true;
+}
+
+function argParser(inputArgs) {
```
kaustubh-karkare: This is a better name for a class, not a static function.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,108 @@
...
+    if (groupList[ii].isRequired === true) {
+      if (count === 0) {
+        throw new Error(`At least one argument from group of argument '${groupList[ii].argList[0].arg}' required.`);
+      }
+    }
+    if (count > 1) {
+      throw new Error(`Multiple arguments of group same as argument '${groupList[ii].argList[0].arg}' present.`);
+    }
+  }
+return true;
```
kaustubh-karkare: indenting issue.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,145 @@
...
+function addArgType (argTypeObj) {
+  argTypeFunctions[argTypeObj.name] = argTypeObj.typeFunction;
+  argDataType[argTypeObj.name] = argTypeObj.dataType;
+}
+
+function createGroup(groupName) {
+  if (!groupName) {
+    throw new Error(`Group name required while creating new group`);
+  }
+  let newGroupObj = {
```
kaustubh-karkare: This can be `const`, instead of `let`.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,145 @@
...
+  return newGroupObj;
+}
+
+function addArg(argObj) {
+  if (argObj.key === null) {
+    throw new Error(`Key of argument '${argObj.arg}' cannot be null`);
+  }
+  if (!(argObj.type in argTypeFunctions)) {
+    throw new Error(`Invalid type of argument '${argObj.arg}'`);
+  }
```
kaustubh-karkare: This sections looks duplicated.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,145 @@
...
+    isRequired: false,
+    argList: [],
+    addGroupArg: function(argObj) {
+      if (argObj.key === null) {
+        throw new Error(`Key of argument '${argObj.arg}' cannot be null`);
+      }
+      if (!(argObj.type in argTypeFunctions)) {
+        throw new Error(`Invalid type of argument '${argObj.arg}'`);
+      }
+      for (let ii = 0; ii < groupList.length; ii++) {
```
kaustubh-karkare: Instead of looping through this, why not just add `newGroupObj` here? It should work.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,145 @@
...
+
+function addArg(argObj) {
+  if (argObj.key === null) {
+    throw new Error(`Key of argument '${argObj.arg}' cannot be null`);
+  }
+  if (!(argObj.type in argTypeFunctions)) {
+    throw new Error(`Invalid type of argument '${argObj.arg}'`);
+  }
+  let argGroupObj = { isRequired: argObj.isRequired || false, argList: [ { key: argObj.key, arg: argObj.arg, type: argObj.type, position: argObj.position } ] };
+  groupList.push(argGroupObj);
```
kaustubh-karkare: Interesting.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,145 @@
...
+      throw new Error(`Multiple arguments of group same as argument '${groupList[ii].argList[0].arg}' present.`);
+    }
+  }
+return true;
+}
+
+function parseArg() {
+  let ii, jj, kk, argValue, arg, parseObj = {};
+  outerLoop:
+    for (ii = 0; ii < inputArgs.length; ii++) {
```
kaustubh-karkare: Damn, this is dangerous. Your `inputArgs` is defined in the outer scope, but it should be an argument (given you're using it like one on line 141). I disagree with how JavaScript handles this, because of this exact reason.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,145 @@
...
+              parseObj[groupList[jj].argList[kk].key] = arg;
+              continue outerLoop;
+            }
+          }
+        }
+      } else {
+        arg = arg.split('=');
+        for (jj = 0; jj < groupList.length; jj++) {
+          for (kk = 0; kk < groupList[jj].argList.length; kk++) {
+            if (groupList[jj].argList[kk].arg === arg[0]) {
```
kaustubh-karkare: Shouldn't it be "key" here?

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,145 @@
...
+        for (jj = 0; jj < groupList.length; jj++) {
+          for (kk = 0; kk < groupList[jj].argList.length; kk++) {
+            if (groupList[jj].argList[kk].position === ii) {
+              parseObj[groupList[jj].argList[kk].key] = arg;
+              continue outerLoop;
+            }
+          }
+        }
+      } else {
+        arg = arg.split('=');
```
kaustubh-karkare: So at different point in code, this variable contains different types of things? Given the same name, this is bad naming.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,145 @@
...
+        for (jj = 0; jj < groupList.length; jj++) {
+          for (kk = 0; kk < groupList[jj].argList.length; kk++) {
+            if (groupList[jj].argList[kk].arg === arg[0]) {
+              argValue = arg[1];
+              argValue = argTypeFunctions[groupList[jj].argList[kk].type](argValue);
+              if (argValue) {
```
kaustubh-karkare: What if the boolean `false` was valid?

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,145 @@
...
+                parseObj[groupList[jj].argList[kk].key] = argValue;
+                continue outerLoop;
+              }
+            }
+          }
+        }
+      }
+      throw new Error(`Invalid argument '${inputArgs[ii]}' entered. Check Type and Key`);
```
kaustubh-karkare: "key" is a term that people who read your code understand. If they haven't, it is confusing.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,147 @@
...
+
+function createGroup(groupName) {
+  if (!groupName) {
+    throw new Error(`Group name required while creating new group`);
+  }
+  const newGroupObj = {
+    name: groupName,
+    isRequired: false,
+    argList: [],
+    addGroupArg: function(argObj) {
```
kaustubh-karkare: For simple functions like this, consider using the arrow syntax.
```
addGroupArg: argObj => addArg(argObj, this),

File: solutions/paritosh-sharma/week-1/problem-1.2.js [hunk 52]
```diff
@@ -0,0 +1,147 @@
...
+
+function addArg(argObj, groupObj) {
+  if (argObj.argName === null) {
+    throw new Error(`argName of argument '${argObj.arg}' cannot be null`);
+  }
+  if (!(argObj.type in argTypeFunctions)) {
+    throw new Error(`Invalid type of argument '${argObj.arg}'`);
+  }
+  if (groupObj) {
+    for (let ii = groupList.length-1; ii >= 0; ii--) {
```
kaustubh-karkare: `groupList.forEach(...)` is easier to read, since you dont have to deal with index.

solutions/paritosh-sharma/week-1/problem-1.2.js
kaustubh-karkare: This is still a problem.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+return true;
+}
+
+//main arggument parser funciton
+
+function parseArg(inputArgs) {
+  let ii, jj, kk, argValue, inputArg, parseObj = {};
+  outerLoop:
+    for (ii = 0; ii < inputArgs.length; ii++) {
```
kaustubh-karkare: Yeah, you really want to use `forEach` to avoid the indices, which make things harder to understand.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,154 @@
...
+    let count = 0;
+
+    argGroup.argList.forEach((arg) => {
+      if (arg.argName in parseObj) {
+        count += 1;
+      }
+    });
+    if (argGroup.isRequired === true) {
+      if (count === 0) {
+        throw new Error(`At least one argument from group of argument '${ argGroup.argList[0].argKey }' required.`);
```
kaustubh-karkare: This error message could be confusing if you didn't use any groups.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,154 @@
...
+    inputArg = inputArg.split('=');
+
+    if (inputArg[0].indexOf('-') !== 0) {
+      argParsed = groupList.some((argGroup) => {
+        return argGroup.argList.some((arg) => {
+          if (arg.position === index) {
+            parseObj[arg.argName] = inputArg[0];
+            argParsed = true;
+            return true;
+          } else return false;
```
kaustubh-karkare: Always use blocks for all composite statements like if/else/while/for/etc.
```
} else {
  return false;
}
```
Applies below too.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,154 @@
...
+  inputArgs.forEach((inputArg, index) => {
+    let argParsed = false;
+    inputArg = inputArg.split('=');
+
+    if (inputArg[0].indexOf('-') !== 0) {
+      argParsed = groupList.some((argGroup) => {
+        return argGroup.argList.some((arg) => {
+          if (arg.position === index) {
+            parseObj[arg.argName] = inputArg[0];
+            argParsed = true;
```
kaustubh-karkare: You dont need this.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,154 @@
...
+            return true;
+          } else return false;
+        });
+      });
+    } else {
+      argParsed = groupList.some((argGroup) => {
+        return argGroup.argList.some((arg) =>{
```
kaustubh-karkare: `.some(arg => {`

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,154 @@
...
+
+  inputArgs.forEach((inputArg, index) => {
+    let argParsed = false;
+    inputArg = inputArg.split('=');
+
+    if (inputArg[0].indexOf('-') !== 0) {
+      argParsed = groupList.some((argGroup) => {
+        return argGroup.argList.some((arg) => {
+          if (arg.position === index) {
+            parseObj[arg.argName] = inputArg[0];
```
kaustubh-karkare: Use the argTypeFunctions here?

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,166 @@
...
+  if (argObj.argName === null) {
+    throw new Error(`argName of argument '${ argObj.argKey }' cannot be null`);
+  }
+
+  if (!(argObj.type in argTypeFunctions)) {
+    throw new Error(`Invalid type of argument '${ argObj.argKey }'`);
+  }
+
+  if (groupObj) {
+    groupList.some(argGroup => {
```
kaustubh-karkare: You're not using the result of the `some` call. What if you dont find the group?

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,166 @@
...
+      if (arg.argName in parseObj) {
+        count += 1;
+      }
+    });
+    if (argGroup.isRequired === true) {
+      if (count === 0) {
+        if (argGroup.argList.length === 1) {
+          throw new Error(`Argument '${argGroup.argList[0].argKey}' required`);
+        } else {
+          throw new Error(`At least one argument from group of argument '${ argGroup.argList[0].argKey }' required.`);
```
kaustubh-karkare: Much better.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,166 @@
...
+        }
+      }
+    }
+
+    if (count > 1) {
+      throw new Error(`Multiple arguments of group same as argument '${ argGroup.argList[0].argKey }' present.`);
```
kaustubh-karkare: I'm wondering if the group name is more useful than argKey.

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,166 @@
...
+    }
+  });
+
+  return true;
+}
+
+// main argument parser function
+function parseArg(inputArgs) {
+  let parseObj = {};
```
kaustubh-karkare: s/let/const

solutions/paritosh-sharma/week-1/problem-1.2.js
```diff
@@ -0,0 +1,166 @@
...
+
+  inputArgs.forEach((inputArg, index) => {
+    let argParsed = false;
+    let argValue = null;
+    let keyLessArg;
+    inputArg = inputArg.split('=');
```
kaustubh-karkare: In general, it isn't a good idea to store different types of information in the same variable. You will confused readers about whether this contains an array or a string.

pari08tosh:
Submitting the 10th update to problem 1.2 according to suggestions in previous PR. Have made the following changes:-
1. Improved variable names and type to reduce confusion.

PR #9, opened 2017-03-14
Solution to Assignment-1.2

iamashank:
Please review.

### DIFF
--- a/solutions/ashank-anshuman/assignment-1.2.py
+++ b/solutions/ashank-anshuman/assignment-1.2.py
@@ -1,0 +1,127 @@
+from sys import argv
+import re
+
+group = []
+arguments = {}
+shortnotation = {}
+comp_args = []
+
+
+def add_argument(*args, **kwargs):
+    # Format (arg name,short name(if any),type = int(if required),value = True(if required))
+    # Example : add_argument("--key",type = int,value = True)
+    # Creates a new argument
+    arguments[str(args[0]).lower()] = {}
+    arguments[str(args[0])]["value"]=False
+    if(len(args)==2):
+        arguments[str(args[0])]["arg_shortname"] = str(args[1].lower())
+        shortnotation[str(args[1].lower())] = str(args[0]).lower() 
+    for k,v in kwargs.items():
+        if(k == 'type'):
+            arguments[str(args[0])][type] = v
+        elif(k == "value"):
+            arguments[str(args[0])]["value"] = v
+
+
+def group_args(*args):
+    # Add conflicting arguments to common group
+    group.append([])
+    for arg in args:
+        group[len(group)-1].append(arg)
+
+def add_compargs(*args):
+    # Add arguments which must be present
+    for arg in args:
+        comp_args.append(arg)
+    
+
+def show_help():
+    # Shows usage
+    print ("Usage : ",end="")
+    for args in arguments:
+        if (args in comp_args):
+            print(args, end="  ")
+        else:
+            print ("["+args+"]" , end="  ")
+    print()
+
+add_argument("--help","-h")
+
+def parser():
+
+    # Test Arguments
+    add_argument("--key",type = int,value = True)
+    add_argument("--name",value = True)
+    add_argument("--verbrose","-v")
+    add_argument("alpha")
+    add_argument("beta")
+    add_argument("--local")
+    add_argument("--remote")
+    group_args("--local","--remote")
+
+    # Compulsory arguments
+    add_compargs("--key","--name")
+
+    counter = 0            
+    json_output = {}
+    for args in argv:
+        if(args != "assignment-1.2.py"):
+            arg = args.split('=')
+            if(arg[0].lower() in arguments or arg[0].lower() in shortnotation):         # Check if argument exists
+                if(arg[0].lower()=="--help" or arg[0].lower()=="-h"):                   # Shows help
+                    show_help()
+                    exit(0)
+                if(arg[0].lower() in shortnotation):                                    # Gets name of shortnotation
+                    arg[0] = shortnotation[arg[0].lower()]
+                if(arguments[arg[0]]["value"]):
+                    if(len(arg)==2):
+                        if(type in arguments[arg[0]]):                                  # Checks if value type of argument is correct
+                            try:
+                                temp = arguments[arg[0]][type](arg[1])
+                            except:
+                                print ("Error: command '",arg[0],"' must have a ",
+                                       str(arguments[arg[0]][type])," argument.")
+                                exit(0)
+                    else:
+                        print ("Error: argument '",arg[0],"' is missing its value.")
+                        exit(0)
+                    json_output[arg[0]] = arg[1]
+                else:
+                    if(len(arg)==2):
+                        print ("Error: argument '",arg[0],"' does not require any value.")
+                        exit(0)
+                    if(arg[0][0]=='-'):                                                 # Adds argument to json type object (dictionary in this case)
+                        json_output[arg[0]] = "True"
+                    else:
+                        json_output["sub"*counter+"command"] = arg[0]
+                        counter += 1
+                    
+            else:
+                print ("Error: argument '",arg[0],"' not found.")
+                exit(0)
+
+    for arg in comp_args:                                                               # Checks for presence of required arguments
+        if(arg not in json_output):
+            print("Error: '",arg,"' argument is required, but missing from input.")
+            exit(0)
+
+    for g in group:                                                                     # Checks if conflicting arguments are present
+        count = 0
+        temp_list = []
+        for arg in json_output:
+            if(arg in g):
+                temp_list.append(arg)
+                count += 1
+            if(count>1):
+                print ("Error: arguments '",temp_list[0],"' and '",
+                       temp_list[1],"' cannot be used together.")
+                exit(0)
+
+    for arg in json_output:                                                             # Formats argument name for returing as json type
+        temp = re.sub('[-]', '', arg)
+        json_output[temp] = json_output.pop(arg)
+
+    print (json_output)
+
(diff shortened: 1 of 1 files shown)


solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,131 @@
...
+
+    # Test Arguments
+    add_argument("--key",type = int,value = True)
+    add_argument("--name",value = True)
+    add_argument("--verbrose","-v")
+    add_argument("alpha")
+    add_argument("beta")
+    add_argument("--local")
+    add_argument("--remote")
+    group_args("--local","--remote")
```
kaustubh-karkare: What if group arguments could be provided with a value?

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,131 @@
...
+
+    # Compulsory arguments
+    add_compargs("--key","--name")
+
+    counter = 0            
+    json_output = {}
+    for args in argv:
+        if(args != "assignment-1.2.py"):
```
kaustubh-karkare: This is inelegant. Figure out how to avoid it?

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,131 @@
+from sys import argv
+import re
+
+group = []
+arguments = {}
+shortnotation = {}
+comp_args = []
+
+
+def add_argument(*args, **kwargs):
```
kaustubh-karkare: Instead of accepting random kwargs as input, why not explicitly list them out to avoid issues?

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,131 @@
...
+    # Creates a new argument
+    # Format (arg name,short name(if any),
+    #           type = int(if a specific data type is required),
+    #           value = True(if a value is required))
+    # Example : add_argument("--key",type = int,value = True)
+    if(len(args)>2):
+        raise ValueError('More arguments were supplied than allowed.')
+    arguments[str(args[0]).lower()] = {}
+    arguments[str(args[0])]["value"]=False
```
kaustubh-karkare: So you have spaces around `=` in the previous line, but not this one. Please be consistent.

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,131 @@
...
+comp_args = []
+
+
+def add_argument(*args, **kwargs):
+    # Creates a new argument
+    # Format (arg name,short name(if any),
+    #           type = int(if a specific data type is required),
+    #           value = True(if a value is required))
+    # Example : add_argument("--key",type = int,value = True)
+    if(len(args)>2):
```
kaustubh-karkare:
```
if len(args) > 2:
```
No need for parenthesis, but space after `if`.

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,131 @@
...
+            print(args, end="  ")
+        else:
+            print ("["+args+"]" , end="  ")
+    print()
+
+add_argument("--help","-h")
+
+def parser():
+
+    # Test Arguments
```
kaustubh-karkare: For clean abstraction, please separate your input/output from your parsing logic. Try and make your code generic.

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,131 @@
...
+            if(count>1):
+                print ("Error: arguments '",temp_list[0],"' and '",
+                       temp_list[1],"' cannot be used together.")
+                exit(0)
+
+    for arg in json_output:                                                             # Formats argument name for returing as json type
+        temp = re.sub('[-]', '', arg)
+        json_output[temp] = json_output.pop(arg)
+
+    print (json_output)
```
kaustubh-karkare: No need for space after `print`.

solutions/ashank-anshuman/assignment-1.2.py
iamashank: It works for arguments with value too.
I added these just for the sake of example/testing.

solutions/ashank-anshuman/assignment-1.2.py
iamashank: Actually, I wanted to implement it the way 'Argparse' module does.
`parser.add_argument("square", help="display a square of a given number",
                    type=int)`
Here, the user may or may not add the 'type' argument depending on their need. To implement this I used kwargs.
I'll explicitly list them out in next commit.

iamashank:
I've made the requested changes in the code. Please review.

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,138 @@
...
+if __name__ == '__main__':
+
+    # Test Arguments
+    add_argument("--key",None,int,True)
+    add_argument("--name",None,None,True)
+    add_argument("--verbrose","-v",None,False)
+    add_argument("alpha",None,None,False)
+    add_argument("beta",None,None,False)
+    add_argument("--local",None,None,False)
+    add_argument("--remote",None,None,False)
```
kaustubh-karkare: Your API should be intuitive. Right now, since I've started reading from your main section, I have no clue what these `None`s and boolean values are, which means that it isn't intuitive yet.

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,138 @@
+from sys import argv
+import re
+
+group = []
+arguments = {}
+shortnotation = {}
+comp_args = []
```
kaustubh-karkare: Global variables like this, in general, aren't good design. Use classes and then store this data inside it.

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,138 @@
...
+        return
+    arguments[str(argname).lower()] = {}
+    if short_argname:
+        arguments[str(argname)]["short_argname"] = str(short_argname).lower()
+        shortnotation[str(short_argname).lower()] = str(argname).lower()
+    if datatype is None and value is True:
+        arguments[str(argname).lower()]["type"] = str
+    elif value is True:
+        arguments[str(argname).lower()]["type"] = datatype
+    arguments[str(argname).lower()]["value"] = value
```
kaustubh-karkare: Why not use classes here too? Instead of these dicts which all have the same structure.

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def group_args(*args):
+    # Add conflicting arguments to a common group
+    group.append([])
+    for arg in args:
+        group[len(group)-1].append(arg)
+
+def add_compargs(*args):
```
kaustubh-karkare: Try and make names as obvious as possible. `comp` can refer to complete / comparison / etc. Compulsary is only one of the interpretations. `add_required_args` is a much better name. But I don't think that you should have this method anyway.

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,138 @@
...
+    print()
+
+add_argument("--help","-h",None,False)
+
+def parser():
+    counter = 0            
+    json_output = {}
+    for i in range(1,len(argv)):
+        arg = argv[i].split('=')
+        if arg[0].lower() in arguments or arg[0].lower() in shortnotation:         # Checks if argument exists
```
kaustubh-karkare: Why not convert to `.lower()` once and then use that result? Instead of calling it 5 times.

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,138 @@
+from sys import argv
+import re
+
+group = []
+arguments = {}
+shortnotation = {}
+comp_args = []
+
+def add_argument(argname,short_argname=None,datatype=str,value=False):
```
kaustubh-karkare: `value` isn't a good name. `can_have_value` is much clearer. You can see how it is possible to misinterpret the original name.

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,138 @@
...
+            if arg[0].lower() == "--help" or arg[0].lower() == "-h":               # Shows help
+                show_help()
+                exit(0)
+            if arg[0].lower() in shortnotation:                                    # Gets name of shortnotation
+                arg[0] = shortnotation[arg[0].lower()]
+            if arguments[arg[0]]["value"] is True:
+                if len(arg) == 2:
+                    if "type" in arguments[arg[0]]:                                # Checks if value type of argument is correct
+                        try:
+                            temp = arguments[arg[0]]["type"](arg[1])
```
kaustubh-karkare: In general, `temp` is always a bad name. You want to make it very obvious what the variable contains.

solutions/ashank-anshuman/assignment-1.2.py
```diff
@@ -0,0 +1,138 @@
...
+        else:
+            print("Error: argument '",arg[0],"' not found.")
+            exit(0)
+
+    for arg in comp_args:                                                            # Checks for presence of required arguments
+        if arg not in json_output:
+            print("Error: '",arg,"' argument is required, but missing from input.")
+            exit(0)
+
+    for g in group:                                                                  # Checks if conflicting arguments are present
```
kaustubh-karkare: Looks like the line length exceeds limits. In order to facilitate easy reviews, the convention is to limit it to 80 characters or something similar.

PR #10, opened 2017-03-15
assignment-1.2.cpp

humanoid2795:
Kindly review.

### DIFF
--- a/solutions/anand-kumar/assignment-1.2.cpp
+++ b/solutions/anand-kumar/assignment-1.2.cpp
@@ -1,69 +1,119 @@
 #include <bits/stdc++.h>
 using namespace std;
-// Global vector storing the JSON converted data.
-vector<pair<string, string> > json;
-// Attributes of the parameter.
-struct parameter_node {
-  string parameter_name;
-  string parameter_shorthand;
-  string parameter_data_type;
-  pair<bool, string> parameter_data;
-  set<string> parameter_exclusive_list;
-  string parameter_help;
+
+class parser {
+ private:
+  // Attributes of the parameter.
+  struct parameter_node {
+    string name;
+    string shorthand;
+    string data_type;
+    string value;
+    bool is_data_required;
+    set<string> exclusive_list;
+    string help_instruction;
+  };
+  int number_of_commands_recieved;
+  vector<pair<string, string> > json;
+
+ public:
+  // Constructor for specifying instances at the time of creation
+  parser() {
+    number_of_commands_recieved = 0;
+  } 
+  // structure to store at any instance the parameter and the argument under
+  // consideration
+
+  struct parameter_and_argument_node {
+    bool is_argument_available;
+    string parameter_name;
+    string parameter_argument;
+  } parameter_and_argument;
+  
+  // Vector of object of struct parameter_node to store all parameters.
+
+  vector<struct parameter_node> parameter_list;
+
+  void argument_assignment(
+    string name_argument, string data_type_argument, string argument_existence,
+    string shorthand_argument);
+  int find_index(string parameter);
+  void create_exclusive_parameters_set(string ex_list[], int size_of_ex_list);
+  void extract_parameter_and_argument(string current_parameter);
+  bool is_integer(string num);
+  bool is_floating(string num);
+  bool is_boolean(string num);
+  void if_valid_data_type(int parameter_index, string current_argument);
+  void parameter_and_argument_parsing(string current_parameter);
+  void check_exclusivity_of_json();
+  void help_function();
+  void display_without_hyphen(string parameter);
+  void display_json();
 };
-void parameter_declaration(vector<struct parameter_node> &parameter, string
-    p_name, string p_shorthand, string p_data_type, string p_help) {
+
+void parser::argument_assignment(
+    string name_argument, string data_type_argument, string argument_existence,
+    string shorthand_argument) {
+
+  // Temporary object ob creation.
+
   struct parameter_node ob;
-  ob.parameter_name = p_name;
-  ob.parameter_shorthand = p_shorthand;
-  ob.parameter_data_type = p_data_type;
-  ob.parameter_help = p_help;
-  ob.parameter_data.first = false;
-  ob.parameter_data.second = '\n';
-  parameter.push_back(ob);
+  ob.name = name_argument;
+  ob.shorthand = shorthand_argument;
+  ob.data_type = data_type_argument;
+  if (argument_existence == "compulsory") {
+    ob.is_data_required = true;
+  }
+  else {
+    ob.is_data_required = false;
+  }
+  if (ob.is_data_required == true) {
+    ob.help_instruction = "Compulsory argument";
+  }
+  else {
+    ob.help_instruction = "Optional argument";
+  }
+  ob.help_instruction += " of type " + data_type_argument;
+  parameter_list.push_back(ob);
 }
-// Index of a parameter in the structure object vector
-int find_index(vector<struct parameter_node> parameter, string dummy) {
-  for (int i = 0; i < parameter.size(); ++i) {
+
+// Index of the parameter in the parameter_list
+int parser::find_index(string current_parameter) {
+  if (current_parameter.length() == 0) {
+    return -1;
+  }
+  for (int i = 0; i < parameter_list.size(); ++i) {
     if (
-      parameter[i].parameter_name == dummy || 
-      parameter[i].parameter_shorthand == dummy) {
+      parameter_list[i].name == current_parameter ||
+      parameter_list[i].shorthand == current_parameter ) {
       return i;
     }
   }
   return -1;
 }
-void create_exclusive_parameters(vector<struct parameter_node> parameter,
-    string ex_list[], int s) {
-  for (int i = 0; i < s; ++i) {
-    int index = find_index(parameter, ex_list[i]);
+
+void parser::create_exclusive_parameters_set
+    (string ex_list[], int size_of_ex_list) {
+  if (size_of_ex_list == 0) {
+    return;
+  }
+  for (int i = 0; i < size_of_ex_list; ++i) {
+    int index = find_index(ex_list[i]);
     if (index == -1) {
-      cout << "Undefined parameter\n" ;
+      display_without_hyphen(ex_list[i]);
+      cout << " is an undefined parameter\n" ;
       exit(1);
     }
-    for (int j = 0; j < s; ++j) {
+    for (int j = 0; j < size_of_ex_list; ++j) {
       if (i == j) {
         continue;
       }
-    parameter[index].parameter_exclusive_list.insert(ex_list[j]);
+      parameter_list[index].exclusive_list.insert(ex_list[j]);
     }
   }
 }
-// Extract the string content without '-' at any index of argv
-string extract_string(char *argv[], int index) {
-  int i = 0;
-  string curr_temp = argv[index];
-  while (curr_temp[i] == '-') {
-    ++i;
-  }
-  string dummy;
-  while (curr_temp[i] != '\n' && i < curr_temp.length() && curr_temp[i] != '=') {
-    dummy +=  curr_temp[i];
-    ++i;
-  }
-  return dummy;
-}
-bool is_integer(string num) {
+
+bool parser::is_integer(string num) {
   bool is_integer = true;
   for (int i = 0; i < num.length() && is_integer; ++i) {
     if (!(num[i] >= '0' && num[i] <= '9')) {
@@ -72,7 +122,8 @@ bool is_integer(string num) {
   }
   return is_integer;
 }
-bool is_floating(string num) {
+
+bool parser::is_floating(string num) {
   bool is_float = true;
   int decimal = 0;
   for (int i = 0; i < num.length() && is_float; ++i) {
@@ -89,13 +140,14 @@ bool is_floating(string num) {
   }
   return is_float;
 }
-bool is_boolean(string num) {
+
+bool parser::is_boolean(string num) {
   if (
     num == "true" ||
     num == "TRUE" ||
     num == "True" ||
     num == "false" ||
-    num == "FALSE" || 
+    num == "FALSE" ||
     num == "False") {
     return true;
   }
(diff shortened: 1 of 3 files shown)


solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,277 @@
...
+			if (i == j)
+				continue;
+			parameter[index].parameter_exclusive_list.insert(ex_list[j]);
+		}
+	}
+}
+//  Extract the string content without '-' at any index of argv
+string extract_string(char *argv[], int index) {
+  int i = 0;
+	string curr_temp = argv[index];
```
kaustubh-karkare: The previous 2 lines have inconsistent indenting.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,277 @@
...
+	while (curr_temp[i] == '-')
+		++i;
```
kaustubh-karkare: Don't they recommend that all `if`, `for`, `while`, etc conditions have braces?

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,277 @@
...
+			is_float = false;
+	}
+	if (decimal > 1)
+		is_float = false;
+	return is_float;
+}
+bool is_boolean(string num)
+{
+	if (num == "true" || num == "TRUE" || num == "True" || num == "false" ||
+		    num == "FALSE" || num == "False")
```
kaustubh-karkare: This is acceptable, I guess, but I usually prefer:
```
if (
  num == "true" ||
  num == "TRUE" ||
  ...
) {
```

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,277 @@
...
+    char *argv[], int &index) {
+	string dummy = extract_string(argv, index);
+	int option_index = find_index(parameter, dummy);
+	if (option_index == -1) {
+		cout << dummy << " ";
+		cout << "parameter does not exist\n";
+		exit(1);
+	}
+	if (parameter[option_index].parameter_data.first == false) {
+			string data_dummy;
```
kaustubh-karkare: Why does this have 3 tabs? I would expect 2.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,277 @@
...
+	}
+	int counter = 0;
+	while (index < argc) {
+		string curr_temp = argv[index];
+		if (curr_temp == "-h" || curr_temp == "--help") {
+				help_function(parameter);
+				++index;
+				continue;
+		}
+		//	If a parameter.
```
kaustubh-karkare: There should be a space after `//`, not a tab.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,277 @@
...
+  ob.parameter_data_type = p_data_type;
+  ob.parameter_help = p_help;
+  ob.parameter_data.first = false;
+  ob.parameter_data.second = '\n';
+  parameter.push_back(ob);
+}
+//  Index of a parameter in the structure object vector
+int find_index(vector<struct parameter_node> parameter, string dummy) {
+  for (int i = 0; i < parameter.size(); ++i) {
+	  if (parameter[i].parameter_name == dummy ||
```
kaustubh-karkare: You're using tabs for indenting below and spaces above. And this line seems to have both, which is bad.
Based on how this looks, I'd recommend consistently using 2 spaces for each level of indenting.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,291 @@
...
+      }
+      // Sub - command
+      else {
+        for (int i = 0; i < counter; ++i) {
+          id += "sub-";
+        }
+      id += "command";
+      json.push_back(make_pair(id, dummy));
+      }
+    ++index;
```
kaustubh-karkare: Incorrect indenting. Bunch of other places too.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,291 @@
...
+    cout << "No parameter or command given\n";
+    exit(1);
+  }
+  int counter = 0;
+  while (index < argc) {
+    string curr_temp = argv[index];
+    if (curr_temp == "-h" || curr_temp == "--help") {
+      help_function(parameter);
+      ++index;
+      continue;
```
kaustubh-karkare: In general, if you're displaying help, you dont do anything else.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,291 @@
...
+  // Enter the commands here.
+  
+  int index = 1;
+  if (argc < 2) {
+    cout << "No parameter or command given\n";
+    exit(1);
+  }
+  int counter = 0;
+  while (index < argc) {
+    string curr_temp = argv[index];
```
kaustubh-karkare: `curr_temp` isn't a good name. How about `current_argument`?

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,291 @@
...
+    if (curr_temp == "-h" || curr_temp == "--help") {
```
kaustubh-karkare: Why isn't this a parameter? Why special case it?

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,291 @@
...
+      ++index;
+      continue;
+    }
+    // If a parameter.
+    if (curr_temp[0] == '-') {
+      parameter_argument(parameter, argc, argv, index);
+    }
+    else {	
+      // If a command or a sub command.
+      string dummy = extract_string(argv, index);
```
kaustubh-karkare: This line shows that `extract_string` isn't a good name. `extract_argument_name` maybe.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,291 @@
...
+  create_exclusive_parameters(parameter, ex_list, 2);
+  // Enter mutually exclusive elements.
+  // ...
+
+  // Enter the command list for the parser
+  vector<string> command_list;
+  // Push back the command in the command list.
+  // Test data
+  command_list.push_back("alpha");
+  command_list.push_back("beta");
```
kaustubh-karkare: The values aren't fixed like this.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,291 @@
...
+// To check for exclusivity of parameters.
+void exclusivity(vector<struct parameter_node> parameter) {
+  for (int i = 0; i < json.size(); ++i) {
+    string dummy = json[i].first;
+    int index = find_index(parameter, dummy);
+    for (int j = 0; j < json.size(); ++j) {
+      if (i == j) {
+        continue;
+      }
+      set <string> :: iterator it =
```
kaustubh-karkare: Spacing seems off.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+  // Function definition:-
+  // void argument_assignment(
+  // string name_argument, string data_type_argument,
+  // string argument_existence, string shorthand_argument);
+  // Data types supported are "int", "float", "bool", "string", "none"
+  // The data type is case sensitive and by default all parameters are string.
+  // Argument existence states are "optional" and "compulsory"
+  // Test data
+
+  parser_object.argument_assignment("--help", "none", "optional", "-h");
```
kaustubh-karkare: Instead of strings like "int", "bool", "optional" and "compulsory" (which could have typos), use enums here.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+  parser_object.argument_assignment("--key", "int", "compulsory", "\n");
```
kaustubh-karkare: Why do I need to provide the `\n` as an argument? That is inelegant.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+  cout << "\": " << "\"" << json[json.size() - 1].second << "\"";
+  cout << "}\n";
+}
+
+
+int main(int argc, char *argv[]) {
+  // Test class parser object
+  class parser parser_object;
+
+  // Input the parameter with the argument;
```
kaustubh-karkare: Ideally, I shouldn't need these comments. The function signature should be enough.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+  // Input the exclusive list of elements
+  // Enter the list of mutually exclusive elements in an array of string.
+  // Also define the number of elements pushed in the array.
+  // Function definition:-
+  // void create_exclusive_parameters_set(string ex_list[], int size);
+  // Test data
+
+  string ex_list[] = {"--local", "--remote"};
+  parser_object.create_exclusive_parameters_set(ex_list,
+      sizeof(ex_list) / sizeof(ex_list[0]));
```
kaustubh-karkare: Interesting. I guess this works.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+  // Enter mutually exclusive elements.
+  // ...
+
+  int index = 1;
+  if (argc < 2) {
+    cout << "No parameter or command given\n";
```
kaustubh-karkare: You cant assume that this is a bad thing. What if I created a parser that only had optional arguments? So nothing provided was a valid case?

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+    exit(1);
+  }
+  while (index < argc) {
+    string current_parameter = argv[index];
+    parser_object.parameter_and_argument_parsing(current_parameter);
+    ++index;
+  }
+  parser_object.check_exclusivity_of_json();
+  parser_object.display_json();
```
kaustubh-karkare: This looks hard to use. Imagine if this was a library: would you want to use it for your programs? Wouldn't you prefer something simpler? You want to hide these details inside a `Parser::parse()` method, which is the only thing people should need to call.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+    num == "FALSE" ||
+    num == "False") {
+    return true;
+  }
+  else {
+    return false;
+  }
+}
+
+void parser::if_valid_data_type(int parameter_index, string current_argument) {
```
kaustubh-karkare: - `ensure_valid_data_type` is a better name. `if` sounds like you're going to return a boolean yes or no.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+    valid_data_type = is_boolean(current_argument);
+  }
+  else {
+    // By default all are considered to be string.
+    valid_data_type = true;
+  }
+  if (valid_data_type == false) {
+    display_without_hyphen(parameter_list[parameter_index].name);
+    cout << " has an invalid data\n";
+    exit(1);
```
kaustubh-karkare: Instead of exiting abruptly, prefer throwing exceptions. That gives the calling program flexibility to continue with default options if necessary, or ask the user to fix only the ones that are invalid.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+      parameter_list[index].exclusive_list.insert(ex_list[j]);
+    }
+  }
+}
+
+bool parser::is_integer(string num) {
+  bool is_integer = true;
+  for (int i = 0; i < num.length() && is_integer; ++i) {
+    if (!(num[i] >= '0' && num[i] <= '9')) {
+      is_integer = false;
```
kaustubh-karkare: Just return `false` as soon as you figure it out. You wouldn't need your `is_integer` flag then.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+bool parser::is_floating(string num) {
+  bool is_float = true;
+  int decimal = 0;
+  for (int i = 0; i < num.length() && is_float; ++i) {
+    if (num[i] == '.') {
+      ++decimal;
+      continue;
+    }
+    if (!(num[i] >= '0' && num[i] <= '9')) {
+      is_float = false;
```
kaustubh-karkare: Same as above. Return false as soon as you know something is wrong, which will allow you to eliminate  the flag.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+      parameter_list[i].name == current_parameter ||
+      parameter_list[i].shorthand == current_parameter ) {
+      return i;
+    }
+  }
+  return -1;
+}
+
+void parser::create_exclusive_parameters_set
+    (string ex_list[], int size_of_ex_list) {
```
kaustubh-karkare: Better formatting:
```
void parser::create_exclusive_parameters_set(
  string ex_list[],
  int size_of_ex_list,
) {
```
Also, avoid abbreviations like `ex`.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+  if (size_of_ex_list == 0) {
+    return;
```
kaustubh-karkare: Why bother? The loop will not run anyway.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+  }
+  for (int i = 0; i < size_of_ex_list; ++i) {
+    int index = find_index(ex_list[i]);
```
kaustubh-karkare: This is a dangerously named method: sounds like part of some array library, which was what I though initially.
Could you update the name to make it more useful? `find_index_in_paramters_list` or something to make it clearer.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+  }
+}
+
+void parser::extract_parameter_and_argument(
+    string current_parameter) {
+  int i = 0;
+  string current_extracted_parameter;
+
+  // If the input is of the form --key=1234
+  // Extract only parameter i.e --key
```
kaustubh-karkare: Without your example, there was no way to tell if your use of the word parameter referred to "--key" or "1234".
Which means .. that the name parameter here is practically useless. Prefer something like "key" or "value", which doesn't need an example. Or doesn't even need a comment.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+  while (i < current_parameter.length() && current_parameter[i] != '=') {
+      current_extracted_parameter +=  current_parameter[i];
+    ++i;
```
kaustubh-karkare: The last 2 lines are inconsistently indented.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+  string current_extracted_argument;
+  while (i < current_parameter.length()) {
+      current_extracted_argument += current_parameter[i];
+    ++i;
+  }
+  parameter_and_argument.parameter_name =
+      current_extracted_parameter;
+  parameter_and_argument.parameter_argument =
+      current_extracted_argument;
+  if (current_extracted_argument.length() == 0) {
```
kaustubh-karkare: There has to be a better way to do this. What if I wanted to provide an empty string as my argument?

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+    parameter_and_argument.is_argument_available = false;
+  }
+  else {
+    parameter_and_argument.is_argument_available = true;
+  }
+}
+// To check for the argument along with the parameter given by the user.
+void parser::parameter_and_argument_parsing(string current_parameter) {
```
kaustubh-karkare: This function name feels grammatically wrong. You want to describe what it will do, like `is_boolean`, but this sounds like a random statement.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+  if (current_parameter == "--help" || current_parameter == "-h") {
```
kaustubh-karkare: You shouldn't hardcode things. Treat it just like any other parameter.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+    valid_data_type = true;
+  }
+  if (valid_data_type == false) {
+    display_without_hyphen(parameter_list[parameter_index].name);
+    cout << " has an invalid data\n";
+    exit(1);
+  }
+}
+
+void parser::extract_parameter_and_argument(
```
kaustubh-karkare: Why doesn't this return `parameter_and_argument`? Avoid class properties as much as possible, since they add to complexity.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+void parser::parameter_and_argument_parsing(string current_parameter) {
+  if (current_parameter == "--help" || current_parameter == "-h") {
+    help_function();
+  }
+  extract_parameter_and_argument(current_parameter);
+  int parameter_index = find_index(parameter_and_argument.parameter_name);
+  if (parameter_index == -1) {
+    display_without_hyphen(parameter_and_argument.parameter_name);
+    cout <<" is undefined\n";
+    exit(1);
```
kaustubh-karkare: As mentioned, use exceptions.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+  parameter_list[parameter_index].value = 
+      parameter_and_argument.parameter_argument;
+  // If a command or a sub-command
+  
+  if (current_parameter.length() > 0 && current_parameter[0] != '-') {
+    string current_id_value ;
+    for (int i = 0; i < number_of_commands_recieved; ++i) {
+      current_id_value += "sub-";
+    }
+    current_id_value += "command";
```
kaustubh-karkare: Nope. `command` and `subcommand` are positional arguments that would have been given to you as input.
Basically, something like `parser_object.position_argument("command", ...)`. And then "alpha" and "beta" are just example values. I could pass in "abc" and "def", and they should still have been saved in command and subcommend, as a virtue of their position.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+    json.push_back(make_pair(current_id_value, current_parameter));
+    ++number_of_commands_recieved;
+    return;
+  }
+  if (parameter_and_argument.is_argument_available == false) {
+    if (parameter_list[parameter_index].is_data_required == true) {
```
kaustubh-karkare: Some of your names are a bit too verbose. `parameter_list[parameter_index].is_required` is just as readable as `is_data_required`, but faster to read. Similarly, `is_available` is better than `is_argument_available`. However, I'm not I like the idea of `is_available`, because you're storing more state. Which means that even a variable called "json" should not exist.

solutions/anand-kumar/assignment-1.2.cpp
```diff
@@ -0,0 +1,376 @@
...
+    if_valid_data_type(parameter_index,
+        parameter_list[parameter_index].value);
+    json.push_back(make_pair(parameter_list[parameter_index].name,
+        parameter_list[parameter_index].value));
+  }
+}
+
+// To check for check_exclusivity_of_json of parameters.
+
+void parser::check_exclusivity_of_json() {
```
kaustubh-karkare: If you look up the definition of "JSON", you will discover that it has nothing to do with this. It is just a way to serialize data.

kaustubh-karkare:
Inactive for too long.

PR #11, opened 2017-03-18
Solution to Assignment 2.1 updated 

divyanshjr:
I've tried to implement Client-Server interaction using Java. Please review and suggest changes.


solutions/divyansh-jamuaar/Assignment-2.1/Client.java
```diff
@@ -0,0 +1,64 @@
+/* Client side program to scan 2 numbers as Command Line Arguments,
+ * send them to the server, receive the result and verify it's correctness
+ */
+import java.io.*;
+import java.net.*;
+
+public class Client {
+  int self_computed_result; // To verify received result
```
kaustubh-karkare: Based on the rules about naming in section 5.2 of the [style guide](https://google.github.io/styleguide/javaguide.html), you be be using lowerCamelCase, not underscores.

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
```diff
@@ -0,0 +1,64 @@
...
+  String ip_address = "localhost"; // IP of server
+  int port = 9999;  // Port on which server is hosted
```
kaustubh-karkare: Extra space before `//`

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
```diff
@@ -0,0 +1,64 @@
...
+  Socket socket;
+  // Constructor
+  public Client() {
+    try {
+      socket = new Socket(ip_address,port);
```
kaustubh-karkare: Space after `,` (section 4.6.2)

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+    catch(Exception e) {
```
kaustubh-karkare: Put the catch on the previous line?

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
```diff
@@ -0,0 +1,64 @@
+/* Client side program to scan 2 numbers as Command Line Arguments,
+ * send them to the server, receive the result and verify it's correctness
+ */
+import java.io.*;
+import java.net.*;
+
+public class Client {
+  int self_computed_result; // To verify received result
+  String ip_address = "localhost"; // IP of server
```
kaustubh-karkare: This isn't an IP address.

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
```diff
@@ -0,0 +1,64 @@
...
+          System.out.println("Result verified");
+        }
+        else {
+          System.out.println("Incorrect result obtained");
+        }
+    }
+  }
+  // Main method
+  public static void main(String args[]) throws Exception {
+    Client client = new Client();
```
kaustubh-karkare: Isn't it better to specify the hostname / port as constructor arguments? Make things generic?

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
```diff
@@ -0,0 +1,64 @@
...
+  String ip_address = "localhost"; // IP of server
+  int port = 9999;  // Port on which server is hosted
+  Socket socket;
+  // Constructor
+  public Client() {
+    try {
+      socket = new Socket(ip_address,port);
+    }
+    catch(Exception e) {
+      System.out.println("Connection refused by server or no server found");
```
kaustubh-karkare: Prefer to throw Exceptions till the main function, and let it handle graceful termination.

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
```diff
@@ -0,0 +1,64 @@
...
+          System.out.println("Incorrect result obtained");
+        }
+    }
+  }
+  // Main method
+  public static void main(String args[]) throws Exception {
+    Client client = new Client();
+    client.computeResult(args);
+    client.sendData(args);
+    client.getData();
```
kaustubh-karkare: This method is not properly named. `getData` suggests that I'm going to receive some data as input.
However, the function returns void, and also does validation inside it.

Instead of mixing framework + application logic like this, you want to keep things separate. That way, your Client class can be used for more than exactly 1 thing.

In this case, how about if the details about the addition are isolated to your main function, and nothing else. Alternatively, don't have `main` functions in either of your classes. Create a 3rd class that uses these 2, which invokes both and ensures that communication works.

solutions/divyansh-jamuaar/Assignment-2.1/Server.java
```diff
@@ -0,0 +1,64 @@
+// Server side program to receive 2 arguments from Client and return the sum if possible
+import java.io.*;
+import java.net.*;
+
+public class Server {
+  static ServerSocket serversocket;
```
kaustubh-karkare: Use camelCase to make it easier to notice word boundaries. `serverSocket`

solutions/divyansh-jamuaar/Assignment-2.1/Server.java
```diff
@@ -0,0 +1,64 @@
...
+  static Socket socket;
+  static String[] input;
+  // Constructor
+  public Server() {
+    try {
+      serversocket = new ServerSocket(9999); // Trying to setup server on port 9999
+    }
+    catch(Exception e) {
```
kaustubh-karkare: Same as before: let `main` handle exceptions, instead of your individual methods doing it.

solutions/divyansh-jamuaar/Assignment-2.1/Server.java
```diff
@@ -0,0 +1,64 @@
...
+      System.out.println("Port not available");
+      System.exit(0);
+    }
+  }
+  public void start() throws Exception {
+    System.out.println("Server started");
+    while(true) {
```
kaustubh-karkare: Space after while.

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
```diff
@@ -0,0 +1,64 @@
...
+  public void getData() throws Exception {
+    BufferedReader br =
+    new BufferedReader(new InputStreamReader(socket.getInputStream()));
+    String input_result = br.readLine();
+    // If Server returns some Error message
+    if(input_result.startsWith("Error:")) {
+      System.out.println("Message from server side: " + input_result);
+    }
+    else {
+        System.out.println("Result received from server: " + input_result);
```
kaustubh-karkare: Instead of printing directly to stdout, you should use a Logger class that can be configured.

solutions/divyansh-jamuaar/Assignment-2.1/Server.java
```diff
@@ -0,0 +1,64 @@
...
+      System.out.println("Waiting for client to connect");
+      socket = serversocket.accept(); // Wait for client to get connected
+      System.out.println("Client connected");
+      Thread.sleep(2000); // Wait for 2 seconds
+      getInput();
+      sendOutput();
```
kaustubh-karkare: Both of these functions need better names. `getInput` sounds like it would return something. `sendOutput` sounds like it should take an argument. Try and make names and behavior intuitive.

solutions/divyansh-jamuaar/Assignment-2.1/Server.java
```diff
@@ -0,0 +1,64 @@
...
+  public static void getInput() throws Exception {
+    ObjectInputStream ois = new ObjectInputStream(socket.getInputStream());
+    input = (String[]) ois.readObject();
+  }
+  // Method to send result to client
+  public static void sendOutput() throws Exception {
+    OutputStreamWriter osw = new OutputStreamWriter(socket.getOutputStream());
+    PrintWriter out = new PrintWriter(osw);
+    // In case less or more arguments are received than required
+    if(input.length != 2) {
```
kaustubh-karkare: Space after `if`.

solutions/divyansh-jamuaar/Assignment-2.1/Server.java
```diff
@@ -0,0 +1,64 @@
...
+      out.println("Error: Number of arguments should be 2");
+      osw.flush();
+    }
+    else {
+      try {
+        int number1 = Integer.parseInt(input[0]);
+        int number2 = Integer.parseInt(input[1]);
+        int ans = number1 + number2;
```
kaustubh-karkare: This is application logic leaking into framework code again.
A method called `getOutput` shouldn't have anything to do with addition like this.

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
divyanshjr: Will change..

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
divyanshjr: I'll change it to 127.0.0.1

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
divyanshjr: Yes, that would be better. I'll do it.

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
divyanshjr: I didn't know about Logger class earlier. I'll implement it.

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
divyanshjr: Sure.

solutions/divyansh-jamuaar/Assignment-2.1/Server.java
divyanshjr: Okay

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
divyanshjr: Okay, I'll make sure that application logic and framework are dealt separately. 

Do you mean to say that I should have a common main method for both Client and Server classses?

solutions/divyansh-jamuaar/Assignment-2.1/Server.java
divyanshjr: Alright

solutions/divyansh-jamuaar/Assignment-2.1/Server.java
divyanshjr: Yes, I'll change their names.

solutions/divyansh-jamuaar/Assignment-2.1/Server.java
divyanshjr: Alright

solutions/divyansh-jamuaar/Assignment-2.1/Client.java
kaustubh-karkare: Yeah, so when I coded this to get a sense of how much time it would take, I had the server and client running on separate threads, communicating via a random socket (instead of a fixed one like 9999).

divyanshjr:
I've updated the solution to Assignment 2.1. Brief working of my code - 
-> The program now supports communication between a single server and multiple clients which run as 
    separate threads.
-> The Server has an additional Worker thread (ClientServiceThread) which processes the requests of the 
    clients. The Server accepts requests from the clients and passes them to the worker thread which 
    processes them and respnds back to the respective client.
-> All the messages generated during the communication are stored in a Log File. This feature is 
     implemented with the help of Logger Class in Java.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+        logger.log(Level.WARNING, "Client "+clientID+": Incorrect result obtained");
+      }
+    }
+  }
+}
+
+// Class containing the main method
+class Assignment2_1 {
+  private final static Logger logger = Logger.getLogger(Assignment2_1.class.getName());
+  private static FileHandler fh = null;
```
kaustubh-karkare: No abbreviations please.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
+/*JAVA program to implement Multiple client-server interaction
+ *Multithreading approach is used where Server and Clients run as separate threads
+ *Server has an additional ClientServiceThread which processes the requests of the clients
+ *The Server accepts requests from Clients and passes the requests to ClientServiceThread
+ *All messages are recorded in a LOG File
```
kaustubh-karkare: According to the style guide, you should have spaces after `*`.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+class Server extends Thread {
+  private static final Logger logger = Logger.getLogger(Server.class.getName());
+  static int port;
+  static ServerSocket serverSocket;
+  static Socket socket;
+  // Constructor
+  public Server(int port) throws Exception {
+      Server.port = port;
+      serverSocket = new ServerSocket(port);
+    }
```
kaustubh-karkare: The indenting for the last 3 lines is off.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+  @Override
+  public void run() {
+    logger.log(Level.INFO, "Server Started");
```
kaustubh-karkare: I dont know the exact API, but based on how other languages do it, I would expect that there will be methods like `logger.info` here. Could you confirm?

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+    while (true) {
+      logger.log(Level.INFO, "Waiting for client to connect");
+      try {
+        socket = serverSocket.accept();
+      } catch(Exception e) {}
+      logger.log(Level.INFO, "Client connected");
```
kaustubh-karkare: Specify information about the client?

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+      ClientServiceThread worker = new ClientServiceThread(socket);
+      worker.start();
+    }
+  }
+}
+
+// Class which is responsible for processing Client request
+class ClientServiceThread extends Thread {
+  Socket socket;
+  //Constructor
```
kaustubh-karkare: nit: Space after `//`

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+  public ClientServiceThread(Socket socket) {
+    this.socket = socket;
+  }
+  @Override
+  public void run() {
+    try {
+      Thread.sleep(2000); // Wait for 2 seconds
+    } catch(Exception e) {}
```
kaustubh-karkare: In general, catch-all exceptions are a bad idea, since they hide information that could otherwise be useful.
In fact, I'd rather not have exceptions caught at all, unless it is information that you actively use.
What's a valid usecase of catching? Database connection failure, since that is something you can retry.
In most cases, just let it bubble up to `main`, which can then figure out what to do.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+    try {
+      String[] input; // Input received from Client
+      String reply; // Reply to be sent to Client
+      input = fetchClientRequest();
```
kaustubh-karkare: nit: Use `get` instead of `fetch`? Fetch usually suggests that you have to make a network call to obtain the data, which get is for already available data. This is borderline, but the data is already delivered, so `get` makes more sense to me.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+      reply = processData(input);
+      replyClient(reply);
```
kaustubh-karkare: `replyToClient` might be better.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+  @Override
+  public void run() {
+    try {
+      Thread.sleep(2000); // Wait for 2 seconds
+    } catch(Exception e) {}
+    try {
+      String[] input; // Input received from Client
+      String reply; // Reply to be sent to Client
+      input = fetchClientRequest();
+      reply = processData(input);
```
kaustubh-karkare: It's cool that you have isolated the adding logic to this one method. Let's take it one step further, and take it is an input to your `Server` class.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+      replyClient(reply);
+    } catch(Exception e) {}
+  }
+  // Method to fetch data from client
+  String[] fetchClientRequest() throws Exception {
```
kaustubh-karkare: I'm not sure it makes sense to even have this method (and replyClient).
This design works when the client makes one request, and gets one response, but in order to make things truly generic, you cant make that assumption in the general case.
I'd think you could store your ObjectInputStream and OutputStreamWriter as class properties, and use them directly from `run`.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+    reply = "Error: Number of arguments should be 2";
+    else {
+      try {
+        int number1 = Integer.parseInt(input[0]);
+        int number2 = Integer.parseInt(input[1]);
+        int ans = number1 + number2;
+        reply = Integer.toString(ans);
+      }
+      // In case input arguments are not of integer type
+      catch(Exception e) {
```
kaustubh-karkare: Your comment helps, but you should use the proper exception corresponding to that case.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+class Client extends Thread {
+  private static final Logger logger = Logger.getLogger(Client.class.getName());
+  int selfComputedResult; // to verify received result
+  String[] arguments; // input from user
+  Socket socket;
+  int clientID;
+  // Constructor
+  public Client(String IP, int port, String[] input, int clientID) throws Exception {
+    socket = new Socket(IP, port);
+    arguments = input;
```
kaustubh-karkare: This is confusing. Would prefer to have a single name for a given piece of data consistently.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+  public void run() {
+    String response = null;
+    try {
+      queryServer();
+      response = getResponse();
+    } catch(Exception e) {}
+    validate(response);
+  }
+  // Method to send client data to server
+  public void queryServer() throws Exception {
```
kaustubh-karkare: My comments about `fetchClientRequest` and `replyClient` apply here too.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+  }
+  // Method to receive response from server
+  public String getResponse() throws Exception {
+    BufferedReader br =
+    new BufferedReader(new InputStreamReader(socket.getInputStream()));
+    String response = br.readLine();
+    return response;
+  }
+  // Method to validate the answer recieved from server
+  public void validate(String response) {
```
kaustubh-karkare: This method is a bit hard to read. Can you simplify?

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+      logger.log(Level.WARNING, "Client "+clientID+": Error message from server side - "+response);
+    }
+    else {
+      logger.log(Level.INFO, "Client "+clientID+": Answer received from server: "+response);
+      int result = Integer.parseInt(response);
+      if (result == selfComputedResult) {
+        logger.log(Level.FINE, "Client "+clientID+": Result verified");
+      }
+      else {
+        logger.log(Level.WARNING, "Client "+clientID+": Incorrect result obtained");
```
kaustubh-karkare: According to style guide, spaces around operators.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+  static String[] input;
+  // Method to create log file
+  public static void initLog() {
+    try {
+      fh = new FileHandler("Status.log", false);
+    }
+    catch (SecurityException | IOException e) {
+      e.printStackTrace();
+    }
+    Logger l = Logger.getLogger("");
```
kaustubh-karkare: Same as above.

solutions/Assignment2_1.java
```diff
@@ -0,0 +1,206 @@
...
+  }
+  // Main method
+  public static void main(String args[]) {
+    initLog();
+    Scanner scanner = new Scanner(System.in);
+    System.out.println("Enter the number of clients");
+    clientsCount = scanner.nextInt();
+    client = new Client[clientsCount];
+    // Trying to set up the server
+    System.out.println("Enter port on which server has to be hosted");
```
kaustubh-karkare: Use command line arguments for this. It is okay to use an existing options parser library this time :)

solutions/Assignment2_1.java
divyanshjr: I'll correct that.

solutions/Assignment2_1.java
divyanshjr: logger.info() takes just one parameter i.e. the message and sets it to the level INFO. logger.log() is a generalised method which takes 2 arguments (Level , Message). I could even use methods like logger.info(), logger.warning() etc if you wish.

solutions/Assignment2_1.java
divyanshjr: Alright.

solutions/Assignment2_1.java
divyanshjr: Agree, I'll try to specify the Exception type in the catch statement. But here I won't be able to propagate my Exception to the main method since the run() method is not allowed to throw any exception.

solutions/Assignment2_1.java
divyanshjr: Sure. I'll change it.

solutions/Assignment2_1.java
divyanshjr: Alright.

solutions/Assignment2_1.java
divyanshjr: I didn't get you. Could you please explain a bit more?

solutions/Assignment2_1.java
divyanshjr: Okay, so that means I should only have a processData() method other than run() and rest of the job should be done inside the run() method.

solutions/Assignment2_1.java
divyanshjr: Yes, I'll catch the specific Exception only.

solutions/Assignment2_1.java
divyanshjr: Sure.

solutions/Assignment2_1.java
divyanshjr: Agree, So only one method should be there to validate the response. Rest of the job is to be done inside run().

solutions/Assignment2_1.java
divyanshjr: I'll try to simplify.

solutions/Assignment2_1.java
divyanshjr: oops.. need to change my habit.. :P

solutions/Assignment2_1.java
divyanshjr: Alright. I'll do it.

solutions/Assignment2_1.java
kaustubh-karkare: Prefer `logger.info`. This looks unnecessarily verbose.

solutions/Assignment2_1.java
kaustubh-karkare: Right now, the logic for addition is hardcoded in `processData`. But we want to make it even more flexible, and allow the users of this class to specify how to interact with the client. Basically, make it work beyond this one exact example (you'll hear be say this a lot).

solutions/Assignment2_1.java
kaustubh-karkare: Hmm. It seems weird to have 2 consecutive statements like this. I get why you're doing it, but still, looks weird :P

solutions/Assignment2_1.java
divyanshjr: Alright.. :)

solutions/Assignment2_1.java
divyanshjr: yeah.. it is.. but I'll try..

PR #14, opened 2017-03-21
Solution to Problem 2.1

chaudharyS:
Kindly review.


solutions/srishti-chaudhary/assignment 2.1/client.py
```diff
@@ -0,0 +1,33 @@
...
+
+import socket
+import sys
+
+# Creating TCP/IP socket
+sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+# Connecting the socket to the port where the server is listening 
+server_address = ('localhost', 10000)
+print >>sys.stderr, 'connecting to %s port %s' % server_address
```
kaustubh-karkare: Interesting. I've not see this usage of streams before.
However, I don't think that printing is the right model to use here.
Instead, read up on the `logging` module, and use a logger instead.
That allows you to separate logging information from different sources if necessary, even if it's not applicable here.

solutions/srishti-chaudhary/assignment 2.1/client.py
```diff
@@ -0,0 +1,33 @@
+#! /usr/bin/env python
```
kaustubh-karkare: Not sure if there should be a space after `!`

solutions/srishti-chaudhary/assignment 2.1/client.py
```diff
@@ -0,0 +1,33 @@
...
+sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+# Connecting the socket to the port where the server is listening 
+server_address = ('localhost', 10000)
+print >>sys.stderr, 'connecting to %s port %s' % server_address
+sock.connect(server_address)
+
+def calculate_sum():
+	first_number = int(sys.argv[1])
+	second_number = int(sys.srgv[2])
```
kaustubh-karkare: typo `s/srgv/argv/`
I was kinda hoping you'd use a proper options parser at this point instead of accessing the raw list, but I dont care enough. In fact, random numbers would be fine too.

solutions/srishti-chaudhary/assignment 2.1/client.py
```diff
@@ -0,0 +1,33 @@
...
+	result = first_number + second_number
+	return result
+
+try:
+	# Sending data
+	sock.sendall(sys.argv[1] + ' ' + sys.argv[2])
```
kaustubh-karkare: Accessing input from the arguments is repeated twice (here and on lines 15-16). Which means, that if I modified the problem to something like './client --multiply 10 20', you'd have to update multiple places, which would need to be kept in sync. And never underestimate the power of human error, and so I'd rather avoid inconsistencies from even being possible.

solutions/srishti-chaudhary/assignment 2.1/server.py
```diff
@@ -0,0 +1,34 @@
...
+		print >>sys.stderr, 'connection from', client_address
+		
+		# Receiving data from client
+		while True:
+			data = connection.recv(1024)
+			if not data: break
+			print >>sys.stderr, 'received %s' %data
+			time.sleep(2)
+			data = data.split()
+			server_sum = data[0] + data[1]
```
kaustubh-karkare: Huh. I thought you'd need to cast these to integers.

solutions/srishti-chaudhary/assignment 2.1/client.py
```diff
@@ -0,0 +1,40 @@
...
+
+port = 2004
+# Creating TCP/IP socket
+sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+#For maintaining a log of events
+LOG_FILENAME = 'client.log'
+logging.basicConfig(filename=LOG_FILENAME,
+                    level=logging.DEBUG,
+                    )
```
kaustubh-karkare: So I know this is valid according to the style guide, but I prefer:
```
logging.basicConfig(
  filename=LOG_FILENAME,
  level=logging.DEBUG,
)
```
because it is closer to how other languages do it, and I prefer consistency.

solutions/srishti-chaudhary/assignment 2.1/client.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python
+
+import socket
+import sys
+import logging
+
+port = 2004
```
kaustubh-karkare: This should be an input.

solutions/srishti-chaudhary/assignment 2.1/client.py
```diff
@@ -0,0 +1,40 @@
...
+#For maintaining a log of events
+LOG_FILENAME = 'client.log'
+logging.basicConfig(filename=LOG_FILENAME,
+                    level=logging.DEBUG,
+                    )
+
+# Connecting the socket to the port where the server is listening 
+server_address = ('localhost', 10000)
+logging.info('connecting to %s port %s' % server_address)
+sock.connect(server_address,port)
```
kaustubh-karkare: Wait, I'm confused. `server_address` already contains a port (10000), so what do you need the other port (2004) for?

solutions/srishti-chaudhary/assignment 2.1/client.py
```diff
@@ -0,0 +1,40 @@
...
+	server_sum = int(sock.recv(1024))
+	if server_sum == client_sum:
+		logging.info('Result Veified')
+	else:
+		logging.warning('Incorrect Result')
+finally:
+	logging.info('Closing socket')
+	sock.close()
+
+
```
kaustubh-karkare: Unnecessary extra lines at the end of the file. You want only 1.

solutions/srishti-chaudhary/assignment 2.1/server.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python
+
+import socket
+import sys
+import time
+import logging
+import threading 
+
+#For maintaining a log of events
```
kaustubh-karkare: You added spaces after `#` in the previous file, but not here. Please be consistent.

solutions/srishti-chaudhary/assignment 2.1/server.py
```diff
@@ -0,0 +1,58 @@
...
+LOG_FILENAME = 'server.log'
+logging.basicConfig(filename=LOG_FILENAME,
+                    level=logging.DEBUG,
+                    )
```
kaustubh-karkare: Same as previous comment.

solutions/srishti-chaudhary/assignment 2.1/server.py
```diff
@@ -0,0 +1,58 @@
...
+			server_sum = int(data[0]) + int(data[1])
+			connection.send(str(server_sum))
+
+
+# Creating TCP/IP socket
+sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+# Binding the socket to the port
+server_address = ('localhost', 10000)
+logging.info('starting server up on %s port %s' % server_address)
+sock.bind(server_address)
```
kaustubh-karkare: Instead of doing all this as global statements, make a class, so that I can do something like:
```
Server(on_client=do_this_thing).start_accepting_connections()
# Do something else while the server is working on a different thread
```

solutions/srishti-chaudhary/assignment 2.1/client.py
```diff
@@ -4,20 +4,20 @@
...
+logging.basicConfig(
+  filename=LOG_FILENAME,
+  level=logging.DEBUG,
+)
 
 # Connecting the socket to the port where the server is listening 
 server_address = ('localhost', 10000)
 logging.info('connecting to %s port %s' % server_address)
-sock.connect(server_address,port)
+sock.connect(server_address, 10000)
```
kaustubh-karkare: Why do you need to repeat the port number? Didn't you specify it on line 18?

solutions/srishti-chaudhary/assignment 2.1/server.py
kaustubh-karkare: Still waiting for this.

solutions/srishti-chaudhary/assignment 2.1/client.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+
+import socket
+import sys
+import logging
+
```
kaustubh-karkare: You wnat to put all of the following in a function that you call from `main`.
In general, global variables are a bad idea, since they can be used / modified from anywhere, and so make things hard to reason about.

kaustubh-karkare:
Inactive for too long.

PR #17, opened 2017-03-22
assignment-2.1.py

gupta-rohan:
Kindly review.

### DIFF
--- a/solutions/rohan-gupta/assignment-2.1.py/client.py
+++ b/solutions/rohan-gupta/assignment-2.1.py/client.py
@@ -1,30 +1,27 @@
 import socket
+import sys
 
 
 # Create a socket object
 client_socket_obj = socket.socket()
 # Get local machine name
-client_host = socket.gethostname()  
+client_host = sys.argv[1]
 # Reserve a port for your service
-client_port = 1011
-print 'Enter first number.'  
-num1 = int(raw_input())
-print 'Enter second number.'  
-num2 = int(raw_input())
-add = str(num1 + num2);  
-print 'The sum calculated by the client is',add
+client_port = int(sys.argv[2])
+num1 = int(sys.argv[3])
+num2 = int(sys.argv[4])
+add = str(num1 + num2)
+print 'The sum calculated by the client is', add
 # The client object is connecting to the given address
 client_socket_obj.connect((client_host, client_port))
 print 'Connection to the server is made.'
 # The expression to be sent to the server is evaluated
 expression = str(num1) + ' ' + str(num2)
-client_socket_obj.send(expression)   
-evaluated_expression = client_socket_obj.recv(1024)
+client_socket_obj.send(expression)
+valuated_expression = client_socket_obj.recv(1024)
 if evaluated_expression == add:
     print 'Sum calculated by server and client is the same.'
     print 'VERIFIED.'
 else:
     print 'Sum calculated by server and client is not the same.'
     print 'NOT VERIFIED.'
-client_socket_obj.close()
-
--- a/solutions/rohan-gupta/assignment-2.1.py/client.py
+++ b/solutions/rohan-gupta/assignment-2.1.py/client.py
@@ -1,0 +1,30 @@
+import socket
+
+
+# Create a socket object
+client_socket_obj = socket.socket()
+# Get local machine name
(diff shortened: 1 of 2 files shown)


solutions/rohan-gupta/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,56 @@
...
+        while (j > index and j < len(self.express)):
+            num2 += self.express[j]
+            j = j + 1
+        add = int(num1) + int(num2)
+        return add
+
+    def multi_thread_implementation(client_socket_obj, client_addr):
+        received_string = client_socket_obj.recv(1024)
+        Addition_obj = Addition(received_string)
+        print 'The client address is', client_addr
```
kaustubh-karkare: Use the logging module, instead of print statements. That way, the caller can choose to ignore / write to file / etc.

solutions/rohan-gupta/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,27 @@
...
+
+# Create a socket object
+client_socket_obj = socket.socket()
+# Get local machine name
+client_host = sys.argv[1]
+# Reserve a port for your service
+client_port = int(sys.argv[2])
+num1 = int(sys.argv[3])
+num2 = int(sys.argv[4])
+add = str(num1 + num2)
```
kaustubh-karkare: "add" is a bad name. It sounds like a function whose job it is to add numbers. Not the result of that function.

solutions/rohan-gupta/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,27 @@
+import socket
+import sys
+
+
+# Create a socket object
+client_socket_obj = socket.socket()
```
kaustubh-karkare: Suffixes like "_obj" generally add no value, and shouldn't be used.
If they do add value, it generally means that there is a different naming problem that causes them to be useful.

solutions/rohan-gupta/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,27 @@
...
+num2 = int(sys.argv[4])
+add = str(num1 + num2)
+print 'The sum calculated by the client is', add
+# The client object is connecting to the given address
+client_socket_obj.connect((client_host, client_port))
+print 'Connection to the server is made.'
+# The expression to be sent to the server is evaluated
+expression = str(num1) + ' ' + str(num2)
+client_socket_obj.send(expression)
+valuated_expression = client_socket_obj.recv(1024)
```
kaustubh-karkare: I dont know what `valuated_expression` means. And that means it isn't useful as a name.

solutions/rohan-gupta/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,56 @@
...
+server_port = int(sys.argv[2])
+# Binding server address to the socket
+server_socket_obj.bind((server_host, server_port))
+print 'Server code being executed.'
+server_socket_obj.listen(5)
+while True:
+    client_socket_obj, client_addr = server_socket_obj.accept()
+    # Implementing multi-threading
+    try:
+        thread.start_new_thread(multi_thread_implementation,
```
kaustubh-karkare: Isn't this an instance method? How are you accessing it directly?

solutions/rohan-gupta/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,56 @@
...
+    def multi_thread_implementation(client_socket_obj, client_addr):
+        received_string = client_socket_obj.recv(1024)
+        Addition_obj = Addition(received_string)
+        print 'The client address is', client_addr
+        time.sleep(2)
+        server_evaluated_sum = Addition_obj.answer()
+        print 'The sum calculated by the server is', server_evaluated_sum
+        client_socket_obj.send(str(server_evaluated_sum))
+
+# Create a socket object
```
kaustubh-karkare: I actually meant that the section below should be in a class. Something like:
```
Server(host='localhost', post=12345, on_client_connection=<function>).start_serving()`

kaustubh-karkare:
Inactive for too long.

PR #20, opened 2017-03-22
assignment-2.1.py

humanoid2795:
Kindly review.


solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,51 @@
+# The client side code requires the server host name,
+# port number and the arguments to calucalte the sum
+
+import sys
+import socket
+import json
+
+
+class client:
```
kaustubh-karkare: Naming conventions require that class names be UpperCamelCase

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(self, server_host, server_port):
+        self.server_host = server_host
+        self.server_port = server_port
+        self.client_socket_object = socket.socket(socket.AF_INET,
```
kaustubh-karkare: In general, suffixes like "_object" dont have any value. If they do, then it usually means that you've made a different naming mistake, and fixing that should make them unnecessary.
Additionally, I prefer this style, for consistency with other languages:
```
self.client_socket = socket.socket(
  socket.AF_INET,
  socket.SOCK_STREAM,
)
```

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,51 @@
...
+def main():
+    if(len(sys.argv) < 5):
+        print('Insufficient argument')
+        sys.exit(2)
+    server_host = sys.argv[1]
+    server_port = int(sys.argv[2])
+    client_object = client(server_host, server_port)
+    client_object.data_to_send_to_server(int(sys.argv[3]), int(sys.argv[4]))
+    client_object.data_exchange_from_server()
+    client_object.check_server_and_client_result()
```
kaustubh-karkare: Your client class, as the name suggests, is responsible for setting up the socket client. Nowhere in the name "client" does it deal with adding numbers, and so there's no way anyone would know (without looking at your implementation), that these methods even exist. Which means that it isn't intuitive, and therefore not good enough.
How about `client(on_connection=function_that_sends_numbers_and_verifies_sum)`

solutions/anand-kumar/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,57 @@
...
+import time
+import thread
+
+
+class server:
+    first_number = 0
+    second_number = 0
+    data_from_client = ''
+
+    def __init__(self, client_socket_object):
```
kaustubh-karkare: Like before, `_object` doesn't add value.

solutions/anand-kumar/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,57 @@
...
+        self.sum_of_numbers = {'sum': self.sum_value}
+        self.encode_sum_of_numbers = json.dumps(self.sum_of_numbers)
+        time.sleep(2)
+        return self.encode_sum_of_numbers
+
+
+def thread_of_a_client(client_socket_object, client_address):
+    print('Connection received from ', client_address)
+    server_object = server(client_socket_object)
+    server_object.extract_numbers_from_client_data()
```
kaustubh-karkare: Same as before: no one could expect these methods to exist as they're not intuitive on a class called `server`. Also, your server class has nothing to do with a server anyway! It's just dealing with clients.

solutions/anand-kumar/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,57 @@
...
+
+def main():
+    if (len(sys.argv) < 3):
+        print('Insuffcient argument')
+        sys.exit(2)
+    server_socket_object = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    server_host_name = sys.argv[1]
+    server_port_number = int(sys.argv[2])
+    server_socket_object.bind((server_host_name, server_port_number))
+    server_socket_object.listen(5)
```
kaustubh-karkare: The above 5 lines should have been in a class too. Something like
`Server(host=..., port=..., on_client_connection=...).start()`
is much easier to use than what you have here.

solutions/anand-kumar/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,57 @@
...
+    while True:
+        client_socket_object, client_address = server_socket_object.accept()
+        thread.start_new_thread(thread_of_a_client,
+                                (client_socket_object, client_address))
```
kaustubh-karkare: See my previous formatting comment.

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,59 @@
...
+
+import sys
+import socket
+import json
+import logging
+
+
+class Client:
+
+    def __init__(self, server_host, server_port, first_number, second_number):
```
kaustubh-karkare: Each classes is supposed to represent a single concept, but it looks like you're mixing 2 concepts.
First of all, you have the socket client, which is appropriate to the name of the class.
But you also have all this logic to add numbers, which is not mentioned in the name, and should probably be separate. Tbh, I feel that you dont really need an entire class for setting up the socket, since it only takes 3 lines ... a function would be sufficient. That way, you can assume that the socket is the input to your addition class, and you've decoupled the functionality so that you can easily replace it with a multiplication class.

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,59 @@
...
+                self.numbers_to_be_added['second_number'] ==
+                self.sum_received['sum']):
+            print('The server gives the correct sum: ',
+                  self.sum_received['sum'])
+        else:
+            print('The server does not give the correct sum')
+
+
+def main():
+    if(len(sys.argv) < 5):
```
kaustubh-karkare: Space after `if`. But you dont need parenthesis anyway.
It's fine to use the argparse module now. You have a sense of how it works, and it'll help make provide a cleaner API.

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,59 @@
...
+            'first_number': first_number,
+            'second_number': second_number
+        }
+        self.create_client_socket()
+        self.send_numbers_to_add()
+        self.receive_sum_from_the_server()
+        self.verify_received_sum()
+
+    def create_client_socket(self):
+        self.client_socket_object = socket.socket(
```
kaustubh-karkare: Suffixes like "_object" are generally not useful IMO. The rest of the name makes it clear what it is.
In cases where a suffix like that does prove valuable, it is my experience that it means that you're making a naming mistake elsewhere. Like, if you called something "test" (bad name, since it doesn't make it clear that it is an ID) instead of "test_id", then "test_object" makes sense. Otherwise, avoid that suffix and other similar ones.

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,59 @@
...
+        self.server_host = server_host
+        self.server_port = server_port
+        self.numbers_to_be_added = {
+            'first_number': first_number,
+            'second_number': second_number
+        }
+        self.create_client_socket()
+        self.send_numbers_to_add()
+        self.receive_sum_from_the_server()
+        self.verify_received_sum()
```
kaustubh-karkare: This involves too many functions. You need to find a balance, since each extra function adds a layer of indirection, causing me to have to context switch. In this case, I would have just used a class, so that the reviewer could read code continuously (instead of having to jump around).

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,59 @@
...
+import json
+import logging
+
+
+class Client:
+
+    def __init__(self, server_host, server_port, first_number, second_number):
+        self.server_host = server_host
+        self.server_port = server_port
+        self.numbers_to_be_added = {
```
kaustubh-karkare: Dictionaries like this are generally not a good idea. Prefer enums or classes, when the number of elements is fixed.

solutions/anand-kumar/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,67 @@
...
+import time
+import threading
+
+
+class Server:
+
+        def __init__(self, server_host, server_port):
+            self.address = (server_host, server_port)
+
+        def start_surveying(self):
```
kaustubh-karkare: lol. Did you mean `start_serving()`?

solutions/anand-kumar/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,67 @@
...
+            self.server_socket = socket.socket()
+            self.server_socket.bind(self.address)
+            self.server_socket.listen(1)
+            while True:
```
kaustubh-karkare: How do I stop the server once I'm ready? There should be a stop method.

solutions/anand-kumar/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,67 @@
...
+                client_socket, client_address = self.server_socket.accept()
+                ClientThread(client_socket, client_address).start()
+            self.server_socket.close()
+
+
+class ClientThread(threading.Thread):
```
kaustubh-karkare: See previous feedback about complexity that might not be necessary.

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __init__(self, client_socket, first_number, second_number):
+        self.first_number = first_number
+        self.second_number = second_number
+        self.client_socket = client_socket
+
+    def send_numbers_and_verify_sum(self):
+        self.client_socket.send(json.JSONEncoder().encode({
+            'first_number': self.first_number,
+            'second_number': self.second_number}))
```
kaustubh-karkare: This is fine for now, but since JSON is a flexible data format, there is always a risk that the data you need might be incomplete ('first_number' without 'second_number'). As a result, people have come up with frameworks to send structured data over the wire (like structs, enums, exceptions, in a consistent manner). Examples being Google's Protocol Buffers / Apache Thrift, etc. You should look them up.

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,65 @@
...
+        self.sum_received = json.JSONDecoder().decode(
+            self.client_socket.recv(255))
```
kaustubh-karkare: I prefer this formatting, because it is consistent with other languages.:
```
self.sum_received = json.JSONDecoder() \
  .decode(self.client_socket.recv(255))
```

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,65 @@
...
+        if (
```
kaustubh-karkare:
```
if (
   ...
):
   ...
```

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,65 @@
...
+                self.second_number ==
+                self.sum_received['sum']):
+            print(
+                'The server gives the correct sum: ',
+                self.sum_received['sum'])
+        else:
+            print('The server does not give the correct sum')
+
+
+def client_socket_description(host, port):
```
kaustubh-karkare: Why description? It sounds like you're returning a string.

solutions/anand-kumar/assignment-2.1.py/client.py
```diff
@@ -0,0 +1,65 @@
...
+        type=int,
+        required=True,
+        help='First number for summation')
+    parser.add_argument(
+        '--second_number',
+        '-num2',
+        type=int,
+        required=True,
+        help='Second number for summation')
+    req_argument = parser.parse_args()
```
kaustubh-karkare: "req" :P

solutions/anand-kumar/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,62 @@
...
+            socket.IPPROTO_TCP)
+        self.server_socket.bind(self.address)
+        self.server_socket.listen(1)
+        server_is_listening = True
+        while server_is_listening:
+            try:
+                client_socket, client_address = self.server_socket.accept()
+                ReceiveNumbersAndReturnSum(
+                    client_socket, client_address).start()
+            except KeyboardInterrupt:
```
kaustubh-karkare: This is cool, but you're then stuck in this loop until Ctrl+C.
I imagine that the server runs in its own thread, and then your main function can do other things (like sleep for 5 seconds, then `server.close()`).

solutions/anand-kumar/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,62 @@
...
+
+class ReceiveNumbersAndReturnSum(threading.Thread):
+
+    def __init__(self, client_socket, client_address):
+        self.client_socket = client_socket
+        self.client_address = client_address
+        threading.Thread.__init__(self)
+
+    def run(self):
+        print('Client connection received from:'), self.client_address
```
kaustubh-karkare: Use the logging module, instead of print statements. That'll make sure that your output from different threads doesn't get mixed up.

solutions/anand-kumar/assignment-2.1.py/server.py
```diff
@@ -0,0 +1,86 @@
...
+                ReceiveNumbersAndReturnSum(
+                    client_socket,
+                    client_address
+                ).start()
+            except KeyboardInterrupt:
+                # ctrl+C was hit - server stopped listening
+                self.server_is_listening = False
+            except socket.timeout:
+                if self.server_is_listening is False:
+                    break
```
kaustubh-karkare: You don't need this. A simple "pass" would have been sufficient, since the while loop has the condition anyway.

PR #22, opened 2017-03-23
Solution to Assignment 2.2 added

arch-DJ:
- I've created a Download Manager in Java (Command Line Interface) which uses multithreading to download a file from a particular URL.

- The file to be downloaded will be divided into 'n' number of parts which will be taken input as CommandLine Argument. 'n' separate threads will download those 'n' parts separately.

- The maximum value of 'n' can be 20. Giving input for 'n' is like configuring the downloader manually to control to number of simultaneous requests sent to the website.

- The downloaded file will be saved in the directory in which code is present.


solutions/divyansh-jamuaar/Assignment-2.2/Assignment2_2.java
```diff
@@ -0,0 +1,154 @@
+/** Java code to implement a Multithreaded Download Manager
+  * It divides the incoming file into n parts and each part is downloaded by an individual thread
+  * The value of n is scanned as CommandLine Argument (MAX 20)
+  */
```
kaustubh-karkare: I thought multiline comments were generally structured as:
```
/**
 * ...
 * ...
 */

File: solutions/divyansh-jamuaar/Assignment-2.2/Assignment2_2.java [hunk 2]
```diff
@@ -0,0 +1,154 @@
...
+import java.io.*;
+import java.net.*;
+
+// Class containing the main method
+class Assignment2_2 {
+  // main method
+  public static void main(String args[]) {
+    Scanner scanner = new Scanner(System.in);
+    if (args.length == 0) {
+      System.out.println("Please enter the number of parts of the file as CommandLine Argument (MAX 20)");
```
kaustubh-karkare: No need of this. Use a default if they don't provide it.

solutions/divyansh-jamuaar/Assignment-2.2/Assignment2_2.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+    else {
+      try {
+        int threadCount = Integer.parseInt(args[0]);
+        if (threadCount <= 0 || threadCount > 20) {
+          throw new NumberFormatException();
```
kaustubh-karkare: I don't see anything wrong with the format of "21". Looks like bad naming.

solutions/divyansh-jamuaar/Assignment-2.2/Assignment2_2.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+        System.out.println("Enter the URL");
+        String url = scanner.nextLine();
```
kaustubh-karkare: Take this as command line input (required.

solutions/divyansh-jamuaar/Assignment-2.2/Assignment2_2.java
```diff
@@ -0,0 +1,154 @@
...
+        DownloadManager manager = new DownloadManager(url, threadCount);
+      }
+        catch(NumberFormatException e) {
+        System.out.println("Number of parts should be a positive Integer and less than 21");
+        }
```
kaustubh-karkare: The indenting is messed up here. Line 27 and 29 should have 2 less spaces.

solutions/divyansh-jamuaar/Assignment-2.2/Assignment2_2.java
```diff
@@ -0,0 +1,154 @@
...
+    this.startByte = startByte;
+    this.endByte = endByte;
+    this.threadNum = threadNum;
+    this.outputFile = outputFile;
+  }
+
+  @Override
+  public void run() {
+    try {
+      HttpURLConnection httpURLConnection = (HttpURLConnection) downloadURL.openConnection();
```
kaustubh-karkare: You're not allowed to use modules like this that do all the hard lifting for you.
You're only allowed to use `Socket`.

solutions/divyansh-jamuaar/Assignment-2.2/Assignment2_2.java
```diff
@@ -0,0 +1,154 @@
...
+        DownloadManager manager = new DownloadManager(url, threadCount);
+      }
+        catch(NumberFormatException e) {
+        System.out.println("Number of parts should be a positive Integer and less than 21");
+        }
+    }
+  }
+}
+
+// Class which manages the Download process
```
kaustubh-karkare: This comment has no value, because I still don't know what "manage" involves.
Something like "This class creates a bunch of Downloader objects ..." is more specific and so more useful.

solutions/divyansh-jamuaar/Assignment-2.2/Assignment2_2.java
```diff
@@ -0,0 +1,154 @@
...
+  private String downloadUrl;
+  private String fileName;
+  private String fileExtension;
+  private URL nonStringUrl;
+  private HttpURLConnection connection;
+  private int fileSize;
+  private int remainingByte;
+  private RandomAccessFile outputFile;
+  private boolean isDownloadPossible = true;
+  private int threadCount;
```
kaustubh-karkare: Uh .... this is ugly. Why do you need to save so much information?

In general, save only the bare minimum information that you cant compute.

solutions/divyansh-jamuaar/Assignment-2.2/Assignment2_2.java
```diff
@@ -0,0 +1,154 @@
...
+    initiateDownload();
+  }
+
+  private void initiateDownload() {
+    fileName = downloadUrl.substring(downloadUrl.lastIndexOf('/') + 1, downloadUrl.length());
+    fileExtension = downloadUrl.substring(downloadUrl.lastIndexOf('.') + 1, downloadUrl.length());
+
+    try {
+      nonStringUrl = new URL(downloadUrl);
+      connection = (HttpURLConnection) nonStringUrl.openConnection();
```
kaustubh-karkare: Cant use "HttpURLConnection".

solutions/divyansh-jamuaar/Assignment-2.2/Assignment2_2.java
```diff
@@ -0,0 +1,154 @@
...
+    int last = fileSize - 1;
+    if(isDownloadPossible) {
+      System.out.println("Downloading...");
+      Downloader downloadParts[] = new Downloader[threadCount];
+      // Creating and starting threads to download different parts of the file
+      for (int i = 0; i < threadCount; i++) {
+        if (i < threadCount) {
+          downloadParts[i] = new Downloader(nonStringUrl, first, last, (i+1), outputFile);
+          downloadParts[i].start();
+          try{
```
kaustubh-karkare: Bunch of style issues around this area. Example? There should have been a space before `{`

divyanshjr:
**Second Iteration**

- I've used only Sockets for downloading the file from a particular URL unlike the previous commit.

- "Assignment2_2.java" is the driver code which takes URL and number of threads as command line arguments. It starts the FileDownloader thread which initiates the download process.

- "FileDownloader.java" is a part of "downloader" package which verifies the URL and starts multiple threads to download different parts of the file simultaneously.

- "DownloadThread.java" is the worker thread which downloads a particular portion of the file.

solutions/Assignment2_2.java
```diff
@@ -0,0 +1,56 @@
...
+	private final static Logger logger = Logger.getLogger(Assignment2_2.class.getName());
+	private static String url;
+	private static int threadCount;
+	private static Namespace namespace;
+	private static FileDownloader fileDownloader;
+	
+	public static void main(String[] args) {
+		ArgumentParser parser = ArgumentParsers
+				.newArgumentParser("Downloader")
+				.defaultHelp(true);
```
kaustubh-karkare: I prefer consistent indenting. Note the difference between these section, and the parser.addArgument section, which is inconsistent. Stick to 1 tab more than parent line.

solutions/Assignment2_2.java
```diff
@@ -0,0 +1,56 @@
...
+		try {
+			namespace = parser.parseArgs(args);
+		} catch (ArgumentParserException e) {
+			logger.warning(e.getMessage());
+			return;
+		}
+		
+		url = namespace.getString("url");
+		threadCount = namespace.getInt("threads");		
+		fileDownloader = new FileDownloader(url, threadCount); // Initiate downloading
```
kaustubh-karkare: This exceeds 80 characters, since tabs are generally displayed as more than 1 space.

solutions/downloader/DownloadThread.java
```diff
@@ -0,0 +1,91 @@
...
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.RandomAccessFile;
+import java.net.Socket;
+import java.net.URL;
+import java.net.UnknownHostException;
+import java.util.logging.Logger;
+
+public class DownloadThread extends Thread {
+	private final static Logger logger = Logger.getLogger(FileDownloader.class.getName());
```
kaustubh-karkare:
```
private final static Logger logger =
	Logger.getLogger(FileDownloader.class.getName());
```
so that line length limit is respected.

solutions/downloader/DownloadThread.java
```diff
@@ -0,0 +1,91 @@
...
+import java.io.InputStreamReader;
+import java.io.RandomAccessFile;
+import java.net.Socket;
+import java.net.URL;
+import java.net.UnknownHostException;
+import java.util.logging.Logger;
+
+public class DownloadThread extends Thread {
+	private final static Logger logger = Logger.getLogger(FileDownloader.class.getName());	
+	private final static String newLine = "\r\n";
```
kaustubh-karkare: Constants are generally stored in CAPITAL_LETTER variables like the next line. Consistency.

solutions/downloader/FileDownloader.java
```diff
@@ -0,0 +1,160 @@
...
+  */
+package downloader;
+
+import java.io.BufferedReader;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.MalformedURLException;
+import java.net.Socket;
+import java.net.URL;
```
kaustubh-karkare: For this assignment, you cant use this. Lets do regex instead.

solutions/downloader/FileDownloader.java
```diff
@@ -0,0 +1,160 @@
...
+				}				
+				// Extract content length
+				if (header.startsWith("Content-Length:")) {
+					contentLength = Integer.parseInt(header.substring(header.indexOf(' ') + 1));
+				}
+				
+				if (header.equals("")) {
+					isHeader = false;					
+				}				
+			}
```
kaustubh-karkare: Both in case of request & response, you should maintain HashMap<string,string> containing key value pairs, and then serializing / extracting the necessary data. That is much more flexible than this string concatenation.

solutions/downloader/FileDownloader.java
```diff
@@ -0,0 +1,160 @@
...
+			
+			// Starting multiple worker threads to download parts of the file simultaneously.
+			for (int ii = 0; ii < threadCount; ii++) {				
+				if (ii < threadCount - 1) {
+					downloadThreads[ii] = new DownloadThread(startByte, endByte);						
+				}	
+				
+				else {
+					downloadThreads[ii] = new DownloadThread(startByte, endByte + remain);
+				}
```
kaustubh-karkare: endByte + (last ? remain : 0)

solutions/downloader/DownloadThread.java
```diff
@@ -0,0 +1,91 @@
...
+	private BufferedInputStream inFromServer = null;
+	private DataOutputStream outToServer = null;
+	private RandomAccessFile outputFile;	
+	
+	public DownloadThread(int startByte, int endByte) {		
+		this.startByte = startByte;
+		this.endByte = endByte;
+		
+		try {
+			outputFile = new RandomAccessFile(FileDownloader.fileName, "rw");
```
kaustubh-karkare: You're creating a circular dependency, which you want to avoid. Get the file name as an argument.

Also, could you confirm that this allows you to write at multiple points in the file across threads simultaneously?

solutions/downloader/DownloadThread.java
```diff
@@ -0,0 +1,91 @@
...
+		} catch (FileNotFoundException e) {
+			logger.warning(e.getMessage());
```
kaustubh-karkare: Reraise an exception? This silent failure is dangerous. Sure you can figure it out through logging, but if 1 of the 100 threads waits, there's no easy way for your application to detect and handle it.

solutions/downloader/DownloadThread.java
```diff
@@ -0,0 +1,91 @@
...
+	public void run() {
+		byte data[] = new byte[BUFFER_SIZE];
+		int numRead;
+		int offset = 0;	
+		boolean isHeader = true;
+		
+		try {
+			socket = new Socket(FileDownloader.host, FileDownloader.port);
+			inFromServer = new BufferedInputStream(socket.getInputStream());
+			outToServer = new DataOutputStream(socket.getOutputStream());
```
kaustubh-karkare: Can you abstract away all this boilerplate code? Keep this simple to read.

solutions/downloader/FileDownloader.java
```diff
@@ -0,0 +1,209 @@
...
+		if (verifiedURL == null) {
+			logger.warning("Invalid URL");
+			return;
+		}
+		logger.info("URL verified");
+
+		urlInfo.fileName  = verifiedURL.substring(verifiedURL.lastIndexOf('/') + 1);
+		urlInfo.port = getPort(verifiedURL);
+		urlInfo.filePath = getPath(verifiedURL);
+		urlInfo.host = getHost(url);
```
kaustubh-karkare: You should use a single regex for both validation and extraction. You can have multiple saved groups inside a regex.

solutions/downloader/FileDownloader.java
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	@Override
+	public void run() {
+		boolean isHeader = true;
+		String header;
+		String status = null;
+		LinkedHashMap<String, String> headers = new LinkedHashMap<String, String>();
+		headers.put("GET", urlInfo.filePath + " HTTP/1.1");
+		headers.put("Host:", urlInfo.host);
```
kaustubh-karkare: You shouldn't have ":" in the key. It's okay to make the first line a special case to make it possible.

solutions/downloader/FileDownloader.java
```diff
@@ -0,0 +1,209 @@
...
+		try {
+			connectToServer();
+		} catch (IOException e) {
+			logger.warning(e.getMessage());
+			return;
+		}
+
+		try {
+			// Sending "GET" request in order to extract information from the Response headers.
+			for (Map.Entry h : headers.entrySet()) {
```
kaustubh-karkare: I will always complain on variable names like this.

solutions/downloader/FileDownloader.java
```diff
@@ -0,0 +1,209 @@
...
+		return "";
+	}
+
+	// Method used for establishing connection to server
+	void connectToServer() throws IOException {
+		socket = new Socket(urlInfo.host, urlInfo.port);
+		inFromServer =
+			new BufferedReader(new InputStreamReader(socket.getInputStream()));
+		outToServer = new DataOutputStream(socket.getOutputStream());
+	}
```
kaustubh-karkare: IMO, you shouldn't be using sockets in this class at all. Make the other one (DownloadThread) more flexible so that you can perform HEAD requests. That way, you can avoid this duplication.

solutions/downloader/DownloadThread.java
```diff
@@ -0,0 +1,114 @@
...
+          outputFile.close();
+        } catch(IOException e) {}
+      }
+      if (inFromServer != null) {
+        try {
+          inFromServer.close();
+        } catch(IOException e) {}
+      }
+		}
+	}
```
kaustubh-karkare: What is going on with this indenting.

solutions/downloader/FileDownloader.java
divyanshjr: Actually, I have used it just to find the content length so that I could set ranges for each of the thread. I couldn't find another way to get that info.

solutions/downloader/FileDownloader.java
divyanshjr: I'll try to do that,

solutions/downloader/FileDownloader.java
divyanshjr: :+1:

solutions/downloader/FileDownloader.java
divyanshjr: I'll change it.. :P

solutions/downloader/FileDownloader.java
kaustubh-karkare: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD

divyanshjr: So basically each thread should 1st perform HEAD request, determine the range it is responsible for and then finally send GET request with that range. Am I right?

kaustubh-karkare: No, just use a single HEAD request to get the content length. Based on that decide the range for each GET. And finally start all the GETs in parallel.

kaustubh-karkare:
Inactive for too long.

PR #23, opened 2017-03-23
Assignment 2.1 submitted

lalrishav:
Please review

### DIFF
--- a/solutions/lal-rishav/Assignment-2.1/client.js
+++ b/solutions/lal-rishav/Assignment-2.1/client.js
@@ -1,0 +1,19 @@
+var net = require('net');
+var port = 8080;
+var client = new net.Socket();
+var a;
+client.connect(port, '127.0.0.1', function(data) {
+	 a = process.argv[2]+" "+process.argv[3];
+	client.write(a);
+});
+
+client.on('data', function(data) {
+	console.log('Received: ' + data);
+	if((parseInt(String(a)[0])+parseInt(String(a)[2]))==(parseInt(String(a)[0])+parseInt(String(a)[2])))
+		console.log('verified');
+	client.destroy(); // kill client after server's response
+});
+
+client.on('close', function() {
+	console.log('Connection closed');
+});
--- a/solutions/lal-rishav/Assignment-2.1/client.js
+++ b/solutions/lal-rishav/Assignment-2.1/client.js
@@ -17,23 +17,30 @@ newClient = function() {
 	    return client;
 }
 
-let client = newClient();
+sendNumbers = function(firstNumber,secondNumber,client) {
+    numbers = String(firstNumber) + " " + String(secondNumber);
+	client.write(numbers);
+}
 
+verifySum = function(data) {
+	logger.info('Received: ' + data);
+	numbers = numbers.split(" ");
+	let sum = String(parseInt(numbers[0]) + parseInt(numbers[1]));
+	if (sum == data)
+		logger.info('verified');
+}
+
+let client = newClient();
 const ip = program.ip;
 const port = program.port;
 let numbers;
 
 client.connect(port, ip, function(data) {
- 	numbers = String(program.firstNumber) + " " + String(program.secondNumber);
- 	client.write(numbers);
+	sendNumbers(program.firstNumber,program.secondNumber,client) 
 });
 
 client.on('data', function(data) {
-	logger.info('Received: ' + data);
-	numbers = numbers.split(" ")
-	let sum = String(parseInt(numbers[0]) + parseInt(numbers[1]));
-	if (sum == data)
-		logger.info('verified');
+	verifySum(data);
 	client.destroy(); 
 });
 
--- a/solutions/lal-rishav/Assignment-2.1/client.js
+++ b/solutions/lal-rishav/Assignment-2.1/client.js
@@ -1,19 +1,23 @@
-var net = require('net');
-var port = 8080;
-var client = new net.Socket();
-var a;
+let net = require('net');
+let readline = require('readline')
+let client = new net.Socket();
+let port = process.argv[2]
+let numbers
+
 client.connect(port, '127.0.0.1', function(data) {
-	 a = process.argv[2]+" "+process.argv[3];
-	client.write(a);
+ 	numbers = process.argv[3] + " " + process.argv[4];
+ 	client.write(numbers);
 });
 
 client.on('data', function(data) {
 	console.log('Received: ' + data);
-	if((parseInt(String(a)[0])+parseInt(String(a)[2]))==(parseInt(String(a)[0])+parseInt(String(a)[2])))
+
+	if((parseInt(String(numbers)[0])+parseInt(String(numbers)[2])) == data)
 		console.log('verified');
-	client.destroy(); // kill client after server's response
+	client.destroy(); 
 });
 
 client.on('close', function() {
 	console.log('Connection closed');
-});
+}); 
+
--- a/solutions/lal-rishav/Assignment-2.1/client.js
+++ b/solutions/lal-rishav/Assignment-2.1/client.js
@@ -3,30 +3,37 @@ let readline = require('readline');
 let Logging = require('js-logging');
 let program = require('commander');
 
-program
-  .version('0.0.1')
-  .option('-p, --port <n>', 'Port as argument', parseInt)
-  .option('-i, --ip <n>', 'IP as argument', String)
-  .option('-fn, --fn <n>', 'First Number', parseInt)
-  .option('-sn, --sn <n>', 'SecondNumber', parseInt)
-  .parse(process.argv);
-
 let logger = new Logging();
-let client = new net.Socket();
 
-let ip = program.ip;
-let port = program.port;
+newClient = function(){
+	 program
+    .version('0.0.1')
+    .option('-p, --port <>', 'Port as argument', parseInt)
+    .option('-i, --ip <>', 'IP as argument', String)
+    .option('-fn, --firstNumber <>', 'First Number', parseInt)
+    .option('-sn, --secondNumber <>', 'Second Number', parseInt)
+    .parse(process.argv);
+     let client = new net.Socket();
+     let logger = new Logging();
+     return client
+}
+
+let client = newClient()
+
+const ip = program.ip;
+const port = program.port;
 let numbers;
 
 client.connect(port, ip, function(data) {
- 	numbers = String(program.fn) + " " + String(program.sn);
+ 	numbers = String(program.firstNumber) + " " + String(program.secondNumber);
  	client.write(numbers);
 });
 
 client.on('data', function(data) {
 	logger.info('Received: ' + data);
-	let sum = parseInt(String(numbers)[0])+parseInt(String(numbers)[2]);
-	if( sum == data)
+	numbers = numbers.split(" ")
+	let sum = String(parseInt(numbers[0]) + parseInt(numbers[1]));
+	if (sum == data)
 		logger.info('verified');
 	client.destroy(); 
 });
--- a/solutions/lal-rishav/Assignment-2.1/client.js
+++ b/solutions/lal-rishav/Assignment-2.1/client.js
@@ -1,23 +1,36 @@
 let net = require('net');
-let readline = require('readline')
-let client = new net.Socket();
-let port = process.argv[2]
-let numbers
+let readline = require('readline');
+let Logging = require('js-logging');
+let program = require('commander');
 
-client.connect(port, '127.0.0.1', function(data) {
- 	numbers = process.argv[3] + " " + process.argv[4];
+program
+  .version('0.0.1')
+  .option('-p, --port <n>', 'Port as argument', parseInt)
+  .option('-i, --ip <n>', 'IP as argument', String)
+  .option('-fn, --fn <n>', 'First Number', parseInt)
+  .option('-sn, --sn <n>', 'SecondNumber', parseInt)
+  .parse(process.argv);
+
+let logger = new Logging();
+let client = new net.Socket();
+
+let ip = program.ip;
+let port = program.port;
+let numbers;
+
+client.connect(port, ip, function(data) {
+ 	numbers = String(program.fn) + " " + String(program.sn);
  	client.write(numbers);
 });
 
 client.on('data', function(data) {
-	console.log('Received: ' + data);
-
-	if((parseInt(String(numbers)[0])+parseInt(String(numbers)[2])) == data)
-		console.log('verified');
+	logger.info('Received: ' + data);
+	let sum = parseInt(String(numbers)[0])+parseInt(String(numbers)[2]);
+	if( sum == data)
+		logger.info('verified');
 	client.destroy(); 
 });
 
 client.on('close', function() {
-	console.log('Connection closed');
+	logger.info('Connection closed');
 }); 
-
--- a/solutions/lal-rishav/Assignment-2.1/client.js
+++ b/solutions/lal-rishav/Assignment-2.1/client.js
@@ -5,20 +5,19 @@ let program = require('commander');
 
 let logger = new Logging();
 
-newClient = function(){
+newClient = function() {
 	 program
-    .version('0.0.1')
-    .option('-p, --port <>', 'Port as argument', parseInt)
-    .option('-i, --ip <>', 'IP as argument', String)
-    .option('-fn, --firstNumber <>', 'First Number', parseInt)
-    .option('-sn, --secondNumber <>', 'Second Number', parseInt)
-    .parse(process.argv);
-     let client = new net.Socket();
-     let logger = new Logging();
-     return client
+	    .version('0.0.1')
+	    .option('-p, --port <>', 'Port as argument', parseInt)
+	    .option('-i, --ip <>', 'IP as argument', String)
+	    .option('-fn, --firstNumber <>', 'First Number', parseInt)
+	    .option('-sn, --secondNumber <>', 'Second Number', parseInt)
+	    .parse(process.argv);
+	    let client = new net.Socket();
+	    return client;
 }
 
-let client = newClient()
+let client = newClient();
 
 const ip = program.ip;
 const port = program.port;
--- a/solutions/lal-rishav/Assignment-2.1/server.js
+++ b/solutions/lal-rishav/Assignment-2.1/server.js
@@ -1,0 +1,18 @@
+var net = require('net');
+var host = '127.0.0.1';
+var port = 8080;
+
+net.createServer(function(sock){
+	console.log('Connected: '+sock.remoteAddress + ':' +sock.remotePort);
+	sock.on('data',function(data){
+		var value = parseInt(String(data)[0])+parseInt(String(data)[2]);
+		sock.write(String(value));
+	})
+	sock.on('close',function(data){
+	console.log('CLOSED')
+})
+}).listen(port,host);
+
+
+
+console.log('server listining on port '+port);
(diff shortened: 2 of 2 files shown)


solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,19 @@
+var net = require('net');
```
kaustubh-karkare: Use `let` and `const`, instead of `var`.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,19 @@
...
+var port = 8080;
```
kaustubh-karkare: Instead of assuming that the host/port is fixed, you should take it as a command line option.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,19 @@
...
+var client = new net.Socket();
+var a;
+client.connect(port, '127.0.0.1', function(data) {
+	 a = process.argv[2]+" "+process.argv[3];
+	client.write(a);
```
kaustubh-karkare: There is inconsistent indenting in the previous lines. Choose something, and then stick to it.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,19 @@
+var net = require('net');
+var port = 8080;
+var client = new net.Socket();
+var a;
+client.connect(port, '127.0.0.1', function(data) {
+	 a = process.argv[2]+" "+process.argv[3];
```
kaustubh-karkare: According to the style guide, there should be spaces around operators like `+`.
Please read the style guide.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,19 @@
...
+	client.write(a);
+});
+
+client.on('data', function(data) {
+	console.log('Received: ' + data);
+	if((parseInt(String(a)[0])+parseInt(String(a)[2]))==(parseInt(String(a)[0])+parseInt(String(a)[2])))
```
kaustubh-karkare: This violates the 80 char limit on lines, also recommended by the style guide.
And I don't even understand what you're doing here. That is a sign that readability is a problem.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,19 @@
+var net = require('net');
+var port = 8080;
+var client = new net.Socket();
+var a;
```
kaustubh-karkare: `a` is a bad name, because it tells me nothing about what the variable contains.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,19 @@
...
+client.connect(port, '127.0.0.1', function(data) {
+	 a = process.argv[2]+" "+process.argv[3];
+	client.write(a);
+});
+
+client.on('data', function(data) {
+	console.log('Received: ' + data);
+	if((parseInt(String(a)[0])+parseInt(String(a)[2]))==(parseInt(String(a)[0])+parseInt(String(a)[2])))
+		console.log('verified');
+	client.destroy(); // kill client after server's response
```
kaustubh-karkare: This comment is unnecessary. I conclude the same thing by reading the actual line.

solutions/lal-rishav/Assignment-2.1/server.js
```diff
@@ -0,0 +1,18 @@
+var net = require('net');
+var host = '127.0.0.1';
+var port = 8080;
+
+net.createServer(function(sock){
+	console.log('Connected: '+sock.remoteAddress + ':' +sock.remotePort);
+	sock.on('data',function(data){
+		var value = parseInt(String(data)[0])+parseInt(String(data)[2]);
```
kaustubh-karkare: Lets not assume that the only thing that the client wants is to add 2 numbers.
Make it flexible, and allow that to be specified as an input. Ideal case, your code looks like:
```
const server = new Server(host, port);
server.start(function onClientConnection(...) { ... })
```

Which means that your server can do more than exactly 1 thing. Solving only the given problem works in programming contests, but here, you have to anticipating that requirements will change, and make it easy to handle that.

solutions/lal-rishav/Assignment-2.1/server.js
```diff
@@ -0,0 +1,18 @@
...
+		sock.write(String(value));
+	})
+	sock.on('close',function(data){
+	console.log('CLOSED')
+})
```
kaustubh-karkare: Inconsistent indenting. The previous 2 lines should both have an extra tab.

solutions/lal-rishav/Assignment-2.1/server.js
```diff
@@ -0,0 +1,18 @@
+var net = require('net');
+var host = '127.0.0.1';
+var port = 8080;
+
+net.createServer(function(sock){
```
kaustubh-karkare: Avoid abbreviations, since like bad names, they make it hard to figure out what this contains.
In small programs like this, it doesn't really matter. But in large ones, it becomes very important.

solutions/lal-rishav/Assignment-2.1/server.js
```diff
@@ -0,0 +1,18 @@
...
+	console.log('Connected: '+sock.remoteAddress + ':' +sock.remotePort);
```
kaustubh-karkare: Spaces around operators.

solutions/lal-rishav/Assignment-2.1/server.js
```diff
@@ -0,0 +1,18 @@
...
+	sock.on('data',function(data){
+		var value = parseInt(String(data)[0])+parseInt(String(data)[2]);
+		sock.write(String(value));
+	})
+	sock.on('close',function(data){
+	console.log('CLOSED')
```
kaustubh-karkare: Missing semicolon.

solutions/lal-rishav/Assignment-2.1/server.js
```diff
@@ -0,0 +1,18 @@
...
+})
+}).listen(port,host);
+
+
```
kaustubh-karkare: Too many unnecessary lines.

solutions/lal-rishav/Assignment-2.1/server.js
lalrishav: do you mean that in this case i have to specify that client also send his requirement to the server that weather he wants to add or do other operations ??

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,23 @@
+let net = require('net');
+let readline = require('readline')
+let client = new net.Socket();
+let port = process.argv[2]
+let numbers
+
+client.connect(port, '127.0.0.1', function(data) {
```
kaustubh-karkare: Why isn't the IP address configurable?

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,23 @@
+let net = require('net');
+let readline = require('readline')
+let client = new net.Socket();
+let port = process.argv[2]
```
kaustubh-karkare: Instead of accessing `argv` directly, use a proper cli options parser module (like the one we built in problem 1.2).

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,23 @@
...
+let numbers
+
+client.connect(port, '127.0.0.1', function(data) {
+ 	numbers = process.argv[3] + " " + process.argv[4];
```
kaustubh-karkare: Use random numbers here.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,23 @@
+let net = require('net');
+let readline = require('readline')
+let client = new net.Socket();
```
kaustubh-karkare: Separate your require statements from actual logic.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,23 @@
...
+client.on('data', function(data) {
+	console.log('Received: ' + data);
+
+	if((parseInt(String(numbers)[0])+parseInt(String(numbers)[2])) == data)
+		console.log('verified');
+	client.destroy(); 
+});
+
+client.on('close', function() {
+	console.log('Connection closed');
```
kaustubh-karkare: Use https://www.npmjs.com/package/js-logging instead of console methods.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,23 @@
...
+
+client.connect(port, '127.0.0.1', function(data) {
+ 	numbers = process.argv[3] + " " + process.argv[4];
+ 	client.write(numbers);
+});
+
+client.on('data', function(data) {
+	console.log('Received: ' + data);
+
+	if((parseInt(String(numbers)[0])+parseInt(String(numbers)[2])) == data)
```
kaustubh-karkare: Please abide by style guide for formatting. My comments apply in general across your code, not just to the 1 place where I pointed them out.

solutions/lal-rishav/Assignment-2.1/server.js
```diff
@@ -0,0 +1,20 @@
...
+	socket.on('data',function(data){
+		let value = parseInt(String(data)[0]) + parseInt(String(data)[2]);
+		socket.write(String(value));
+	})
+	socket.on('close',function(data){
+	console.log('CLOSED')
+	})
+}).listen(port,host);
+
+
```
kaustubh-karkare: Too many empty lines. 1 is sufficient and intuitive. Anything more is not.

solutions/lal-rishav/Assignment-2.1/client.js
lalrishav: i really dont have any idea what the problem with argv directly..Will you please explain the probelm i can face due to argv

kaustubh-karkare: You're assuming that the user will provide the arguments in a very specify sequence (ie - the port has to be the 2nd argument) without telling them that. In order to make your code reusable, expect that people will use your code without reading it, which means they should be able to provide a `--help` argument and figure out the inputs. And existing libraries already have that functionality.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -1,23 +1,36 @@
...
+let readline = require('readline');
+let Logging = require('js-logging');
+let program = require('commander');
+
+program
+  .version('0.0.1')
+  .option('-p, --port <n>', 'Port as argument', parseInt)
+  .option('-i, --ip <n>', 'IP as argument', String)
+  .option('-fn, --fn <n>', 'First Number', parseInt)
+  .option('-sn, --sn <n>', 'SecondNumber', parseInt)
```
kaustubh-karkare: Please be consistent. If "First Number" had a space in between the words, why not the 2nd? I'm very particular about this.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -1,23 +1,36 @@
...
+
+let logger = new Logging();
 let client = new net.Socket();
-let port = process.argv[2]
-let numbers
 
-client.connect(port, '127.0.0.1', function(data) {
- 	numbers = process.argv[3] + " " + process.argv[4];
+let ip = program.ip;
+let port = program.port;
```
kaustubh-karkare: These can be `const` right?

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -1,23 +1,36 @@
...
+let numbers;
+
+client.connect(port, ip, function(data) {
+ 	numbers = String(program.fn) + " " + String(program.sn);
```
kaustubh-karkare: fn / sn are abbreviations, which means that they're harder to understand, and therefore bad names.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -1,23 +1,36 @@
...
+program
+  .version('0.0.1')
+  .option('-p, --port <n>', 'Port as argument', parseInt)
+  .option('-i, --ip <n>', 'IP as argument', String)
+  .option('-fn, --fn <n>', 'First Number', parseInt)
+  .option('-sn, --sn <n>', 'SecondNumber', parseInt)
+  .parse(process.argv);
+
+let logger = new Logging();
 let client = new net.Socket();
```
kaustubh-karkare: Shouldn't all this be done in a function? To make it resuable, you need to handle the case where someone will want to use this module for creating more than 1 client.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -1,23 +1,36 @@
...
  	client.write(numbers);
 });
 
 client.on('data', function(data) {
-	console.log('Received: ' + data);
-
-	if((parseInt(String(numbers)[0])+parseInt(String(numbers)[2])) == data)
-		console.log('verified');
+	logger.info('Received: ' + data);
+	let sum = parseInt(String(numbers)[0])+parseInt(String(numbers)[2]);
```
kaustubh-karkare: - s/let/const.
- Spaces around operators.
- Are you assuming that the numbers are single digits? Why?

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -1,23 +1,36 @@
...
+	if( sum == data)
```
kaustubh-karkare: There should be space before `(`, not after. Also, there should always be braces.

solutions/lal-rishav/Assignment-2.1/server.js
```diff
@@ -1,20 +1,27 @@
...
+let readline = require('readline');
+let Logging = require('js-logging');
+let program = require('commander');
 
-let host = '127.0.0.1';
-let port = process.argv[2];
+program
+  .version('0.0.1')
+  .option('-p, --port <n>', 'Port as argument', parseInt)
+  .option('-i, --ip <n>', 'ip as argument', String)
```
kaustubh-karkare: What does `<n>` represent?

solutions/lal-rishav/Assignment-2.1/server.js
```diff
@@ -1,20 +1,27 @@
...
 net.createServer(function(socket){
-	console.log('Connected: '+socket.remoteAddress + ':' + socket.remotePort);
+	logger.info('Connected: '+ socket.remoteAddress + ':' + socket.remotePort);
 	socket.on('data',function(data){
 		let value = parseInt(String(data)[0]) + parseInt(String(data)[2]);
 		socket.write(String(value));
 	})
 	socket.on('close',function(data){
-	console.log('CLOSED')
+	logger.info('Connection Closed')
```
kaustubh-karkare: Need more indenting here.

solutions/lal-rishav/Assignment-2.1/server.js
```diff
@@ -1,20 +1,27 @@
...
 	})
 }).listen(port,host);
 
-
-
-console.log('server listining on port '+port);
+logger.info('server listining on port ' + port);
```
kaustubh-karkare: Do not assume that requiring the module is equivalent to starting the server.
These are non-intuitive side effects.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,43 @@
+let net = require('net');
+let readline = require('readline');
+let Logging = require('js-logging');
+let program = require('commander');
+
+let logger = new Logging();
+
+newClient = function(){
+	 program
```
kaustubh-karkare: The indenting is messed up here.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,43 @@
+let net = require('net');
+let readline = require('readline');
+let Logging = require('js-logging');
+let program = require('commander');
+
+let logger = new Logging();
+
+newClient = function(){
```
kaustubh-karkare: Space before `{`

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,43 @@
...
+	 program
+    .version('0.0.1')
+    .option('-p, --port <>', 'Port as argument', parseInt)
+    .option('-i, --ip <>', 'IP as argument', String)
+    .option('-fn, --firstNumber <>', 'First Number', parseInt)
+    .option('-sn, --secondNumber <>', 'Second Number', parseInt)
+    .parse(process.argv);
+     let client = new net.Socket();
+     let logger = new Logging();
+     return client
```
kaustubh-karkare: Missing semicolon.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+let client = newClient()
```
kaustubh-karkare: You created a function as I mentioned, but it isn't really reusable.

solutions/lal-rishav/Assignment-2.1/server.js
```diff
@@ -0,0 +1,41 @@
...
+	socket.on('close',function(data){
+		logger.info('Connection Closed');
+	})
+}).listen(port,host);
+
+try{
+	server.listen(port,host);
+	logger.info('server listining on port ' + port);
+}
+catch(err){
```
kaustubh-karkare: Space before `{`.
`err` is an obvious abbreviation, but prefer better names as a matter of habit.

solutions/lal-rishav/Assignment-2.1/client.js
lalrishav: What should be the motive to reuse the function???as i think whenever a client want to join the server he will run the client.js and the function newClient will execute.Why to reuse that fuction?

kaustubh-karkare: This isn't good API, since you give the users no control over the client other than requiring the module. Which means that the only thing this client can do is to get 2 numbers from command line arguments and send them to the server.
- What if I wanted to provide the numbers from some other source? Why should this client be restricted to command line arguments?
- What if I wanted to create more than one client using this logic?
- What if I wanted to reuse the `client.connect` functionality to multiply numbers?
Before I assign a problem, I like to solve it a problem, and all 3 things listed above were part of my implementation.

solutions/lal-rishav/Assignment-2.1/client.js
```diff
@@ -0,0 +1,49 @@
...
+
+verifySum = function(data) {
+	logger.info('Received: ' + data);
+	numbers = numbers.split(" ");
+	let sum = String(parseInt(numbers[0]) + parseInt(numbers[1]));
+	if (sum == data)
+		logger.info('verified');
+}
+
+let client = newClient();
```
kaustubh-karkare: This is still not sufficient. How about I give you an example of how it should look like?
```
const Client = require('./client'); // your code
new Client(host, port).connect(function(socket) {
  // add 2 numbers
});
```
It doesn't have to be exactly this, but it does need to satisfy at least the 3 conditions I mentioned before.

kaustubh-karkare:
Inactive for too long.

PR #24, opened 2017-03-24
Created download manager

lalrishav:
Please review!!

### DIFF
--- a/solutions/lal-rishav/assignment-2.1/download.js
+++ b/solutions/lal-rishav/assignment-2.1/download.js
@@ -1,0 +1,55 @@
+var path = require('path');
+var dm = require('mt-files-downloader');
+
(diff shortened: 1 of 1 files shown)


solutions/lal-rishav/assignment-2.1/download.js
```diff
@@ -0,0 +1,55 @@
+var path = require('path');
+var dm = require('mt-files-downloader');
```
kaustubh-karkare: As mentioned in the problem statement, "You are still only allowed to use the basic socket library like in the previous assignment." Which means that things like this, which do all the hard work for you, are not allowed.
As a result, you'll need to completely rewrite this.

kaustubh-karkare:
Inactive for too long.

PR #28, opened 2017-03-30
Solution to Assignment 2.3 added

divyanshjr:
- I've created a HTTP Web Server on Java. It runs on a specified IP and port number (passed as Command Line Argument).

- 2 additional command line arguments can be passed -
1.  LogConfigurationFilePath(Address of the file containing the Log Configuration File) 
2.  Backlog (Requested maximum length of the queue of incoming connections)

- All the 4 command line arguments are optional. Default values will be used in case they are not provided.

- Assignment2_3.java contains the driver code which uses the httpserver API. So this file is to be run in order to start the server. The other 3 files namely - HttpServer.java, HttpServerBuilder.java and Clienthandler.java are part of httpserver API and must be placed in a directory "httpserver" which should be in the current directory of Assignment2_3.java

- ***Details about the functionality of the httpserver API  -**

- The HttpServer instance is created using the builder pattern technique and it is implemented with the help of HttpServerBuilder class.

- HttpServer is the server's main thread where the server keeps listening and accepts incoming clients.
As soon as a Client gets connected, the server redirects the Client to ClientHandler thread which reads the requests from the Client and handles them. HttpServer keeps waiting for other clients to connect after handling the request to ClientHandler thread.

- The ClientHandler thread checks if the request if of "GET" type. If not it returns a 405 Status Code to the Client.
 If it is of "GET" type, it checks whether the Client seeks for a file or directory.
 If the Client seeks for a file, it checks whether the file exists. If it exists it returns the file else return a 404 Status Code.
 If the Client seeks for a directory, it checks if an index.html is present in that directory. If found, it returns that index.html to the client. Else, it generates a list of files and folders in that directory and returns that list.

**This interface has been tested on web browsers and is fully functional**

_PS - You should have Argparse4j package (3rd party API) installed on your machine in order to run this Java HTTP Server._


solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+    try {
+      port = Integer.parseInt(args[0]);
+      serverSocket = new ServerSocket(port); // Start,listen on given port
+      while (true) {
+        try {
+          logger.info("TCPServer Waiting for client on port " + port);
+          Socket socket = serverSocket.accept(); // Wait for a client to connect
+          new ClientHandler(socket); // A separate thread to handle each client
+        } catch (IOException e) {
+            e.printStackTrace();
```
kaustubh-karkare: Please use consistent indenting. This is 4 space in, while the previous block was 2 spaces in.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+    fileHandler.setFormatter(new SimpleFormatter());
+    logger.addHandler(fileHandler);
+    logger.setLevel(Level.FINE);
+  }
+
+  public static void main(String args[]) throws IOException {
+    initLog();
+    try {
+      port = Integer.parseInt(args[0]);
+      serverSocket = new ServerSocket(port); // Start,listen on given port
```
kaustubh-karkare: Instead of this, a better API would be:
```
HTTPServer server = new HTTPServer(port);
server.start();
time.sleep(60);
server.stop();
```
Not sure how you provide the `ClientHandler` as input to this server object, but you'll figure it out.

This is also why I'd rather have your main function in a separate file, so that you're forced to see how your API might make this inconvenient to use.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+        try {
+          logger.info("TCPServer Waiting for client on port " + port);
+          Socket socket = serverSocket.accept(); // Wait for a client to connect
+          new ClientHandler(socket); // A separate thread to handle each client
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+      }
+    } catch (NumberFormatException e) {
+        logger.warning("Port number should be a positive integer");
```
kaustubh-karkare: Once again, inconsistent indenting. But importantly, you're just ignoring this. I'd think it's better to raise the exception, since otherwise the calling method would have no idea that something was wrong.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+  }
+}
+
+// A ClientHandler reads an HTTP request and responds
+class ClientHandler extends Thread {
+  private static final Logger logger = Logger.getLogger(ClientHandler.class.getName());
+  static final String HTML_START = "<html>" + "<title>HTTP Server in java</title>" +
+    "<body>";
```
kaustubh-karkare: You should try and make these easier to read:
```
static final String HTML_START =
  "<html>" + 
    "<title>HTTP Server in java</title>" +
    "<body>";

File: solutions/Assignment-2.3/HttpServer.java [hunk 5]
```diff
@@ -0,0 +1,200 @@
...
+    this.connectedClient = connectedClient;
+    start();
+  }
+
+  @Override
+  public void run() {
+
+    try {
+      logger.info("The Client " + connectedClient.getInetAddress() +
+        ":" + connectedClient.getPort() + " is connected");
```
kaustubh-karkare: You know how I say that if you cant fit the entire function signature/call in one line, each part should be in its own line? It makes this easier to read:
```
logger.info(
  "The Client " +
  connectedClient.getInetAddress() +
  ":" +
  connectedClient.getPort() +
  " is connected"
);

File: solutions/Assignment-2.3/HttpServer.java [hunk 6]
```diff
@@ -0,0 +1,200 @@
...
+      inFromClient = new BufferedReader(new InputStreamReader (connectedClient.getInputStream()));
```
kaustubh-karkare: Should there be a space after `InputStreamReader` ?

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+      outToClient = new DataOutputStream(connectedClient.getOutputStream());
+      String requestString = inFromClient.readLine();
+      String headerLine = requestString;
+      StringTokenizer tokenizer = new StringTokenizer(headerLine);
+      String httpMethod = tokenizer.nextToken();
+      String httpQueryString = tokenizer.nextToken();
+      StringBuffer responseBuffer = new StringBuffer();
+      logger.info("The HTTP request string is...<BR>");
```
kaustubh-karkare: `<BR>` ?

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+  public void run() {
+
+    try {
+      logger.info("The Client " + connectedClient.getInetAddress() +
+        ":" + connectedClient.getPort() + " is connected");
+      inFromClient = new BufferedReader(new InputStreamReader (connectedClient.getInputStream()));
+      outToClient = new DataOutputStream(connectedClient.getOutputStream());
+      String requestString = inFromClient.readLine();
+      String headerLine = requestString;
+      StringTokenizer tokenizer = new StringTokenizer(headerLine);
```
kaustubh-karkare: Java has too many of these helper classes. Any chance you could just use strings? I'd rather you know how to work without these.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+      String httpMethod = tokenizer.nextToken();
+      String httpQueryString = tokenizer.nextToken();
+      StringBuffer responseBuffer = new StringBuffer();
+      logger.info("The HTTP request string is...<BR>");
+      responseBuffer.append("<b> This is the HTTP Server Home Page.... </b><BR>");
+      responseBuffer.append("Created by DJ<BR>");
```
kaustubh-karkare: lol. You should use a custom header for this. But actually, this reveals a bigger problem: you're assuming that the response will always be HTML, which isn't true. What if we want to download PDFs through this?
Update: I just reached the point below where I see that the `responseBuffer` is only used on path = `/`. Lets not do that either. Unnecessary special casing just leads to complications.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+      while (inFromClient.ready()) {
+        // Read the HTTP Complete Query
+        logger.info(requestString);
```
kaustubh-karkare: There is such a thing as "too much logging". The point of logging is to provide enough data to figure whats wrong or where to look. But if you log every byte of request, you're spamming the logs so much that nobody will bother reading them, and it'll be equivalent to not having logs, but taking up unnecessary space.
Yes, it is possible that this information could be useful, but it's okay to rely on people to add their own debugging statements later. It's a balance, but this definitely is beyond the limits.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+          if (fileName.endsWith("/")) {
+            String indexFile = fileName + "index.html";
+
+            // if index.html exists
+            if (new File(indexFile).isFile()) {
+              sendResponse(200, indexFile, true);
+            }
+            // if index.html does not exists, we send the list of files and folders in that directory
+            else {
+              try {
```
kaustubh-karkare: You should avoid large functions because they make code harder to read. This section looks like a logically independent part and should be its own function.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+                    response.append("File - " + listOfFiles[i].getName() + "<BR>");
+                  }
+                  else if (listOfFiles[i].isDirectory()) {
+                    response.append("Folder -  " + listOfFiles[i].getName() + "<BR>");
+                  }
+                }
+                sendResponse(200, response.toString(), false);
+              } catch (NullPointerException e) {
+                sendResponse(404, "<b>" +
+                  "Error 404: The Requested resource not found ....</b>", false);
```
kaustubh-karkare: No need for responses body in our case. The status code is good enough.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+                for (int i = 0; i < listOfFiles.length; i++) {
+                  if (listOfFiles[i].isFile()) {
+                    response.append("File - " + listOfFiles[i].getName() + "<BR>");
+                  }
+                  else if (listOfFiles[i].isDirectory()) {
+                    response.append("Folder -  " + listOfFiles[i].getName() + "<BR>");
+                  }
+                }
+                sendResponse(200, response.toString(), false);
+              } catch (NullPointerException e) {
```
kaustubh-karkare: It looks like this catches an exception thrown in line 111 / 112, but I'm not sure.
The fact that I'm not sure means that by catching exceptions in the entire block when 1-2 lines would have been sufficient, you've decreased the readability of your code. Lets fix that.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+            // if index.html does not exists, we send the list of files and folders in that directory
+            else {
+              try {
+                File folder = new File(fileName.substring(0, fileName.lastIndexOf('/')));
+                File[] listOfFiles = folder.listFiles();
+                StringBuffer response = new StringBuffer();
+                response.append("<b>List of files and folders</b> <BR>");
+                for (int i = 0; i < listOfFiles.length; i++) {
+                  if (listOfFiles[i].isFile()) {
+                    response.append("File - " + listOfFiles[i].getName() + "<BR>");
```
kaustubh-karkare: Add `<a>` tags so that you can click. Also maybe `<li>` tags for looking nice.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+      statusLine = "HTTP/1.1 200 OK" + "\r\n";
+    else
+      statusLine = "HTTP/1.1 404 Not Found" + "\r\n";
+
+    if (isFile) {
+      fileName = responseString;
+      fileInputStream = new FileInputStream(fileName);
+      contentLengthLine = "Content-Length: " + Integer.toString(fileInputStream.available()) + "\r\n";
+      if (!fileName.endsWith(".htm") && !fileName.endsWith(".html"))
+      contentTypeLine = "Content-Type: \r\n";
```
kaustubh-karkare: Why is this empty?

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+      outToClient.writeBytes(responseString);
+
+    outToClient.close();
+  }
+
+  public void sendFile(FileInputStream fileInputStream, DataOutputStream out) throws IOException {
+    byte[] buffer = new byte[1024] ;
+    int bytesRead;
+
+    while ((bytesRead = fileInputStream.read(buffer)) != -1 ) {
```
kaustubh-karkare: No need for space before `)`

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,200 @@
...
+  // Method to create log file
+  public static void initLog(){
+    try {
+      fileHandler = new FileHandler("Status.log", false);
+    }
+    catch (SecurityException | IOException e) {
+      e.printStackTrace();
+    }
+    Logger logger = Logger.getLogger("");
+    fileHandler.setFormatter(new SimpleFormatter());
```
divyanshjr: If there is an exception that means the log file could not be created. What should I do in such a case?

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Alright, I'll keep the main mathod in a separate file and try to use the API.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Okay, I will throw an user defined exception.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: (y)

divyanshjr: Alright

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Okay..

divyanshjr: :+1:

solutions/Assignment-2.3/HttpServer.java
divyanshjr: lol.. Okay, I'll try to use Strings only.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: oops.. :P

solutions/Assignment-2.3/HttpServer.java
divyanshjr: I and sending responseBuffer if the requested link is the homepage itself. Okay so I will check if the QueryString is "/" or empty and then send the responseBuffer. But how will I figure out what will be the response type for the homepage?

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Yup, I will remove unnecessarylogging.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Sure, I'll do..

solutions/Assignment-2.3/HttpServer.java
divyanshjr: :+1:

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Alright..

solutions/Assignment-2.3/HttpServer.java
divyanshjr: :+1:

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Will add..

solutions/Assignment-2.3/HttpServer.java
kaustubh-karkare: What do you think is the right thing to do? I will not alway shave answers for you.

solutions/Assignment-2.3/HttpServer.java
kaustubh-karkare: I'm saying that a homepage shouldn't be special cased in your webserver. If you want it so much, create an index.html file in your current directory.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+  static String inputPort;
+  static int port;
+
+  public HttpServer(String inputPort) {
+    HttpServer.inputPort = inputPort;
+  }
+
+  // Method to create log file
+  public static void initLog() throws SecurityException, IOException {
+    fileHandler = new FileHandler("Status.log", false);
```
kaustubh-karkare: Do not assume that all logs will be written to file. Can you make this configurable too?

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+    logger.setLevel(Level.FINE);
+  }
+  // The run() method of HttpServer
+  @Override
+  public void run() {
+    try {
+      initLog();
+      try {
+        port = Integer.parseInt(inputPort);
+        if (port <= 0)
```
kaustubh-karkare: According to most style guides, you should always use braces around if/while/for blocks.
Also, I'm not sure if `0` is a bad thing here.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+    Logger logger = Logger.getLogger("");
+    fileHandler.setFormatter(new SimpleFormatter());
+    logger.addHandler(fileHandler);
+    logger.setLevel(Level.FINE);
+  }
+  // The run() method of HttpServer
+  @Override
+  public void run() {
+    try {
+      initLog();
```
kaustubh-karkare: Shouldn't this be done in the constructor?

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+      } catch (NumberFormatException | NegativeNumberException e) {
+        logger.warning("Port number should be a positive integer");
+      } catch (ArrayIndexOutOfBoundsException e) {
+        logger.warning("Port number must be entered as command line argument");
+      } catch (IOException e) {
+        System.out.println("Unable to start server on port " + port);
+      }
+    } catch (SecurityException | IOException e) {
+      System.out.println("Unable to create log file. Can't start server");
```
kaustubh-karkare: Use `logger.error`, and then throw some exception, not sure what. But basically, dont print to stdout. The logging argument still applies.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+    "<title>HTTP Server in java</title>" +
+    "<body>";
+  static final String HTML_END = "</body>" + "</html>";
+  private Socket connectedClient = null;
+  private BufferedReader inFromClient = null;
+  private DataOutputStream outToClient = null;
+
+  // Constructor where thread startsWith
+  public ClientHandler(Socket connectedClient) {
+    this.connectedClient = connectedClient;
```
kaustubh-karkare: I think "socket" is a better name than "connectedClient".

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+  @Override
+  public void run() {
+
+    try {
+      logger.info(
+        "The Client " +
+        connectedClient.getInetAddress() +
+        ":" +
+        connectedClient.getPort() +
+        " is connected");
```
kaustubh-karkare: The `)` should be on the next line.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+      inFromClient = new BufferedReader(new InputStreamReader(connectedClient.getInputStream()));
+      outToClient = new DataOutputStream(connectedClient.getOutputStream());
```
kaustubh-karkare: Since these are related to initialization, do these lines in the constructor.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+      String headerLine = inFromClient.readLine();
+      String httpMethod = headerLine.substring(0,3);
```
kaustubh-karkare: Don't assume that the type of request is always 3 characters.
Use regular expressions or something (will help in problem 3.1 too).

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+      String httpQueryString = headerLine.substring(
+        4, headerLine.indexOf(" ", headerLine.indexOf(" ") + 1));
+
+      if (httpMethod.equals("GET")) {
+        String requestPath = httpQueryString.replaceFirst("/", "");
+        // if request is for a directory
+        if (requestPath.endsWith("/") | httpQueryString.equals("/")) {
```
kaustubh-karkare: Uhh .. this is inelegant. I checked, and Java does have standard modules to join paths, which you might as well take advantage of here.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+          try {
+            sendDirectoryInformation(requestPath);
+          } catch (Exception e) {
+            sendResponse(404, "", false);
```
kaustubh-karkare: Shouldn't this be `500` and not `404`?

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+        // if request is for a file
+        else {
+          if (new File(requestPath).isFile()) {
+            sendResponse(200, requestPath, true);
+          }
+          else {
+            sendResponse(404, "", false);
+          }
+        }
+      }
```
kaustubh-karkare: There's too many if-else conditions here. Can you simplify the nesting?

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  public void sendDirectoryInformation(String requestPath) throws Exception  {
+    String indexFile = requestPath + "index.html";
+    // if index.html exists
+    if (new File(indexFile).isFile()) {
+      sendResponse(200, indexFile, true);
+    }
+    // if index.html does not exists, we send the list of files and folders in that directory
+    else {
```
kaustubh-karkare: Instead of having this `else` clause, prefer an early return. That way, you avoid this additional nesting.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+        folder = new File(requestPath.substring(0, requestPath.lastIndexOf('/')));
+      File[] listOfFiles = folder.listFiles();
+      StringBuffer response = new StringBuffer();
+      response.append("<b>List of files and folders</b> <BR>");
+      response.append("<ul>");
+      for (int i = 0; i < listOfFiles.length; i++) {
+        if (listOfFiles[i].isFile()) {
+          response.append("<li>File - " +
+            "<a href=" +
+            inetAddress.substring(0, inetAddress.indexOf("0.0.0.0")) +
```
kaustubh-karkare: This feels wrong. Why do you need to deal with API addresses? Relative links should be sufficient.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+      // If homepage is requested
+      if (requestPath.equals("")) {
+        requestPath = ".";
+        folder = new File(requestPath);
+      }
+      else
+        folder = new File(requestPath.substring(0, requestPath.lastIndexOf('/')));
+      File[] listOfFiles = folder.listFiles();
+      StringBuffer response = new StringBuffer();
+      response.append("<b>List of files and folders</b> <BR>");
```
kaustubh-karkare: As with most things that I complain about, lets be consistent about how you use tags. If `<b>` was lower case, why is `<BR>` uppercase? On the other hand, I'm not sure you need a <br> anyway.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+      response.append("<ul>");
+      for (int i = 0; i < listOfFiles.length; i++) {
```
kaustubh-karkare: Use `ii`, `jj` and `kk` for loop variables. They're easier to update when you use multiple cursors.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+          response.append("<li>File - " +
+            "<a href=" +
+            inetAddress.substring(0, inetAddress.indexOf("0.0.0.0")) +
+            "/" +
+            requestPath +
+            "/" +
+            listOfFiles[i].getName() +
+            ">" +
+            listOfFiles[i].getName() +
+            "</a></li><BR>");
```
kaustubh-karkare: No need for <br>

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+      }
+      response.append("</ul>");
+      sendResponse(200, response.toString(), false);
+    }
+  }
+
+  public void sendResponse(int statusCode, String responseString, boolean isFile) throws IOException {
```
kaustubh-karkare: Boolean arguments are generally a bad idea, since it forces people to look at the function signature to see what the argument was supposed to meant. Like, when you used it above, "sendResponse(200, response.toString(), false)", the first 2 arguments are obvious, but I had no idea what the 3rd was.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+    String statusLine = null;
+    String serverdetails = "Server: Java HTTPServer";
+    String contentLengthLine = null;
+    String requestPath = null;
+    String contentType = null;
+    String contentTypeLine = "Content-Type: text/html" + "\r\n";
```
kaustubh-karkare: Given the number of types you use "\r\n", you should really put it in a constant.
Additionally, the number of variables you created is scary (ie - large, and so I'm worried about the complexity you have below). Use a HashMap to contain all headers, so that things are easy to read.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+    FileInputStream fileInputStream = null;
+
+    if (statusCode == 200)
+      statusLine = "HTTP/1.1 200 OK" + "\r\n";
+    else
+      statusLine = "HTTP/1.1 404 Not Found" + "\r\n";
```
kaustubh-karkare: Lets not assume that these are the only 2 status codes possible.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,238 @@
...
+
+    outToClient.writeBytes(statusLine);
+    outToClient.writeBytes(serverdetails);
+    outToClient.writeBytes(contentTypeLine);
+    outToClient.writeBytes(contentLengthLine);
+    outToClient.writeBytes("Connection: close\r\n");
+    outToClient.writeBytes("\r\n");
+
+    if (isFile)
+     sendFile(fileInputStream, outToClient);
```
kaustubh-karkare: Inconsistent indenting.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: I'll try to make it configurable..

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Yup, will correct it.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Yup, 0 is not a bad thing. My fault.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Alright.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: :+1:

solutions/Assignment-2.3/HttpServer.java
divyanshjr: :+1:

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Yup. I'll use regex..

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Actually, here I'm just checking if the request is for a directory by checking if the last character is '/'.  And I'll use standard modules to join paths from now on.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Yup..

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Sure..

solutions/Assignment-2.3/HttpServer.java
divyanshjr: My ignorance. And I'll remove <BR>. It's not required here.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: :+1:

solutions/Assignment-2.3/HttpServer.java
divyanshjr: I'll correct that..

solutions/Assignment-2.3/HttpServer.java
divyanshjr: I'll use relative links.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: I'll try to use an alternative. Enum most probably.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: I'll do the changes.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+    File[] listOfFiles = folder.listFiles();
+    StringBuffer response = new StringBuffer();
+    response.append("<b>List of files and folders</b> <br>");
+    response.append("<ul>");
+    for (int ii = 0; ii < listOfFiles.length; ii++) {
+      if (listOfFiles[ii].isFile()) {
+        String path = listOfFiles[ii].toString().replaceAll(" ","&nbsp;");
+        String fileName = listOfFiles[ii].getName();
+        response.append("<li>File - " +
```
kaustubh-karkare: If you cant fit the entire function signature/call on 1 line, keep each item on a separate line.
So the string should be on the next line.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+          "<a href=" +
+          "/" +
+          path +
+          ">" +
+          fileName +
+          "</a></li>");
```
kaustubh-karkare: `)` on next line. Why? https://git-scm.com/docs/git-blame

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+    response.append("</ul>");
+    sendResponse(200, response.toString(), FileSendingDecision.DONT_SEND_FILE);
+  }
+
+  public void sendResponse(int statusCode, String responseString, FileSendingDecision decision)
+    throws IOException {
+    final String newLine = "\r\n";
+    HashMap<String, String> headers = new HashMap<String, String>();
+    headers.put("statusLine", null);
+    headers.put("serverdetails", "Server: Java HTTPServer");
```
kaustubh-karkare: You misunderstood me. There's no purpose of the keys you're using right now.
I meant something like: `headers.put("Server", "Java HTTPServer");`.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+    else {
+      responseString = ClientHandler.HTML_START + responseString + ClientHandler.HTML_END;
+      headers.put("contentLengthLine", "Content-Length: " + responseString.length() + newLine);
+    }
+
+    outToClient.writeBytes(headers.get("statusLine"));
+    outToClient.writeBytes(headers.get("serverdetails"));
+    outToClient.writeBytes(headers.get("contentTypeLine"));
+    outToClient.writeBytes(headers.get("contentLengthLine"));
+    outToClient.writeBytes(headers.get("closeConnection"));
```
kaustubh-karkare: This whole section is inelegant. Given my previous comment, a for-loop should be sufficient.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+
+    if (statusCode == 200) {
+      headers.put("statusLine", "HTTP/1.1 200 OK" + newLine);
+    }
+    else if (statusCode == 404) {
+      headers.put("statusLine", "HTTP/1.1 404 Not Found" + newLine);
+    }
+    else if (statusCode == 500) {
+      headers.put("statusLine", "HTTP/1.1 500 Internal Server Error" + newLine);
+    }
```
kaustubh-karkare: Use a switch statement. And then create a HashMap<int, string> = {200: "OK", 404: "Not Found", ...} to avoid all the duplication in this code.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+        responseString.length()
+      );
+      fileInputStream = new FileInputStream(headers.get("requestPath"));
+      headers.put(
+        "contentLengthLine", "Content-Length: " +
+        Integer.toString(fileInputStream.available()) +
+        newLine
+      );
+      if (!headers.get("requestPath").endsWith(".htm") && !headers.get("requestPath").endsWith(".html"))
+      headers.put("contentTypeLine", "Content-Type: " + contentType + newLine);
```
kaustubh-karkare: This is a major red flag. It's not obvious by looking at the indenting that this is conditional.
And anyway, you should always uses braces even if the if/while/for condition contains only 1 statement.

solutions/Assignment-2.3/Assignment2_3.java
```diff
@@ -0,0 +1,24 @@
...
+class Assignment2_3 {
+
+  public static void main(String args[]) throws IOException {
+    HttpServer httpServer = new HttpServer(args[0]);
+    httpServer.start();
+    try {
+      Thread.sleep(60);
+      httpServer.join();
+    } catch (InterruptedException e) {}
+ }
```
divyanshjr: :P

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+    try {
+      initLog();
+    } catch (SecurityException | IOException e) {
+      logger.severe("Some error occured in loading configuration. Logging through console only");
+    }
+  }
+
+  // Method to create log file
+  public static void initLog() throws SecurityException, IOException {
+    Logger logger = Logger.getLogger("");
```
divyanshjr: I'll correct it.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+    // Reading Configuration file which contains predefined instructions regarding logging
+    logManager.readConfiguration(new FileInputStream("./server.properties"));
```
divyanshjr: I'll use Builder pattern this time.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+  private enum FileSendingDecision {
+    SEND_FILE,
+    DONT_SEND_FILE
+  }
+
+  // Constructor where thread startsWith
+  public ClientHandler(Socket socket) throws IOException {
+    this.socket = socket;
+    inFromClient = new BufferedReader(new InputStreamReader(socket.getInputStream()));
+    outToClient = new DataOutputStream(socket.getOutputStream());
```
divyanshjr: "this" is added implicitly to refer objects of the same class so I don't have to specify it generally But when there is a clash between local variable of a method and class variable, I have to use "this" to distinguish between the 2 variables. Here socket is a local as well as a class variable so I used "this" but inFromClient is a class variable only so I don't need to use "this" explicitly.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+        " is connected"
+      );
+      String headerLine = inFromClient.readLine();      
+      String[] tokens = headerLine.split("\\s");
+      String httpMethod = tokens[0];
+      String httpQueryString = tokens[1].substring(1);
+
+      if (httpMethod.equals("GET")) {
+        String requestPath = httpQueryString;
+        requestPath = requestPath.replaceAll("%C2%A0", "\\ ");
```
divyanshjr: I'll try to implement it.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+        // If requestPath ends with a '/' or is empty, it is a request for a directory
+        if (requestPath.endsWith("/") | httpQueryString.isEmpty()) {
+          try {
+            sendDirectoryInformation(requestPath);
+          } catch (Exception e) {
+            sendResponse(500, "", FileSendingDecision.DONT_SEND_FILE);
```
divyanshjr: Yes, I understand. So I am just sending the 500 Status Code to the user here and nothing else. I forgot to log the error. I'll do it.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        // If request is for a file
+        else {
+          if (new File(requestPath).isFile()) {
+            sendResponse(200, requestPath, FileSendingDecision.SEND_FILE);
+          }
+        }
+      }
+      sendResponse(404, "", FileSendingDecision.DONT_SEND_FILE);
```
divyanshjr: Agree. I'll do it...

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+        socket.getInetAddress() +
+        ":" +
+        socket.getPort() +
+        " is disconnected"
+      );
+    }
+  }
+
+  public void sendDirectoryInformation(String requestPath) throws Exception  {
+    String indexFile = requestPath + "index.html";
```
divyanshjr: Alright, I'll do that.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+    // if index.html exists
+    if (new File(indexFile).isFile()) {
+      sendResponse(200, indexFile, FileSendingDecision.SEND_FILE);
+      return;
+    }
+    // if index.html does not exists, we send the list of files and folders in that directory
+    File folder;
```
divyanshjr: :P

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+    // If homepage is requested
+    if (requestPath.equals("")) {
+      folder = new File(".");
+    }
+    else {
+      folder = new File(requestPath.substring(0, requestPath.lastIndexOf('/')));
```
divyanshjr: I am trying to remove the last '/' from the requestPath. The end index in the substring method is exclusive.

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+    File[] listOfFiles = folder.listFiles();
+    StringBuffer response = new StringBuffer();
+    response.append("<b>List of files and folders</b> <br>");
```
divyanshjr: Sorry, I overlooked that. Will correct it.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: :+1:

solutions/Assignment-2.3/HttpServer.java
divyanshjr: :+1:

solutions/Assignment-2.3/HttpServer.java
```diff
@@ -0,0 +1,258 @@
...
+    response.append("<ul>");
+    for (int ii = 0; ii < listOfFiles.length; ii++) {
+      if (listOfFiles[ii].isFile()) {
+        String path = listOfFiles[ii].toString().replaceAll(" ","&nbsp;");
```
divyanshjr: I was trying to deal with spaces in  a file name but I'll do URL encoding this time.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: I got it now. I'll change it.

solutions/Assignment-2.3/HttpServer.java
divyanshjr: :+1:

solutions/Assignment-2.3/HttpServer.java
divyanshjr: Alright.

solutions/Assignment-2.3/Assignment2_3.java
```diff
@@ -0,0 +1,71 @@
...
+  */
+
+import httpserver.*;
+import java.io.*;
+import java.net.*;
+import java.util.*;
+import java.util.logging.*;
+import net.sourceforge.argparse4j.*;
+import net.sourceforge.argparse4j.inf.*;
+class Assignment2_3 {
```
kaustubh-karkare: It should be easy to locate the start of the class. Put a new line before this. The style guide certainly has something about this.

solutions/Assignment-2.3/Assignment2_3.java
```diff
@@ -0,0 +1,71 @@
...
+  private static final Logger logger = Logger.getLogger(Assignment2_3.class.getName());
+
+  public static void main(String args[]) throws IOException {
+    int port;
+    String logConfigurationFilePath;
+    ArgumentParser parser = ArgumentParsers.newArgumentParser("Server")
+      .defaultHelp(true)
+      .description("Java Server");
+
```
kaustubh-karkare: Can we take the IP address as input too? If you understand the significance of `0.0.0.0`, you'll understand why.

solutions/Assignment-2.3/Assignment2_3.java
```diff
@@ -0,0 +1,71 @@
...
+            .type(String.class)
+            .dest("configPath")
+            .help("Path containing the Configuration File");
+
+    try {
+      Namespace namespace = parser.parseArgs(args);
+      port = namespace.getInt("port");
+      logConfigurationFilePath = namespace.getString("configPath");
+      HttpServer httpServer = new HttpServerBuilder()
+                                          .setPort(port)
```
kaustubh-karkare: Instead of pushing these lines all the way there, lets stick to 2 more spaces than previous line.

solutions/Assignment-2.3/httpserver/HttpServerBuilder.java
```diff
@@ -0,0 +1,22 @@
...
+  private int port;
+  private String logConfigurationFilePath;
+
+  public HttpServerBuilder setPort(int port) {
+    this.port = port;
+    return this;
+  }
+
+  public HttpServerBuilder setLogConfigurationFilePath(String logConfigurationFilePath) {
+    this.logConfigurationFilePath = logConfigurationFilePath;
```
kaustubh-karkare: Is there a default for this? In case of builder pattern, you need to enforce that things are provided, or assume defaults.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,227 @@
...
+    SEND_FILE,
+    DONT_SEND_FILE
+  }
+
+  // Constructor where thread starts
+  public ClientHandler(Socket socket) throws IOException {
+    this.socket = socket;
+    inFromClient = new BufferedReader(new InputStreamReader(socket.getInputStream()));
+    outToClient = new DataOutputStream(socket.getOutputStream());
+    statusCodes = new HashMap<Integer, String>();
```
kaustubh-karkare: Since it is the same for all clients, how about making it static or something.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,227 @@
...
+          if (new File(requestPath).isFile()) {
+            logger.info("File requested. Preparing to send file - " + requestPath);
+            sendResponse(200, requestPath, FileSendingDecision.SEND_FILE);
+          }
+          else {
+            sendResponse(415, "", FileSendingDecision.DONT_SEND_FILE);
+            logger.info("Unsupported Media Type requested. Denied.");
+          }
+        }
+      }
```
kaustubh-karkare: This is the end of the `if (httpMethod.equals("GET")) {` right? What happens if we didnt get a `GET`?

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,227 @@
...
+    throws IOException {
+
+    FileInputStream fileInputStream = null;
+    final String newLine = "\r\n";
+    String requestPath = "";
+    headers = new LinkedHashMap<String, String>();
+
+    switch (statusCode) {
+      case 200:
+        headers.put("HTTP/1.1 ", statusCodes.get(200));
```
kaustubh-karkare: I'm not sure if this qualifies as a header. And given my next comment, it is probably okay to special case this. Also, "HTTP/1.1" is repeated which is inelegant.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,227 @@
...
+        break;
+      case 415:
+        headers.put("HTTP/1.1 ", statusCodes.get(415));
+        break;
+      case 500:
+        headers.put("HTTP/1.1 ", statusCodes.get(500));
+        break;
+    }
+
+    headers.put("Server: ", "Java HTTPServer");
```
kaustubh-karkare: Dont need ": " in the key. That you can add in the for loop below.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,227 @@
...
+    headers.put("Content-Type: ", "text/html" + newLine);
+
+    if (decision == FileSendingDecision.SEND_FILE) {
+      requestPath = responseString;
+      String contentType = responseString.substring(
+        responseString.lastIndexOf('.'),
+        responseString.length()
+      );
+      if (!requestPath.endsWith(".htm") && !requestPath.endsWith(".html")) {
```
kaustubh-karkare: Why not always send it?

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,227 @@
...
+      fileInputStream = new FileInputStream(requestPath);
+      headers.put(
+        "Content-Length: ",
+        Integer.toString(fileInputStream.available()) +
+        newLine
+      );
+
+    }
+    else {
+      responseString = ClientHandler.HTML_START + responseString + ClientHandler.HTML_END;
```
kaustubh-karkare: Instead of adding the HTML wrapper here, I'd rather do it in `sendDirectoryInformation`

solutions/Assignment-2.3/Assignment2_3.java
divyanshjr: :+1:

solutions/Assignment-2.3/Assignment2_3.java
divyanshjr: Sure, I'll do it..

solutions/Assignment-2.3/Assignment2_3.java
divyanshjr: :+1:

solutions/Assignment-2.3/httpserver/HttpServerBuilder.java
divyanshjr: I'll assume it's default to be null. So if it is not provided, logging will be done on console only.

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: I'll make it static.

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: We'll send Error 405 in that case. I'll do it.

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: I'll correct that. It will be a special case now.

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: :+1:

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: I'll correct it.

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: :+1:

solutions/Assignment-2.3/Assignment2_3.java
```diff
@@ -27,35 +27,59 @@
...
+      .dest("port")
+      .help("Port number on which server will be hosted");
+
+    parser.addArgument("--Backlog")
+      .type(Integer.class)
+      .setDefault(50)
+      .dest("backlog")
+      .help("Requested maximum length of the queue of incoming connections");
+
+    parser.addArgument("--LogConfigurationFilePath")
```
kaustubh-karkare: These should all be lower case (`--log-configuration-file-path`)

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -17,8 +17,8 @@
   private Socket socket = null;
   private BufferedReader inFromClient = null;
   private DataOutputStream outToClient = null;
-  HashMap<Integer, String> statusCodes;
-  LinkedHashMap<String, String> headers;
+  private static HashMap<Integer, String> statusCodes;
```
kaustubh-karkare: You should separate your static properties from the instance properties. Put an empty line between them, so that it is clear.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -150,53 +158,38 @@ public void sendResponse(int statusCode, String responseString, FileSendingDecis
 
     FileInputStream fileInputStream = null;
     final String newLine = "\r\n";
-    String requestPath = "";
+    //String requestPath = "";
```
kaustubh-karkare: Remove.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -150,53 +158,38 @@ public void sendResponse(int statusCode, String responseString, FileSendingDecis
...
-        break;
-    }
-
-    headers.put("Server: ", "Java HTTPServer");
-    headers.put("Content-Type: ", "text/html" + newLine);
+    headers.put("Server", "Java HTTPServer");
 
     if (decision == FileSendingDecision.SEND_FILE) {
-      requestPath = responseString;
+      //requestPath = responseString;
```
kaustubh-karkare: Remove.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -150,53 +158,38 @@ public void sendResponse(int statusCode, String responseString, FileSendingDecis
...
-      headers.put("Content-Length: ", responseString.length() + newLine);
+      headers.put("Content-Type", "text/html" + newLine);
+      headers.put("Content-Length", responseString.length() + newLine);
     }
 
-    headers.put("Connection: ", "close" + newLine + newLine);
+    headers.put("Connection", "close" + newLine + newLine);
+    outToClient.writeBytes("HTTP/1.1 " + statusCodes.get(statusCode));
 
     for (Map.Entry ii: headers.entrySet()) {
```
kaustubh-karkare: Instead of `ii`, which by convention should only contain numerical indices, call this "header", which is a lot more useful.

solutions/Assignment-2.3/httpserver/HttpServer.java
```diff
@@ -11,15 +11,24 @@
   private static final Logger logger = Logger.getLogger(HttpServer.class.getName());
   static String logConfigurationFilePath;
   static ServerSocket serverSocket;
+  static String ipAddress;
   static int port;
+  static int backlog;
 
-  public HttpServer(int port, String logConfigurationFilePath) {
+  public HttpServer(int port, String ipAddress, int backlog, String logConfigurationFilePath) {
```
kaustubh-karkare: Exceeds line length limit.

solutions/Assignment-2.3/httpserver/HttpServer.java
```diff
@@ -11,15 +11,24 @@
...
     HttpServer.port = port;
     HttpServer.logConfigurationFilePath = logConfigurationFilePath;
+    HttpServer.ipAddress = ipAddress;
+    HttpServer.backlog = backlog;
     try {
       initLog();
     } catch (SecurityException | IOException e) {
```
kaustubh-karkare: You're basically initializing logging in the catch-block, which should actually be inside your `initLog` method.

solutions/Assignment-2.3/Assignment2_3.java
```diff
@@ -0,0 +1,95 @@
...
+  *
+  * Make sure that HttpServer.java, HttpServerBuilder.java and ClientHandler.java
+  * are kept in a folder "httpserver" which should be in the current directory of this file.
+  */
+
+import httpserver.*;
+import java.io.*;
+import java.net.*;
+import java.util.*;
+import java.util.logging.*;
```
kaustubh-karkare: While I'm not asking you to make a change here, in general, these `import *` statements dont feel right to me. It is unclear which module your class comes from, and unless we're talking about well known standard modules like `java.io`, I'd rather explicitly mention the classes that you'll be using.

solutions/Assignment-2.3/Assignment2_3.java
```diff
@@ -0,0 +1,95 @@
...
+      .help("Requested maximum length of the queue of incoming connections");
+
+    parser.addArgument("--LogConfigurationFilePath")
+      .type(String.class)
+      .dest("configPath")
+      .setDefault("")
+      .help("Path containing the Configuration File");
+
+
+
```
kaustubh-karkare: 1 empty line is sufficient. It doesn't seem like a big deal, but you want consistency the code your write/read, and unexpected / random "features" like this are distracting.

solutions/Assignment-2.3/httpserver/ClientHandler.java
kaustubh-karkare: What you've done is a start, but dont the following lines still mean that the map is reinitialized on each request? That looks unnecessary.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,220 @@
...
+        logger.info(httpMethod + "requested. Denied");
+      }
+
+    } catch (IOException e) {
+      logger.info(
+        "The Client " +
+        socket.getInetAddress() +
+        ":" +
+        socket.getPort() +
+        " is disconnected"
```
kaustubh-karkare: Nitpicking: put a full stop at the end. Make a habit of doing little things like these, to make your products look more professional :P

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,220 @@
...
+      );
+    }
+  }
+
+  public void sendDirectoryInformation(String requestPath) throws Exception {
+    String indexFile = requestPath + "index.html";
+    // if index.html exists
+    if (new File(indexFile).isFile()) {
```
kaustubh-karkare: Should this happen in the previous method? You're no longer doing what the method name says.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,220 @@
...
+    StringBuffer response = new StringBuffer();
+    response.append("<b>List of files and folders</b>");
+    response.append("<ul>");
+    for (int ii = 0; ii < listOfFiles.length; ii++) {
+      if (listOfFiles[ii].isFile()) {
+        String path = URLEncoder.encode(listOfFiles[ii].toString(), "UTF-8");
+        String fileName = listOfFiles[ii].getName();
+        response.append(
+          "<li>File - " +
+          "<a href=" +
```
kaustubh-karkare: I think you want: `"<a href='/" +`.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,220 @@
...
+    responseString = ClientHandler.HTML_START + responseString + ClientHandler.HTML_END;
+    sendResponse(200, response.toString(), FileSendingDecision.DONT_SEND_FILE);
+  }
+
+  public void sendResponse(int statusCode, String responseString, FileSendingDecision decision)
+    throws IOException {
+
+    FileInputStream fileInputStream = null;
+    final String newLine = "\r\n";
+    //String requestPath = "";
```
kaustubh-karkare: Remove all dead code.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,220 @@
...
+    headers.put("Server", "Java HTTPServer");
+
+    if (decision == FileSendingDecision.SEND_FILE) {
+      //requestPath = responseString;
+      String contentType = responseString.substring(
+        responseString.lastIndexOf('.'),
+        responseString.length()
+      );
+
+      headers.put("Content-Type", contentType + newLine);
```
kaustubh-karkare: You dont need to add newLine to every value. Just do it once in the loop. Try and avoid unnecessary duplication.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,220 @@
...
+      );
+
+    }
+    else {
+      headers.put("Content-Type", "text/html" + newLine);
+      headers.put("Content-Length", responseString.length() + newLine);
+    }
+
+    headers.put("Connection", "close" + newLine + newLine);
+    outToClient.writeBytes("HTTP/1.1 " + statusCodes.get(statusCode));
```
kaustubh-karkare: Shouldn't there be a newLine after this?

solutions/Assignment-2.3/httpserver/HttpServer.java
```diff
@@ -0,0 +1,71 @@
...
+
+  // Method to create log file
+  public static void initLog() throws SecurityException, IOException {
+    // Reading Configuration file which contains predefined instructions regarding logging
+    logManager.readConfiguration(new FileInputStream(logConfigurationFilePath));
+  }
+  // The run() method of HttpServer
+  @Override
+  public void run() {
+
```
kaustubh-karkare: Unnecessary line.

solutions/Assignment-2.3/httpserver/HttpServer.java
```diff
@@ -0,0 +1,71 @@
...
+    try {
+      if (port < 0) {
+        throw new NegativeNumberException();
```
kaustubh-karkare: Isnt this part of initialization?

solutions/Assignment-2.3/httpserver/HttpServerBuilder.java
```diff
@@ -0,0 +1,34 @@
...
+    return this;
+  }
+
+  public HttpServer getHttpServer() {
+    return new HttpServer(port, ipAddress, backlog, logConfigurationFilePath);
```
kaustubh-karkare: What happens if `getHttpServer` is called without providing the port?

solutions/Assignment-2.3/Assignment2_3.java
divyanshjr: I understand..

solutions/Assignment-2.3/Assignment2_3.java
divyanshjr: Alright..

solutions/Assignment-2.3/Assignment2_3.java
divyanshjr: Yeah.. My ignorance.. :P

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: Yup, I'll try to to correct that..

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: I'll try not to overlook these things.. :P

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: Yup, I'll do it..

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: Sure..

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: :+1:

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: Yes, I'll add that.

solutions/Assignment-2.3/httpserver/HttpServerBuilder.java
divyanshjr: The parser sends the default value i.e. 8080

solutions/Assignment-2.3/httpserver/HttpServer.java
divyanshjr: Yup.. will do this in the constructor itself..

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: I kept the '/' in the next line. Will move it here..

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: Sure..

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: :+1:

solutions/Assignment-2.3/httpserver/HttpServer.java
divyanshjr: I'll try to correct that. Actually here I am trying to read the properties file in the **intit() method** which contains the logging details i.e. where to log, how to log etc and if that is not provided or couldn't be read then by default logging occurs only on console and that is being notified in the catch block by logging that warning message.

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: :+1:

solutions/Assignment-2.3/httpserver/ClientHandler.java
kaustubh-karkare: Actually, I was more concerned about the fact that you should add quotes around the path.

solutions/Assignment-2.3/Assignment2_3.java
```diff
@@ -0,0 +1,91 @@
...
+        .setBacklog(backlog)
+        .setLogConfigurationFilePath(logConfigurationFilePath)
+        .getHttpServer();
+
+      httpServer.start();
+      try {
+        Thread.sleep(60);
+        httpServer.join();
+      } catch (InterruptedException e) {}
+    } catch (ArgumentParserException e) {
```
kaustubh-karkare: Why does line 74 to 85 need to be in this block?
I'd expect that you return from the catch-block to skip those parts.
Putting too much code in a try-catch introduces ambiguity about the source of the error, which I'm trying to avoid.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,223 @@
...
+    "<body>";
+  static final String HTML_END = "</body>" + "</html>";
+  private static HashMap<Integer, String> statusCodes;
+  // Static block to initialize the HashMap.
+  static {
+    statusCodes = new HashMap<Integer, String>();
+    statusCodes.put(200, "200 OK");
+    statusCodes.put(404, "404 Not Found");
+    statusCodes.put(500, "500 Internal Server Error");
+    statusCodes.put(415, "415 Unsupported Media Type");
```
kaustubh-karkare: Do you really need the number in the string again?

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,223 @@
...
+public class ClientHandler extends Thread {
+  private static final Logger logger = Logger.getLogger(ClientHandler.class.getName());
+  static final String HTML_START =
+    "<html>" +
+    "<title>HTTP Server in java</title>" +
+    "<body>";
+  static final String HTML_END = "</body>" + "</html>";
+  private static HashMap<Integer, String> statusCodes;
+  // Static block to initialize the HashMap.
+  static {
```
kaustubh-karkare: Ah, interesting. Didn't know this was possible.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,223 @@
...
+        // If request is neither for a folder or file, then it means requested resource is not found
+        else {
+          sendResponse(404, "", FileSendingDecision.DONT_SEND_FILE);
+          logger.info(requestPath + " - Requested resource not found.");
+        }
+      }
+
+      else {
+        sendResponse(405, "", FileSendingDecision.DONT_SEND_FILE);
+        logger.info(httpMethod + "requested. Denied.");
```
kaustubh-karkare: Looks like this will print "POSTrequest. Denied." Need a space, apparently.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,223 @@
...
+    StringBuffer response = new StringBuffer();
+    response.append("<b>List of files and folders</b>");
+    response.append("<ul>");
+    for (int ii = 0; ii < listOfFiles.length; ii++) {
+      if (listOfFiles[ii].isFile()) {
+        String path = URLEncoder.encode(listOfFiles[ii].toString(), "UTF-8");
+        String fileName = listOfFiles[ii].getName();
+        response.append(
+          "<li>File - " +
+          "<a href=/" +
```
kaustubh-karkare: My previous comment was actually about putting the path inside quotes.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,223 @@
...
+      outToClient.writeBytes(header.getValue().toString());
+      outToClient.writeBytes(newLine);
+    }
+
+    outToClient.writeBytes(newLine);
+    logger.info("Headers sent successfully");
+
+    if (decision == FileSendingDecision.SEND_FILE) {
+      sendFile(fileInputStream, outToClient);
+    }
```
kaustubh-karkare: For consistency with the else block, move the logger call from inside the function to here.

solutions/Assignment-2.3/httpserver/HttpServerBuilder.java
kaustubh-karkare: Ah, sorry, I didn't take into consideration the fact that everything had a default value. This is fine then.

solutions/Assignment-2.3/httpserver/HttpServer.java
```diff
@@ -0,0 +1,70 @@
...
+  public static void initLog() {
+    logger.setLevel(Level.ALL);
+    logger.info("Initializing - trying to load configuration file...");
+    Properties preferences = new Properties();
+    try {
+      FileInputStream configFile = new FileInputStream(logConfigurationFilePath);
+      preferences.load(configFile);
+      logManager.readConfiguration(configFile);
+    } catch (IOException e) {
+      if (logConfigurationFilePath.isEmpty()) {
```
kaustubh-karkare: This can be outside the try-catch, right?

solutions/Assignment-2.3/Assignment2_3.java
divyanshjr: I'll correct it..

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: Not required.. I'll remove that..

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: Will correct it..

solutions/Assignment-2.3/httpserver/HttpServer.java
divyanshjr: Yup.. will do..

solutions/Assignment-2.3/httpserver/ClientHandler.java
divyanshjr: :+1:

solutions/Assignment-2.3/Assignment2_3.java
```diff
@@ -0,0 +1,97 @@
+/**
+  * JAVA code to implement a web server that can handle basic GET requests.
+  * This is the driver code which uses the httpserver API.
+  *
+  * A third party package - argparse4j (Command Line Parser) has been used so it must be
+  * installed inorder to run the webserver.
+  *
+  * Four command line arguments(optional) are needed during the execution in the recommended format as shown -
+  *   java Assignment2_3 --P PORT_NUMBER --IP IP_ADDRESS --L LOG_CONFIGURATION_FILE_PATH --B BACKLOG
```
kaustubh-karkare: These arguments are lowercase now.

solutions/Assignment-2.3/Assignment2_3.java
```diff
@@ -0,0 +1,97 @@
...
+      .help("Path containing the Configuration File");
+
+    try {
+      namespace = parser.parseArgs(args);
+    } catch (ArgumentParserException e) {
+      isArgumentParserException = true;
+      logger.warning(e.getMessage());
+    }
+    
+    if (!isArgumentParserException) {
```
kaustubh-karkare: Instead of this condition, just return after line 74.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,223 @@
...
+    statusCodes.put(200, "OK");
+    statusCodes.put(404, "Not Found");
+    statusCodes.put(500, "Internal Server Error");
+    statusCodes.put(415, "Unsupported Media Type");
+  }
+
+  private Socket socket = null;
+  private BufferedReader inFromClient = null;
+  private DataOutputStream outToClient = null;
+  private LinkedHashMap<String, String> headers;
```
kaustubh-karkare: This doesn't need to be on the class, since you're only using it inside 1 method.

solutions/Assignment-2.3/httpserver/ClientHandler.java
```diff
@@ -0,0 +1,223 @@
...
+
+    FileInputStream fileInputStream = null;
+    final String newLine = "\r\n";
+    headers = new LinkedHashMap<String, String>();
+
+    headers.put("Server", "Java HTTPServer");
+
+    if (decision == FileSendingDecision.SEND_FILE) {
+      String contentType = responseString.substring(
+        responseString.lastIndexOf('.'),
```
kaustubh-karkare: The "responseString" contains something that is not a response string.

divyanshjr:
Commited the final changes as asked. Closing the PR now!

PR #29, opened 2017-03-30
Attempt at web-server-and-client GA1 [Aakarshit]

aksh1618:


solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env python3
+"""Download file from a link using socket connection.
+
+Takes url as required argument in command line and destination path as
+optional argument, shows progress durin download and succes message on
```
kaustubh-karkare: typos: during, success

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,177 @@
...
+import socket  # For soket
+import sys  # For argv, exit
+import os  # For handling directory
+import threading  # For Thread
+import time  # For time
+import shutil  # For get_terminal_size
+import collections  # For named tuple
+
+
+class _HttpRequestHandler:
```
kaustubh-karkare: This shouldn't be private. The point of making these reusable is that they can be invoked from other modules.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,177 @@
...
+    # Handles HTTP requests, from url to request to response
+
+    def __init__(self, url):
+        # Set url and buffer size
+        self.url = url
+        self.buff_size = 2048
+
+    def _handle_url(self):
```
kaustubh-karkare: `_parse_url` is more appropriate. "handle" can mean so many things that it ultimately means nothing.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,177 @@
...
+        url = self.url
+        try:
+            url = url.split('//', 1)[1]
+        except IndexError:
+            pass
+
+        separation_point = url.find('/')
+        self.host = url[:separation_point]
+        self.path = url[separation_point:]
+        self.file_name = url[url.rfind('/'):][1:]
```
kaustubh-karkare: If you `+1` the result of `rfind`, you wont need `[1:]`, which is costlier.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,177 @@
...
+
+        if self.file_name is '':
+            print('Error: Nothing to download.')
```
kaustubh-karkare: As part of making your code reusable, you can no longer rely on the assumption that stdout is available. What if this was a background process that was running while the main program is asking a bunch of questions to the user (through stdout) about what to do once the file is downloaded? You dont want to mix outputs.

And so,
- never use `sys.exit` unless you are absolutely sure that is what you want,
- throw exceptions incase of errors to allow the users of this module to decide what to do,
- and use the `logging` module for any output.

And then, in your `__main__` section below, you can connect the logging module output to stdout, and not catch exceptions to let them be surfaced to the user.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,177 @@
...
+            sys.exit(0)
+
+    def _create_request(self):
+        # Create byte encoded request
+        self.request = "GET %s HTTP/1.1\nHost: %s\n\n" % (self.path, self.host)
+        self.request = self.request.encode()
+
+    def _fetch_data(self):
+        # Fetch data using sockets
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as dlsocket:
```
kaustubh-karkare: "dlsocket"? I dont know what "dl" means. Well, I do, but imagine if a line involving "dlsocket" was surfaces to the user as part of a stack trace without additional context. You're making things harder.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,177 @@
...
+        self.file_name = url[url.rfind('/'):][1:]
+
+        if self.file_name is '':
+            print('Error: Nothing to download.')
+            sys.exit(0)
+
+    def _create_request(self):
+        # Create byte encoded request
+        self.request = "GET %s HTTP/1.1\nHost: %s\n\n" % (self.path, self.host)
+        self.request = self.request.encode()
```
kaustubh-karkare: There's no reason why this needs to be saved as a property variable. You should minimize the number of instance properties anyway, compute on demand for simple things like this. If the computation is costly, then it might be more acceptable, but I'd prefer to use memoization decorators in that case, instead of this. Either way, this function is simple enough that I would just do this in `_fetch_data`.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def _fetch_data(self):
+        # Fetch data using sockets
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as dlsocket:
+
+            # Send request and receive message+data
+            dlsocket.connect((self.host, 80))
```
kaustubh-karkare: Lets not assume port 80. Webservers can run on 8080 too. Let that be an input too, which can then default to 80.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,177 @@
...
+            dlsocket.send(self.request)
+            self.start_time = time.time()
+            response = dlsocket.recv(self.buff_size)
+
+            # Extract value of content_length attribute from response message
+            content_length = response[response.find(b'Content-Length') + 15:]
```
kaustubh-karkare: According to the `socket.recv` docs, "The maximum amount of data to be received at once is specified by bufsize." Which means that you could get less too. And so you're cant assume that the response contains this header. It might be at the boundary, so you only get "Content-Le" in this block, and then be forced to call recv again to get "ngth: blah".

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,177 @@
...
+            total = self.length
+            done = self.received_data_length
+            scale = shutil.get_terminal_size()[0] - 30
+            self.progress = int((done * 100) / total)
+            scaled_progress = int(self.progress * scale / 100)
+            time.sleep(0.001)
+            sys.stdout.write("\b" * (scale + 5))
+            progress_str = '[{}{}]{}%'.format('#' * scaled_progress,
+                                              ' ' * (scale - scaled_progress),
+                                              self.progress)
```
kaustubh-karkare: You shouldn't be doing this manually. Create a separate class that you can use in a `with` statement to do this. Also, based on the comment above about not assume the availability of stdout, you should probably have a `suppress_output` option to skip this entirely.

aksh1618:

solutions/aakarshit-uppal/week-2/dl_exceptions.py
```diff
@@ -0,0 +1,19 @@
...
+
+class NoFileException(Exception):
+    def __init__(self):
+        error_message = 'Error: Nothing to download.'
+        super().__init__(error_message)
+
+
+class ConnectionException(Exception):
+    def __init__(self):
+        error_message = 'Error: Unable to connect to download link'
```
kaustubh-karkare: The "download link" is just a string. You mean "unable to connect to server"

solutions/aakarshit-uppal/week-2/dl_exceptions.py
```diff
@@ -0,0 +1,19 @@
...
+        super().__init__(error_message)
+
+
+class InvalidLinkException(Exception):
+    def __init__(self):
+        error_message = 'Error: Invalid link, Nothing to download'
```
kaustubh-karkare: inconsistent capitalization :P

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def __init__(self, url, port=80, parts=4):
+        """Set values of globals."""
+        self.url = url
+        self.buff_size = 2048
+        self.port = port
+        self.parts = parts
+        self.fetched_data = {}
+
+    def parse_url(self):
```
kaustubh-karkare: This looks painful. Use a regular expression to parse. Solves the problem in 1 line.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+
+        if self.file_name is '':
+            logger.error('Nothing to download.')
+            raise NoFileException
+
+    def get_length(self):
+        """Send HEAD request to extract content-length."""
+        with socket.socket(socket.AF_INET,
+                           socket.SOCK_STREAM
+                           ) as download_socket:
```
kaustubh-karkare: This formatting is better:
```
with socket.socket(
  socket.AF_INET,
  socket.SOCK_STREAM
) as download_socket:

File: solutions/aakarshit-uppal/week-2/p22.py [hunk 15]
```diff
@@ -0,0 +1,228 @@
...
+            request = "HEAD {} HTTP/1.1\nHost: {}\n\n".format(self.path,
+                                                              self.host)
+            request = request.encode()
+
+            # Send request and receive header
+            download_socket.connect((self.host, self.port))
+            download_socket.send(request)
+            self.start_time = time.time()
+            response = b''
+            while(b'\r\n\r\n' not in response):
```
kaustubh-karkare: In general, I will also complain about strings like these which randomly appear in code.
Put this in a constant at the top of the file, so that you dont have to keep repeating it.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+                response += download_socket.recv(self.buff_size)
+
+            # Extract value of content_length attribute from response message
+            content_length = response[response.find(b'Content-Length') + 15:]
```
kaustubh-karkare: String manipulation isn't pretty. How about you convert the response (except the 1st line) into a `Map<header-name, value>`, so that you can do `content_length = headers['Content-Length']`.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+            content_length = content_length[:content_length.find(b'\r')]
+            try:
+                self.length = int(content_length.decode())
+            except ValueError:
+                logger.error('Error: No internet/Invalid download link')
+                raise ConnectionException
+            if self.length == 0:
+                logger.error('Error: Invalid link, Nothing to download')
```
kaustubh-karkare: This isn't really an error. It is possible for empty file to exist.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+            request = request.encode()
+
+            # Send request and receive header+data
+            download_socket.connect((self.host, self.port))
+            download_socket.send(request)
+            # self.start_time = time.time()
+            logger.debug('Beginning download of part %d' % part)
+            response = b''
+            while(b'\r\n\r\n' not in response):
+                response += download_socket.recv(self.buff_size)
```
kaustubh-karkare: This whole section looks duplicated from the previous method. Lets reuse code.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+            # Fetch data in chunks
+            fetched_data = b''
+            while True:
+                remaining_bytes = bytes_to_fetch - len(fetched_data)
+                data_chunk = download_socket.recv(remaining_bytes)
+                fetched_data += data_chunk
+                if not data_chunk:
+                    break
+            self.finish_time = time.time()
+            self.fetched_data[part] = initial_data + fetched_data
```
kaustubh-karkare: What if the file so large that your system's RAM is not enough to contain everything?

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+            logger.debug('Finished download of part %d' % part)
+
+    def download_successful(self):
```
kaustubh-karkare: This name does not describe what this function does.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+        for _, data in self.fetched_data.items():
+            received_data_length += len(data)
+        return received_data_length == self.length
+
+    def _join_parts(self):
+        # Join fetched data
+        data = b''
+        for part in range(self.parts):
+            data += self.fetched_data[part]
+        self.data = data
```
kaustubh-karkare: Like before, dont assume it'll fit in memory.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+        logger.debug('Finished joining')
+
+    def get_stats(self):
+        """Return named tuple containing download stats."""
+        kB_downloaded = (self.length / 1024)
+        time_taken = (self.finish_time - self.start_time)
+        download_speed = (kB_downloaded / time_taken)
+        new_stats = collections.namedtuple('new_stats', ['kB_downloaded',
```
kaustubh-karkare: There's no reason you create a new namedtuple each time this method is called. You can create it once outside.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+                                                         'time_taken',
+                                                         'download_speed'])
+        stats = new_stats(kB_downloaded, time_taken, download_speed)
+        return stats
+
+    def finish_up(self):
```
kaustubh-karkare: This name does not describe what this function does.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+                                                         stat_string))
+
+
+class Downloader:
+    """Download file from a url."""
+
+    def __init__(self, url=sys.argv[1], dl_path='.', port=80, parts=4):
+        """Make request object and set path."""
+        self.request = HttpRequestHandler(url, port=port, parts=parts)
+        if len(sys.argv) > 2:
```
kaustubh-karkare: "sys.argv" really? :P

Use argparse or something man. This is way too fragile.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+            dl_path = sys.argv[2]
+        self.dl_path = dl_path
```
kaustubh-karkare: This ... is not a good name. Avoid abbreviations.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+        self.parts = parts
+
+    def download(self):
+        """Download file data."""
+        request = self.request
+        request.parse_url()
+        request.get_length()
```
kaustubh-karkare: A method that starts with a "get" should return something. If you're not returning anything, then it's not getting anything for you, and so the name is inaccurate.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+
+        logger.debug('Beginning download from {}'.format(request.url))
+        request.start_time = time.time()
+        downloads = []
+        for part in range(self.parts):
```
kaustubh-karkare: Use a ThreadPoolExecutor. Will simplify things a lot.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+
+        if(request.download_successful()):
+            logger.debug('Joining parts...')
+            request.finish_up()
+        self._create_file()
+
+    def _create_file(self):
+        # Create file from downloaded data.
+        data_to_write = self.request.data
+        file_with_path = '{}/{}'.format(self.dl_path, self.request.file_name)
```
kaustubh-karkare: Cant assume that the file name must match the URL. What if user wants something else?

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+import os  # For handling directory
+import threading  # For Thread
+import time  # For time
+# import shutil  # For get_terminal_size
+import collections  # For named tuple
+import logging  # For logging
+from dl_exceptions import *  # For exceptions
+
+# Logging specifications
+logger = logging.getLogger(__name__)
```
kaustubh-karkare: A single logger across multiple threads means that your output can be all jumbled up. Not a good idea.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+logger.setLevel(logging.DEBUG)
+file_handler = logging.FileHandler('download.log')
```
kaustubh-karkare: This should be configurable.

solutions/aakarshit-uppal/week-2/p22.py
```diff
@@ -0,0 +1,228 @@
...
+        request.start_time = time.time()
+        downloads = []
+        for part in range(self.parts):
+            downloads.append(threading.Thread(target=request.fetch_data,
+                                              args=(part,)))
+            downloads[part].start()
+        while(threading.activeCount() > 1):
+            time.sleep(0.01)
+
+        if(request.download_successful()):
```
kaustubh-karkare: Are these parenthesis necessary?

kaustubh-karkare:
Inactive for too long.

PR #32, opened 2017-03-31
assignment-2.2.py

humanoid2795:
assignment-2.2.py


solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+    parser.add_argument(
+        '--threads',
+        type=int,
+        required=False,
+        default=4,
+        help='Number of threads to download'
+    )
+    parser.add_argument(
+        '--output-path',
+        '-op',
```
kaustubh-karkare: Short names generally dont involve more than 1 letter.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
+import socket
+import threading
+import math
+import tempfile
+import argparse
+import logging
+
+BUFF_SIZE = 1024
+
```
kaustubh-karkare: You should make you code more reusable. Putting everything in one file doesn't fit with that ideology.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+    )
+    parser.add_argument(
+        '--output-path',
+        '-op',
+        type=str,
+        required=True,
+        help='Downloaded file name'
+    )
+    parsed_argument = parser.parse_args()
+    url_parser = UrlParser(parsed_argument.url)
```
kaustubh-karkare: Feels like this can be a method that returns a `namedtuple`, instead of a class. Otherwise you seem to be doing repeating computation each time `.host()` is called. Alternatively, use regular expressions, which will be useful for problem 3.1 too.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+    threads = parsed_argument.threads
+    output_path = parsed_argument.output_path
+    try:
+        output_file = open(output_path, 'w')
+    except:
+        logging.error('Output file could not be established')
+    DownloadFile(
+        host,
+        port,
+        path,
```
kaustubh-karkare: It shouldn't be your responsibility to do this. A good API would accept a URL and parse it inside the class. Remember, you need to make these easy to use.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+        self.client_socket.send('HEAD ' + self.request)
+        self.received_headers = ''
+        while True:
+            self.data_received = self.client_socket.recv(BUFF_SIZE)
+            if not self.data_received:
+                break
+            self.received_headers += self.data_received
+        self.received_headers = (self.received_headers).splitlines()
+        self.headers = {}
+        self.i = 1
```
kaustubh-karkare: Dude.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+        self.path = path
+        self.threads = threads
+        self.output_file = output_file
+        self.request = self.path + 'HTTP/1.1\n'
+        self.request += 'Host: ' + self.host + '\n'
+        self.request += 'Range: bytes=0-' + '\n\n'
+        threading.Thread.__init__(self)
+
+    def run(self):
+        self.get_response_headers()
```
kaustubh-karkare: A method that starts with "get" should return stuff. But based on this usage, you're not saving any return value, and that suggests that it is saved as a class properly. That is non-intuitive. Always save the bare minimum on the class object, and then compute all derived information on demand.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+        self.download_file()
+
+    def get_response_headers(self):
+        self.client_socket = socket.socket()
+        try:
+            self.client_socket.connect((self.host, self.port))
```
kaustubh-karkare: So you're connecting to a server in a method called "get_response_headers" ?

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+        except:
+            logging.error('Could not establish connection')
```
kaustubh-karkare: You should probably reraise the exception. Ignoring it is a bad idea, since `recv` below will then error out.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+        self.client_socket.send('HEAD ' + self.request)
```
kaustubh-karkare: The fact that HEAD and GET requests are so similar suggests that you should be sharing code.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+            self.thread1 = DownloadRange(
+                self.host,
+                self.port,
+                self.path,
+                self.start_position,
+                self.end_position,
+                self.temp_file,
+                len(self.headers) + 1
+            )
+            self.thread1.start()
```
kaustubh-karkare: Why is this called "thread1" ?

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+        self.download_range()
+
+    def download_range(self):
+        self.client_socket = socket.socket()
+        try:
+            self.client_socket.connect((self.host, self.port))
+        except:
+            logging.error('Could not establish connection')
+        self.client_socket.send(self.request)
+        self.receiving_data_file = tempfile.NamedTemporaryFile()
```
kaustubh-karkare: Why do you need 2 separate temp files?

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+        while (
+            self.i < len(self.received_headers) and
+            self.received_headers[self.i]
+        ):
+            self.curr_header = self.received_headers[self.i]
+            self.split_headers = (self.curr_header).split(':')
+            self.headers[self.split_headers[0]] = self.split_headers[1]
+            self.i = self.i + 1
+
+    def download_file(self):
```
kaustubh-karkare: This function is too long. Split it up into more manageable parts.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+        start_position,
+        end_position,
+        range_download_file,
+        header_count
+    ):
+        self.host = host
+        self.port = port
+        self.path = path
+        self.start_position = start_position
+        self.end_position = end_position
```
kaustubh-karkare: Why do you even save these?

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,279 @@
...
+                )
+                self.thread_instances.append(
+                    DownloadRange(
+                        self.host,
+                        self.port,
+                        self.path,
+                        self.start_range,
+                        self.end_range,
+                        self.range_download_file,
+                        len(self.headers) + 1
```
kaustubh-karkare: This seems like an unnecessarily complex API.

solutions/assignment-2.2.py
humanoid2795: In order to remove the header content from the response by the host, i am temporarily copying it in one file and then after removing the headers, i am copying the data part in another file that will be merged later.

solutions/assignment-2.2.py
humanoid2795: The self.start_position, self.end_position currently store the range of bytes from the data that is to be downloaded by the current invoked thread.

solutions/assignment-2.2.py
humanoid2795: Okay, i'll correct it.

solutions/assignment-2.2.py
humanoid2795: I should not have done that, i'll rectify it.

solutions/assignment-2.2.py
humanoid2795: Agreed, i'll try to implement that.

solutions/assignment-2.2.py
humanoid2795: Okay.

solutions/assignment-2.2.py
humanoid2795: Bad declaration. Will  change it.

solutions/assignment-2.2.py
humanoid2795: Okay.

solutions/assignment-2.2.py
humanoid2795: Okay, i'll change it to be parsed from within the class(DownloaFile)

solutions/assignment-2.2.py
humanoid2795: Okay.

solutions/assignment-2.2.py
kaustubh-karkare: But since you already save the "range_to_get", these should not be necessary, right? You're not using them anyway. Remember what I said: save as little information as possible on the class. In this case, you can either save the limit, or the range, but not both, since they're redundant.

solutions/assignment-2.2.py
kaustubh-karkare: That is inelegant. Why not just avoid writing the headers? Since you know when the headers end (consecutive EOLs), it shouldn't be a problem.

solutions/urlparser.py
```diff
@@ -0,0 +1,24 @@
+import re
+import collections
+
+
+def parse_url(url):
+    pattern = (r'(https?)?(\:\/\/)?([a-zA-Z0-9-.]+)?\:?(\d+)?(.*)')
```
kaustubh-karkare: Ctrl+F `(?:...)` on https://docs.python.org/2/library/re.html

solutions/urlparser.py
```diff
@@ -0,0 +1,24 @@
...
+    url_group = re.match(pattern, url)
+    # url_group.group(1) contains protocol
+    # url_group.group(3) contains host
+    # url_group.group(4) contains port
+    # url_group.group(5) contains path
+    UrlParser = collections.namedtuple('UrlParser', 'host port path')
```
kaustubh-karkare: This should be outside the function. No point it repeating the logic each time the method is called.

solutions/urlparser.py
```diff
@@ -0,0 +1,24 @@
...
+    # Default port is 80(Assuming https)
+    port = 80
+    if url_group.group(4) is not None:
+        port = int(url_group.group(4))
+    parsed_url = (
```
kaustubh-karkare: Just return directly.

solutions/responseheaders.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def get_headers(client_socket, parsed_url):
+    client_request = (
+        'HEAD ' +
+        parsed_url.path +
+        ' HTTP/1.1\r\n' +
+        'Host: ' +
+        parsed_url.host +
+        '\r\n' +
```
kaustubh-karkare: Given the number of times you use `\r\n`, you should put it in a constant.

solutions/responseheaders.py
```diff
@@ -0,0 +1,39 @@
+import logging
+
+BLOCK_SIZE = 1024
+
+
+def get_headers(client_socket, parsed_url):
```
kaustubh-karkare: This isn't a particularly reusable method.

Also, it isn't named properly. `get_headers` sounds like you could extract them from the client-socket, but the fact that you're making a request from here isn't obvious.

solutions/responseheaders.py
```diff
@@ -0,0 +1,39 @@
...
+        parsed_url.host +
+        '\r\n' +
+        'Connection: close' +
+        '\r\n\r\n'
+    )
+    client_socket.send(client_request)
+    received_data = ''
+    while True:
+        data = client_socket.recv(BLOCK_SIZE)
+        if not data:
```
kaustubh-karkare: You're still loading the entire file into memory. We only care about the header here. You can stop early, and then resume later while writing to tempfile.

solutions/responseheaders.py
```diff
@@ -0,0 +1,39 @@
...
+            break
+        received_data += data
+    status, headers_received = received_data.split('\r\n', 1)
+    headers_received = headers_received.splitlines()
+    headers = {}
```
kaustubh-karkare: `response_headers`. Also, you should have a similar map of `request_headers` used to generate the request string.

solutions/responseheaders.py
```diff
@@ -0,0 +1,39 @@
...
+    i = 0
+    while i < len(headers_received) - 1:
```
kaustubh-karkare: s/i/ii. But to be honest, you dont need this variable. You can do `for ?? in <list>`

solutions/responseheaders.py
```diff
@@ -0,0 +1,39 @@
...
+        current_header = headers_received[i]
+        try:
+            parameter, value = current_header.split(':', 1)
```
kaustubh-karkare: Checkout `str::partition`.

solutions/responseheaders.py
```diff
@@ -0,0 +1,39 @@
...
+            headers[parameter] = value[1:]
+        except:
+            logging.warning(
```
kaustubh-karkare: It's fine to throw an exception.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,200 @@
...
+import tempfile
+import argparse
+import logging
+import responseheaders
+import urlparser
+
+BLOCK_SIZE = 1024
+
+
+def download(client_socket, request_message, download_file):
```
kaustubh-karkare: As I write this comment, I've not read your implementation. But the `request_message` argument is confusing me, which means that it isn't intuitive enough.
Update: After actually reading this, it feels like `download` might be the wrong name for this internal method. The general meaning of the term "download" is too specific for this implementation, which could do a lot more than download things.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,200 @@
...
+        data = client_socket.recv(BLOCK_SIZE)
+        if not data:
+            break
+        try:
+            headers, data = data.split('\r\n\r\n')
+            download_file.write(data)
+        except:
+            download_file.write(data)
+    download_file.flush()
+    client_socket.close()
```
kaustubh-karkare: It isn't obvious from the name that you're going to do this. And so you either need a better name, or you shouldn't do this here.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,200 @@
...
+    client_socket = socket.socket()
+    try:
+        client_socket.connect((host, port))
+    except:
+        logging.error('Client socket connection could not be established')
+        raise
+    return client_socket
+
+
+def temporary_file():
```
kaustubh-karkare: Why are these separate methods?

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,200 @@
...
+class DownloadFile(threading.Thread):
+
+    def __init__(self, url, threads, output_path):
+        self.url = url
+        self.threads = threads
+        self.output_path = output_path
+        threading.Thread.__init__(self)
+
+    def run(self):
+        self.parsing_url()
```
kaustubh-karkare: `parse_url`. `parsing` sounds like a status update, not a callable method.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,200 @@
...
+        self.response_headers()
+        self.pre_download_checks()
+        self.download_files()
+        self.merge_downloaded_files()
+
+    def parsing_url(self):
```
kaustubh-karkare: There is a balance between creating new functions (which add a cost of context switching), and adding logic to the original callsite (too much of that increases complexity). I'd say that 1-liner functions dont deserve to be functions, and you might as well do the thing right there, instead of forcing me to jump around.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def __init__(self, url, threads, output_path):
+        self.url = url
+        self.threads = threads
+        self.output_path = output_path
+        threading.Thread.__init__(self)
+
+    def run(self):
+        self.parsing_url()
+        self.response_headers()
```
kaustubh-karkare: response headers what? You need a better name.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,200 @@
...
+                self.threads = 1
+        except:
+            self.threads = 1
+        self.each_file_bytes = int(
+            math.ceil(
+                int(self.headers['Content-Length']) /
+                self.threads
+            )
+        )
+        self.download_request_message = (
```
kaustubh-karkare: So a method which is supposed to "check" something is creating a request? Isn't that lying to the developer?
Also, this shouldn't be saved on the class. You can use return values and arguments here.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+        self.client_socket.close()
+
+    def pre_download_checks(self):
+        try:
+            if (self.headers['Content-Length'] == 0):
+                logging.warning('No data to download')
+        except:
+            logging.error('Download size not available')
+            raise
```
kaustubh-karkare: You dont need to `raise` this. You can just proceed with 1 thread.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,200 @@
...
+            'Range: bytes='
+        )
+
+    def download_files(self):
+        self.start_range = 0
+        self.end_range = self.each_file_bytes
+        self.downloaded_file_names = []
+        self.download_threads = []
+        while self.threads > 0:
+            if (self.start_range > int(self.headers['Content-Length']) - 1):
```
kaustubh-karkare: Isn't it inconvenient to use `int(self.headers['Content-Length'])` thrice? Put it in a variable to avoid the possible spelling mistake in the string.
In general, assume that strings will contain spelling mistakes, and so you should never repeat a string more than once.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,200 @@
...
+        logging.error('Temporary file could not be created')
+        raise
+
+
+class DownloadFile(threading.Thread):
+
+    def __init__(self, url, threads, output_path):
+        self.url = url
+        self.threads = threads
+        self.output_path = output_path
```
kaustubh-karkare: These 3 things are the only ones you need to save on your class. Nothing else.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,200 @@
...
+            self.threads -= 1
+
+        self.i = 0
+        while self.i < len(self.download_threads):
+            self.download_threads[self.i].join()
+            self.i += 1
+
+    def merge_downloaded_files(self):
+        try:
+            self.output_file = open(self.output_path, 'w')
```
kaustubh-karkare: Check out the `with` syntax, but using it is optional.
```
with open(self.output_path, 'w'):
  ...

File: solutions/assignment-2.2.py [hunk 36]
```diff
@@ -0,0 +1,200 @@
...
+                self.downloaded_file_names[self.i]
+            )
+            self.downloaded_file = open(self.downloaded_file_name, 'r')
+            self.downloaded_file.seek(0)
+            self.output_file.write(self.downloaded_file.read())
+            self.downloaded_file.close()
+            self.i += 1
+
+
+def main():
```
kaustubh-karkare: How about putting this CLI related stuff in a separate file? It is conceptually distinct from the actual downloading logic.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,35 @@
...
+        '-o',
+        type=str,
+        required=True,
+        help='Downloaded file name'
+    )
+    parsed_argument = parser.parse_args()
+    downloadfile.DownloadFile(
+        url=parsed_argument.url,
+        threads=parsed_argument.threads,
+        output_path=parsed_argument.output_path
```
kaustubh-karkare: Put commas after the last argument too.

solutions/downloadfile.py
```diff
@@ -0,0 +1,197 @@
...
+BLOCK_SIZE = 1024
+
+
+def receive_file_content_from_host(client_socket, download_file):
+    while True:
+        data = client_socket.recv(BLOCK_SIZE)
+        if not data:
+            break
+        try:
+            headers, content = data.split('\r\n\r\n')
```
kaustubh-karkare: This will try to find a "\r\n\r\n" in every block, instead of only the first.
Also, you might want to do something with the headers. Put them in a object or something.

solutions/downloadfile.py
```diff
@@ -0,0 +1,197 @@
...
+class DownloadFile(threading.Thread):
+
+    def __init__(self, url, threads, output_path):
+        self.url = url
+        self.threads = threads
+        self.output_path = output_path
+        threading.Thread.__init__(self)
+
+    def run(self):
+        self.url = urlparser.parse_url(self.url)
```
kaustubh-karkare: self.parsed_url

solutions/urlparser.py
```diff
@@ -0,0 +1,24 @@
+import re
+import collections
+
+
+def parse_url(url):
+    pattern = r'(?:[a-z]+:/+)?([a-zA-Z0-9.-]+)?(?::*)?(\d+)?(.*)?'
+    url_group = re.match(pattern, url)
+    UrlParser = (
```
kaustubh-karkare: This isn't a class that does anything, which is what the name suggests.
This is a collection that contains results.

solutions/urlparser.py
```diff
@@ -0,0 +1,24 @@
...
+        collections.namedtuple(
+            'UrlParser',
+            'host port path'
+        )
+    )
+    # url_group.group(1) contains host
+    # url_group.group(2) contains port
+    # url_group.group(3) contains path
```
kaustubh-karkare: I always prefer avoid comments if you can express this in code. Not sure if this is possible though:
```
URLParser(
  host=url_group.group(1),
  ...
```

solutions/downloadfile.py
```diff
@@ -0,0 +1,197 @@
...
+            self.threads = 1
+
+    def https_headers_to_send_to_host(self):
+        headers = {}
+        headers['Host: '] = self.url.host
+        headers['Connection: '] = 'close'
+        # Add headers as per the requirement
+        return headers
+
+    def request_and_receive_reponse_headers(self):
```
kaustubh-karkare: What I said about sharing code, it applies here. You should merge this and "receive_file_content_from_host" so that you have a single method which returns all headers and content. In case of head requests, there is no content. In case of get requests, feel free to ignore content. But lets not have duplicate that loop.

solutions/downloadfile.py
```diff
@@ -0,0 +1,197 @@
...
+        client_socket.send(send_to_host)
+        # Request is made only for headers
+        try:
+            status, separator, received_headers = (
+                client_socket.recv(BLOCK_SIZE).partition('\r\n')
+            )
+        except:
+            logging.error('Host not responding')
+            raise
+        received_headers = received_headers.splitlines()
```
kaustubh-karkare: You're assuming that headers fit in a single block. That's not always the case.

solutions/downloadfile.py
```diff
@@ -0,0 +1,197 @@
...
+        client_socket.close()
+        return headers
+
+    def download_files(self):
+        headers = self.request_and_receive_reponse_headers()
+        try:
+            content_length = int(headers['Content-Length'])
+        except:
+            content_length = 0
+        each_file_bytes = int(content_length / self.threads)
```
kaustubh-karkare: It looks like this logic to determine how many threads / size of each (which are related topics) is split across different places. Keep them together.
This method is called "download_files", so lets do only that. Also, call it "download_parts"?

solutions/downloadfile.py
```diff
@@ -0,0 +1,192 @@
...
+
+def receive_data_from_host(client_socket, file_content):
+    headers = ''
+    all_headers_received = False
+    while True:
+        data = client_socket.recv(BLOCK_SIZE)
+        if not data:
+            break
+        if not all_headers_received:
+            received_headers, separator, content = data.partition('\r\n\r\n')
```
kaustubh-karkare: What if that is across block boundaries? `\r\n\r` at the end of 2nd block, and `\n` at the beginning of 3d block.

solutions/downloadfile.py
```diff
@@ -0,0 +1,192 @@
...
+            file_content.write(data)
+    client_socket.close()
+    file_content.flush()
+    return headers
+
+
+class DownloadFile:
+
+    def __init__(self, url, threads, output_path):
+        self.parsed_url = urlparser.parse_url(url)
```
kaustubh-karkare: s/parsed_url/url_data ?

solutions/downloadfile.py
```diff
@@ -0,0 +1,192 @@
...
+        try:
+            if (headers['Accept-Ranges'] != 'bytes'):
+                self.threads = 1
+        except:
+            self.threads = 1
+
+    def https_headers_to_send_to_host(self):
+        headers = {}
+        headers['Host: '] = self.parsed_url.host
+        headers['Connection: '] = 'close'
```
kaustubh-karkare: Your keys shouldn't contain ": ". That is duplicated. You can just put it once after line 76.

solutions/downloadfile.py
```diff
@@ -0,0 +1,192 @@
...
+        line_break = '\r\n'
+        headers_to_send_to_host = (self.https_headers_to_send_to_host())
+        send_to_host = (
+            https_method +
+            ' ' +
+            self.parsed_url.path +
+            ' HTTP/1.1' +
+            line_break
+        )
+        for key, value in headers_to_send_to_host.iteritems():
```
kaustubh-karkare: Use list comprehension / string join.

solutions/downloadfile.py
```diff
@@ -0,0 +1,192 @@
...
+            self.threads = 1
+
+    def https_headers_to_send_to_host(self):
+        headers = {}
+        headers['Host: '] = self.parsed_url.host
+        headers['Connection: '] = 'close'
+        # Add headers as per the requirement
+        return headers
+
+    def get_data_to_send(self, https_method):
```
kaustubh-karkare: get_request?

solutions/downloadfile.py
```diff
@@ -0,0 +1,192 @@
...
+        except:
+            logging.warning('Download size not available')
+            self.threads = 1
+        try:
+            if (headers['Accept-Ranges'] != 'bytes'):
+                self.threads = 1
+        except:
+            self.threads = 1
+
+    def https_headers_to_send_to_host(self):
```
kaustubh-karkare: Too small to be a function IMO.

solutions/downloadfile.py
```diff
@@ -0,0 +1,192 @@
...
+                (
+                    self.parsed_url.host,
+                    int(self.parsed_url.port)
+                )
+            )
+        except:
+            logging.error('Client Socket connection could not be established')
+            raise
+        send_to_host = self.get_data_to_send('HEAD')
+        send_to_host += '\r\n'
```
kaustubh-karkare: Should that `\r\n` be a module level constant?

solutions/urlparser.py
```diff
@@ -0,0 +1,16 @@
+import re
+import collections
+
+
+def parse_url(url):
+    pattern = r'(?:[a-z]+:/+)?([a-zA-Z0-9.-]+)?(?::*)?(\d+)?(.*)?'
```
kaustubh-karkare: ([a-z]+)(?:\://)

solutions/urlparser.py
```diff
@@ -0,0 +1,16 @@
...
+    url_group = re.match(pattern, url)
+    url_tuple = collections.namedtuple('url_tuple', ['host', 'port', 'path'])
+    parsed_url = (
```
kaustubh-karkare: url_data

solutions/filedownloader.py
```diff
@@ -0,0 +1,220 @@
...
+    def __init__(self, url, threads, output_path):
+        self.url_data = urlparser.parse_url(url)
+        self.threads = threads
+        self.output_path = output_path
+
+    def start(self):
+        # If protocol and port are missing, it is assumed to be a https request
+        if (
+            self.url_data.port is None or
+            self.url_data.protocol is 'https' or
```
kaustubh-karkare: Why? You dont actually support https, only http.

solutions/filedownloader.py
```diff
@@ -0,0 +1,220 @@
...
+            )
+        return send_to_host
+
+    def request_and_receive_reponse_headers(self):
+        client_socket = socket.socket()
+        try:
+            client_socket.connect(
+                (
+                    self.url_data.host,
+                    int(self.url_data.port)
```
kaustubh-karkare: Shouldn't you cast to int in your urlparser?

solutions/filedownloader.py
```diff
@@ -0,0 +1,220 @@
...
+                )
+            )
+        except:
```
kaustubh-karkare: Be more specific about the exceptions you want to catch.

solutions/filedownloader.py
```diff
@@ -0,0 +1,220 @@
...
+        # Add more headers when required
+        for key, value in headers.iteritems():
+            send_to_host += (
+                ''.join(
+                    (
+                        key,
+                        ': ',
+                        value,
+                        EOL
+                    )
```
kaustubh-karkare: This seems like unnecessary whitespace.
```
send_to_host += \
  "%s: %s%s" % (key, value, EOL)

File: solutions/filedownloader.py [hunk 58]
```diff
@@ -0,0 +1,220 @@
...
+            )
+        except:
+            logging.error('Client Socket connection could not be established')
+            raise
+        send_to_host = self.get_request('HEAD')
+        send_to_host += EOL
+        client_socket.send(send_to_host)
+        # Request is made only for headers
+        content_file = tempfile.NamedTemporaryFile()
+        received_headers = receive_data_from_host(client_socket, content_file)
```
kaustubh-karkare: How about you make the 2nd argument optional, so that you dont have to create an unnecessary temp file? That way, if you get anything beyond the headers, you can raise an exception.

solutions/filedownloader.py
```diff
@@ -0,0 +1,220 @@
...
+        received_headers = received_headers.splitlines()
+        headers = {}
+        for line in received_headers:
+            if not line:
+                continue
+            try:
+                key, separator, value = line.partition(':')
+                headers[key] = value[1:]
```
kaustubh-karkare: partition(': ') to avoid the `[1:]`

solutions/filedownloader.py
```diff
@@ -0,0 +1,220 @@
...
+
+def receive_data_from_host(client_socket, file_content):
+    headers = ''
+    previous_block = ''
+    all_headers_received = False
+    while True:
+        current_data = client_socket.recv(BLOCK_SIZE)
+        if not current_data:
+            break
+        if not all_headers_received:
```
kaustubh-karkare:
```
EOH = EOL + EOL
if EOH not in current_data:

File: solutions/filedownloader.py [hunk 61]
```diff
@@ -0,0 +1,220 @@
...
+            else:
+                headers += received_headers
+                file_content.write(content)
+                all_headers_received = True
+            previous_block = current_data
+        else:
+            file_content.write(current_data)
+    client_socket.close()
+    file_content.flush()
+    return headers
```
kaustubh-karkare: You might as well convert to a map (like in your webserver assignment) here, instead of asking the users of this method to do it.

solutions/filedownloader.py
```diff
@@ -0,0 +1,220 @@
...
+                    receive_data_from_host,
+                    client_socket,
+                    download_file
+                )
+            )
+            start_range = end_range + 1
+            end_range = start_range + each_part_size
+            ii -= 1
+
+        concurrent.futures.wait(download_threads)
```
kaustubh-karkare: Hmm, this shouldn't be necessary. The point of the threadpoolexecutor is the implicit wait when you use it like:
```
with ThreadPoolExecutor(...) as executor:
  for whatever:
    executor.submit(something)

File: solutions/filedownloader.py [hunk 63]
```diff
@@ -0,0 +1,220 @@
...
+        ii = 1
+        executor = (
+            concurrent.futures.ThreadPoolExecutor(
+                max_workers=self.threads
+            )
+        )
+        while ii <= self.threads:
+            if (start_range > content_length - 1):
+                # File already downloaded
+                break
```
kaustubh-karkare: This conditions feel wrong. I mean, it works, sure, but if you did your calculation of number of threads properly, it shouldn't be necessary.

solutions/filedownloader.py
```diff
@@ -0,0 +1,220 @@
...
+                        self.url_data.host,
+                        int(self.url_data.port)
+                    )
+                )
+            except:
+                logging.error('Client connection could not be made')
+                raise
+            send_to_host = self.get_request('GET')
+            if (self.threads == 1):
+                send_to_host += ('Range: bytes=0-' + EOL)
```
kaustubh-karkare: String concatenation is a bad idea. This should be an argument to your `get_request` method.

solutions/filedownloader.py
```diff
@@ -0,0 +1,220 @@
...
+            else:
+                send_to_host += (
+                    'Range: bytes=' +
```
kaustubh-karkare: `'Range: bytes={}-{}{}'.format(start_range, end_range, EOL)`

solutions/filedownloader.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def merge_downloaded_files(self, downloaded_parts_name):
+        try:
+            output_file = open(self.output_path, 'w')
+        except:
+            logging.error('Output file could not be created')
+            raise
+        for downloaded_file_name in downloaded_parts_name:
+            downloaded_file = open(downloaded_file_name, 'r')
+            downloaded_file.seek(0)
```
kaustubh-karkare: This is only necessary if you didn't close the file, as the pointer is at the end after all the writing. If you're opening a file again, I think this is already the case.

solutions/urlparser.py
```diff
@@ -0,0 +1,22 @@
+import re
+import collections
+
+
+def parse_url(url):
```
kaustubh-karkare: Ooc, how did you decide to put this in a separate file?

solutions/filedownloader.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+BLOCK_SIZE = 1024
+EOL = '\r\n'
+
+
+def receive_data_from_host(client_socket, file_content=None):
+    current_data = client_socket.recv(BLOCK_SIZE)
+    status, _, data = current_data.partition(EOL)
+    EOH = EOL + EOL
```
kaustubh-karkare: This can be outside too.

solutions/filedownloader.py
```diff
@@ -0,0 +1,190 @@
...
+import logging
+import urlparser
+from concurrent.futures import ThreadPoolExecutor
+
+
+BLOCK_SIZE = 1024
+EOL = '\r\n'
+
+
+def receive_data_from_host(client_socket, file_content=None):
```
kaustubh-karkare: This entire function is more complex than it needs to be. My implementation of this logic was around 15 lines, if that helps. Also, why is this outside, and not a class private method?

solutions/filedownloader.py
```diff
@@ -0,0 +1,190 @@
...
+    def initialize_download(self):
+        headers = self.request_and_receive_reponse_headers()
+        try:
+            if (int(headers['Content-Length']) == 0):
+                logging.warning('No data to download')
+        except KeyError:
+            logging.warning('Download size not available')
+            self.threads = 1
+        try:
+            if (headers['Accept-Ranges'] != 'bytes'):
```
kaustubh-karkare: `headers.get('Accept-Ranges')` will avoid the `KeyError`

solutions/filedownloader.py
```diff
@@ -0,0 +1,190 @@
...
+        headers['Host'] = self.url_data.host
+        headers['Connection'] = 'close'
+        # Add more headers when required
+        for key, value in headers.iteritems():
+            client_socket.send('%s: %s%s' % (key, value, EOL))
+        client_socket.send(EOL)
+        status, headers = receive_data_from_host(client_socket)
+        return headers
+
+    def download_parts(self):
```
kaustubh-karkare: How about ordering these methods in the same sequence as they appear in the code?
It was odd to find `request_and_receive_reponse_headers` before I see how it is used.

solutions/filedownloader.py
```diff
@@ -0,0 +1,190 @@
...
+        except socket.error:
+            logging.error('Client Socket connection could not be established')
+            raise
+        client_socket.send('HEAD %s HTTP/1.1%s' % (self.url_data.path, EOL))
+        headers = {}
+        headers['Host'] = self.url_data.host
+        headers['Connection'] = 'close'
+        # Add more headers when required
+        for key, value in headers.iteritems():
+            client_socket.send('%s: %s%s' % (key, value, EOL))
```
kaustubh-karkare: So much easier, na

solutions/filedownloader.py
```diff
@@ -0,0 +1,190 @@
...
+            content_length = int(headers['Content-Length'])
+        except KeyError:
+            content_length = 0
+        each_part_size = int(content_length / self.threads)
+        start_range = 0
+        end_range = each_part_size
+        downloaded_parts_name = []
+        ii = 1
+        with ThreadPoolExecutor(max_workers=self.threads) as executor:
+            while ii <= self.threads:
```
kaustubh-karkare: `for ii in range(self.threads):`

solutions/filedownloader.py
```diff
@@ -0,0 +1,190 @@
...
+                    logging.error('Client connection could not be made')
+                    raise
+                client_socket \
+                    .send('GET %s HTTP/1.1%s' % (self.url_data.path, EOL))
+                headers = {}
+                headers['Host'] = self.url_data.host
+                headers['Connection'] = 'close'
+                # Add more headers when required
+                for key, value in headers.iteritems():
+                    client_socket.send('%s: %s%s' % (key, value, EOL))
```
kaustubh-karkare: IMO you should put this request section in a function.
Make it easy to read the ThreadPoolExecutor block.

solutions/filedownloader.py
```diff
@@ -0,0 +1,190 @@
...
+                if self.threads != 1:
+                    client_socket.send(
+                        'Range: bytes=%s-%s%s' % (
```
kaustubh-karkare: This should be part of the header object we have above.

solutions/filedownloader.py
```diff
@@ -0,0 +1,190 @@
...
+        with ThreadPoolExecutor(max_workers=self.threads) as executor:
+            while ii <= self.threads:
+                if end_range >= content_length:
+                    end_range = content_length - 1
+                try:
+                    download_file = tempfile.NamedTemporaryFile(delete=False)
+                except:
+                    logging.error('Temporary file could not be created')
+                    raise
+                downloaded_parts_name.append(download_file.name)
```
kaustubh-karkare: Why not just store a reference to the file directly? Instead of just the name?
That way, instead of re-opening, you can just reset the point.
I imagine there is some cost of reopening files.

solutions/urlparser.py
```diff
@@ -0,0 +1,25 @@
...
+    pattern = r'(?:([a-z]+)?(?::/+))?([a-zA-Z0-9.-]+)?(?::*)?(\d+)?(.*)?'
+    url_group = re.match(pattern, url)
+    url_tuple = (
+        collections.namedtuple(
```
kaustubh-karkare: You should define the url_tuple outside the method since it never changes across method calls.

solutions/urlparser.py
```diff
@@ -0,0 +1,25 @@
...
+            'url_tuple',
+            ['protocol', 'host', 'port', 'path']
+        )
+    )
+    server_port = None
```
kaustubh-karkare: The "server_" prefix confused me a bit (just "port" might be sufficient), but that's okay.
Although technically, you might just be hitting a proxy, which isn't the actual final server.

solutions/filedownloader.py
```diff
@@ -0,0 +1,150 @@
...
+        self.url_data = urlparser.parse_url(url)
+        self.threads = threads
+        self.output_path = output_path
+
+    def start(self):
+        # If protocol and port are missing, it is assumed to be a http request
+        if (
+            self.url_data.port is None or
+            self.url_data.protocol is 'http' or
+            self.url_data.port == 8080
```
kaustubh-karkare: This is odd. You shouldn't update the port if it is explicitly specified.

solutions/filedownloader.py
```diff
@@ -0,0 +1,150 @@
...
+        ):
+            self.url_data = self.url_data._replace(port=80)
+        downloaded_parts_reference = self.download_parts()
+        self.merge_downloaded_files(downloaded_parts_reference)
+
+    def send_request(
+        self,
+        http_method,
+        client_socket,
```
kaustubh-karkare: I'd think you can create the client_socket in here? That'll avoid duplication outside, since the creation doesn't vary.

solutions/filedownloader.py
```diff
@@ -0,0 +1,150 @@
...
+            headers['Range'] = \
+                'bytes=%s-%s' % (str(start_range), str(end_range))
+        # Add more headers when required
+        for key, value in headers.iteritems():
+            client_socket.send('%s: %s%s' % (key, value, EOL))
+        client_socket.send(EOL)
+
+    def receive_data_from_host(self, client_socket, file_content=None):
+        current_data = client_socket.recv(BLOCK_SIZE)
+        status, _, current_data = current_data.partition(EOL)
```
kaustubh-karkare: Haha, you still assume that first line will be less than 1024 bytes. While it is reasonably, it isn't necessarily true.
Check [this](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers) out, which among other things, says that the upper limit is >2000. I'd rather not making an assumption based on the BLOCK SIZE either way.

solutions/filedownloader.py
```diff
@@ -0,0 +1,150 @@
...
+                header_data += headers
+                break
+            elif EOH in (current_data + previous_block):
+                _, _, current_data = \
+                    (current_data + previous_block).partition(EOH)
+                break
+            else:
+                header_data += current_data
+                previous_block = current_data
+            current_data = client_socket.recv(BLOCK_SIZE)
```
kaustubh-karkare: Might I suggest a simpler version:
```
while True:
  current_data += client_socket.recv(BLOCK_SIZE)
  if EOH not in current_data:
    break
# extract headers
# continue reading from socket to insert in file
```
Or, if you're feeling particularly performance aware,
```
if EOH in previous_block[-1] + current_block:

File: solutions/filedownloader.py [hunk 83]
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self, url, threads, output_path):
+        self.url_data = urlparser.parse_url(url)
+        self.threads = threads
+        self.output_path = output_path
+
+    def start(self):
+        # If protocol and port are missing, it is assumed to be a http request
+        if (
+            self.url_data.port is None or
+            self.url_data.protocol is 'http'
```
kaustubh-karkare: It is possible to run http servers on ports other than 80 too. There's no restriction, only convention, so 80 is a reasonable default, but no need to override it if provided.

solutions/filedownloader.py
```diff
@@ -0,0 +1,146 @@
...
+        for key, value in headers.iteritems():
+            client_socket.send('%s: %s%s' % (key, value, EOL))
+        client_socket.send(EOL)
+
+    def receive_data_from_host(self, client_socket, file_content=None):
+        current_data = ''
+        while True:
+            current_data += client_socket.recv(BLOCK_SIZE)
+            if EOL in current_data:
+                break
```
kaustubh-karkare:
```
current_data = ''
while EOL not in current_data:
  current_data += client_socket.recv(BLOCK_SIZE)

File: solutions/filedownloader.py [hunk 85]
```diff
@@ -0,0 +1,146 @@
...
+        status, _, current_data = current_data.partition(EOL)
+        while True:
+            current_data += client_socket.recv(BLOCK_SIZE)
+            if EOH in current_data:
+                break
+        header_data, _, current_data = current_data.partition(EOH)
+        while True:
```
kaustubh-karkare: If this were a HEAD request, you already know that there's no content. Possible optimization.

solutions/filedownloader.py
```diff
@@ -0,0 +1,146 @@
...
+    def request_and_receive_reponse_headers(self):
+        client_socket = socket.socket()
+        try:
+            client_socket.connect((self.url_data.host, self.url_data.port))
+        except socket.error:
+            logging.error('Client Socket connection could not be established')
+            raise
+        self.send_request('HEAD', client_socket)
+        status, headers = self.receive_data_from_host(client_socket)
+        print(headers)
```
kaustubh-karkare: Remove this.

solutions/urlparser.py
```diff
@@ -0,0 +1,27 @@
+import re
+import collections
+
+
+url_tuple = (
```
kaustubh-karkare: Pretty sure this violates some naming convention.

PR #33, opened 2017-04-01
Problem 2 2

Abhijeetsng97:


solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+			byte_range = 'bytes='+str(start_byte)+'-'+str(last_byte)		
+		header = 'GET ' + self.path + ' HTTP/1.1\r\nHost: ' + self.host
+		header += '\r\nConnection: keep-alive\r\nAccept-Encoding: gzip, '
+		header += 'deflate\r\nAccept: */*\r\nUser-Agent: python-requests/2.13.0'
+		header += '\r\nRange: ' + byte_range + '\r\n\r\n'
+		return header
+
+	
+# takes input from user about what to download
+if __name__ == '__main__':
```
kaustubh-karkare: Put this in a separate file. Your downloading logic is different from your command line interface to it.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+		help ="Port on which file will be downloaded", 
+		default='80'
+	)
+	download_info.add_argument(
+		"-pa", 
+		"--parts", 
+		help="Parts in which file will be downloaded", 
+		default='3'
+	)
+	url = download_info.parse_args().url
```
kaustubh-karkare: This should be done inside your class, not here.

Instead of using string operations to extra host / port / path from URL, use regular expressions (the `re` module).

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+		required=True
+	)
+	download_info.add_argument(
+		"-p",
+		"--port", 
+		help ="Port on which file will be downloaded", 
+		default='80'
+	)
+	download_info.add_argument(
+		"-pa",
```
kaustubh-karkare: Short names are generally a single letter, not more.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+	)
+	download_info.add_argument(
+		"-p",
+		"--port", 
+		help ="Port on which file will be downloaded", 
+		default='80'
+	)
+	download_info.add_argument(
+		"-pa", 
+		"--parts",
```
kaustubh-karkare: `--threads` is more accurate.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+	download_info.add_argument(
+		"-pa", 
+		"--parts", 
+		help="Parts in which file will be downloaded", 
+		default='3'
+	)
+	url = download_info.parse_args().url
+	if url[:7] == 'http://':
+		url = url[7:]
+	download =	DownloadManager(url, download_info.parse_args().port,
```
kaustubh-karkare: This is indented weirdly.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+import tempfile
+
+class DownloadManager():
+	
+	def __init__(self, url, port , parts):
+		self.threads = []
+		self.file_parts = []
+		self.logger = logging.getLogger("Download manager")
+		logging.basicConfig(level=logging.INFO)
+		self.url = url
```
kaustubh-karkare: Since you're already storing host / path, do you really need to store this?
Lets see what is the minimum number of class properties you can do this with.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+		self.parts = int(parts)
+		self.logger.info("Download initied")
+		self.client = socket.socket(socket.AF_INET,socket.SOCK_STREAM)
+		host_index = self.url.find('/')
+		self.host = self.url[: host_index]
+		self.path = self.url[host_index :]
```
kaustubh-karkare: Not sure if there's supposed to be spaces next to `:`

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+		ip = socket.gethostbyname(self.host)
+		port = int(port)
+		self.client.connect((ip,port))
+		header = self.get_http_header(0, 0)
```
kaustubh-karkare: It isn't obvious what these arguments are. In cases like this, provide the argument name too, so that I dont have to go look at the implementation.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+		try:
```
kaustubh-karkare: This is odd. So you'll download exactly 1 block and wait? Why not do this in start too? What is the guarantee that all the headers will fit in 1024 bytes?

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+		self.logger.info(
+			"Downloading File --> {}".format(os.path.basename(self.url))
+		)
+		# for finding content-length from header
+		start_content_length = self.client_received.lower()
+			.find('content-length: ')
+		end_content_length = self.client_received.find(
+			'\r\n', 
+			start_content_length + 16
+		)
```
kaustubh-karkare: Uhh ... this is inelegant. Put all the headers in a dictionary (key = header name) and then extract what you want.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+	
+	# download starts over here
+	def start(self):
+		self.save_file = open(os.path.basename(self.url), "w")
+		self.logger.info(
+			"Downloading File --> {}".format(os.path.basename(self.url))
+		)
+		# for finding content-length from header
+		start_content_length = self.client_received.lower()
+			.find('content-length: ')
```
kaustubh-karkare: What if the webserver doesn't return a content length header?

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+		end_content_length = self.client_received.find(
+			'\r\n', 
+			start_content_length + 16
+		)
+		file_size = int(
+			self.client_received[start_content_length+16 : end_content_length]
+		)
+		start_byte = 0		
+		for thread_num in range(0, self.parts + 1):
```
kaustubh-karkare: So if I say 4 parts, you can create upto 5? Isn't that dishonest? :P

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+  				break
+			temp_file.write(data)
+
+	# returns the http header request
+	def get_http_header(self, start_byte, last_byte):
+		if start_byte == 0 and last_byte == 0:
+			byte_range = ""
+		else: 
+			byte_range = 'bytes='+str(start_byte)+'-'+str(last_byte)		
+		header = 'GET ' + self.path + ' HTTP/1.1\r\nHost: ' + self.host
```
kaustubh-karkare: Instead of GET, you might want to initially use the HEAD request.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+		header += '\r\nConnection: keep-alive\r\nAccept-Encoding: gzip, '
```
kaustubh-karkare: Do you really accept gzip / deflate?

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,127 @@
...
+		header += 'deflate\r\nAccept: */*\r\nUser-Agent: python-requests/2.13.0'
+		header += '\r\nRange: ' + byte_range + '\r\n\r\n'
```
kaustubh-karkare: Put the `\r\n` in a constant. You're using it so many times.
I'd also recommend constructing this string from a dictionary as mentioned before.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+import tempfile
+
+class DownloadManager():
+	
+	def __init__(self, ip, port, url, parts):
+		self.threads = []
+		self.file_parts = []
+		self.logger = logging.getLogger("Download manager")
+		logging.basicConfig(level=logging.INFO)
+		if url[ : 7] == 'http://':
```
kaustubh-karkare: Use regular expressions to parse the path. String manipulation should generally be avoided.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
+import os
+import socket
+import threading
+import logging
+import argparse
+import tempfile
+
+class DownloadManager():
+	
+	def __init__(self, ip, port, url, parts):
```
kaustubh-karkare: You dont need ip/port as input. You can get them from the URL.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+		logging.basicConfig(level=logging.INFO)
+		if url[ : 7] == 'http://':
+			url = url[7 : ]
+		host_index = url.find('/')
+		self.host = url[ : host_index]
+		self.path = url[host_index : ]
+		self.parts = int(parts)
+		self.logger.info("Download initied")
+		self.client = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+		if(ip == '0.0.0.0'):
```
kaustubh-karkare: Yeah, this should never happen.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+			ip_client = socket.gethostbyname(self.host)
+		else:
+			ip_client = ip
+		port = int(port)
+		self.client.connect((ip_client,port))
+		header = self.get_http_header('HEAD', 0, 0)
+		try:
+			self.client.send(header)
+			client_received = self.client.recv(1024)
```
kaustubh-karkare: Put the 1024 in a constant.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+			self.header_information = {
+				'HTTP Version': client_received[ : 8], 
+				'Status': client_received[9 : 12]
```
kaustubh-karkare: This is so fragile. The moment an extra character comes in, your program will stop working. Also, there's no way I can validate these indices easily, which means that they're a source of future bugs.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+			}
+			start = 0
+			self.end_dictionary = client_received.find('\r\n\r\n')
```
kaustubh-karkare: Use constants at the top of the file.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+			while True:
+				start = client_received.find('\r\n', start)
+				tag_start = client_received.find(': ', start)
+				tag_name = client_received[start + 2 : tag_start]
+				end = client_received.find('\r\n', tag_start)
+				start = end
+				self.header_information[tag_name] = client_received[
```
kaustubh-karkare: I like this structure. But there's a much easier way to create it.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+			client_received = self.client.recv(1024)
+			self.header_information = {
+				'HTTP Version': client_received[ : 8], 
+				'Status': client_received[9 : 12]
+			}
+			start = 0
+			self.end_dictionary = client_received.find('\r\n\r\n')
+			while True:
+				start = client_received.find('\r\n', start)
+				tag_start = client_received.find(': ', start)
```
kaustubh-karkare: "Name: Value".partition(": ")

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+				last_byte = (file_size / self.parts) * thread_num
+				if thread_num == self.parts - 1:
+					last_byte = file_size
+				thread = threading.Thread(
+					target=self.thread_download, 
+					args=(start_byte, last_byte, temp_file)
+				)
+				start_byte = (file_size /  self.parts) * thread_num + 1
+				thread.start()
+				self.threads.append(thread)
```
kaustubh-karkare: Use ThreadPoolExecutor.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+			temp_file = tempfile.TemporaryFile()
+			self.file_parts.append(temp_file)
+			# -1, -1 because it has no Content-Length and therefore all 
+			# information will be downloaded in a single tempfile & thread 
+			thread = threading.Thread(
+				target=self.thread_download, 
+				args=(-1, -1, temp_file)
+			)
+			thread.start()
+			self.threads.append(thread)
```
kaustubh-karkare: Instead of duplicating this part as a separate clause, how about something like:
```
if content_length:
  download_size = content_length / thread_count
else:
  thread_count = 1
  download_size = -1
for index in range(thread_count):
  start_index = index * download_size

File: solutions/assignment-2.2.py [hunk 26]
```diff
@@ -0,0 +1,177 @@
...
+		
+	# waiting for all the threads to end and end the download process
+	def wait(self):
```
kaustubh-karkare: A function called "wait", should wait, nothing else. You need a better name here. You shouldn't need that comment.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+  				break
+			temp_file.write(data)
+
+	# returns the http header request
+	def get_http_header(self, request, start_byte, last_byte):
+		next_line = '\r\n'
+		request_header = {
+			'Connection': 'keep-alive',
+			'Accept': '*/*',
+			'Accept-Encoding': 'gzip, deflate',
```
kaustubh-karkare: Do you really accept these?

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+			data = self.client.recv(1024)
+			if not data:
+  				break
+			temp_file.write(data)
+
+	# returns the http header request
+	def get_http_header(self, request, start_byte, last_byte):
+		next_line = '\r\n'
+		request_header = {
+			'Connection': 'keep-alive',
```
kaustubh-karkare: Do you need this?

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+			'Accept': '*/*',
+			'Accept-Encoding': 'gzip, deflate',
+			'User-Agent': 'python-requests/2.13.0'
+		}
+		header = request + ' ' + self.path + ' HTTP/1.1' + next_line
+		header += 'Host: ' + self.host + next_line + 'Connection: '
+		header += request_header['Connection'] + next_line
+		header += 'Accept-Encoding: ' + request_header['Accept-Encoding']
+		header += next_line + 'Accept: ' + request_header['Accept']
+		header += next_line + 'User-Agent: ' + request_header['User-Agent']
```
kaustubh-karkare: You had a nice dict to contain response headers. Do the same thing for request headers.

solutions/assignment-2.2.py
```diff
@@ -0,0 +1,177 @@
...
+			if start_byte == 0 and last_byte == 0:
+				byte_range = ""
+			else:
+				byte_range = 'bytes='+str(start_byte)+'-'+str(last_byte)
+			header += next_line + 'Range: ' + byte_range
+			return_header = header + next_line + next_line
+		return return_header
+	
+# takes input from user about what to download
+if __name__ == '__main__':
```
kaustubh-karkare: Put this in a separate file.
The functionality of your module can be separated from the interface to use it.

kaustubh-karkare:
Inactive for too long.

PR #36, opened 2017-04-12
Solution to Assignment 2.1 added

arch-DJ:
**@kaustubh-karkare** Please review
- I've tried to implement Client-Server Interaction using Java.

- There are 4 source code files namely Server.java, ServerBuilder.java, User.java and Client.java.

- Server.java is the implementation of server package and it is used to start a server on a specified ip, port and backlog. It also takes logic function as input. It continuously waits for Client to connect and as soon as a Client gets connected it uses ClientHandlerThread class to handle those requests on a separate thread. The ClientHandlerThread class processes those requests according to the input logic and sends response back to the Client.

- ServerBuilder.java is the builder class of Server.java based on Builder Pattern technique.

- User.java is the class which is handled by a user sitting on the server side. It uses the server package to start the server on a specified ip and port. This class sends the logic function to the server.

- Client.java is the class handled by the client which connects to a server on a specific ip and port and sends 2 numbers for a particular operation.


solutions/Assignment-2.1/User.java
```diff
@@ -0,0 +1,99 @@
...
+		port = namespace.getInt("port");
+		ip = namespace.getString("ip");
+		backlog = namespace.getInt("backlog");
+
+		// Creating Method object of Logic Function to be passed to the server
+		Class[] parameterTypes = new Class[2];
+		parameterTypes[0] = int.class;
+		parameterTypes[1] = int.class;
+		try {
+			logicFunction = User.class.getMethod("logicFunction", parameterTypes);
```
kaustubh-karkare: Well, I didn't know this was possible in Java, so that's interesting.
A bit unfortunate how verbose this is though. But there's Java for you.

solutions/Assignment-2.1/server/Server.java
```diff
@@ -0,0 +1,107 @@
...
+
+public class Server extends Thread {
+	private static final Logger logger = Logger.getLogger(Server.class.getName());
+	private Method logicFunction;
+	private int port;
+	private String ip;
+	private int backlog;
+	private ServerSocket serverSocket;
+	private Socket socket;
+	private Object userObject;
```
kaustubh-karkare: Make your types as specific as possible.

solutions/Assignment-2.1/server/Server.java
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	// Method to process the received data
+	String processData(List<Integer> input, Object userObject) {
+		String reply = "";
+		Object[] parameters = new Object[2];
+		parameters[0] = input.get(0);
+		parameters[1] = input.get(1);
+		try {
+			reply = logicFunction.invoke(userObject, parameters).toString();
```
kaustubh-karkare: So your logic function is forced to take an exactly 2 numbers, and return 1 thing? That isn't truly reusable.
Imagine if we wanted to implement a webserver in the logicFunction. This wont support it, but in theory, it should.

solutions/Assignment-2.1/server/ServerBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+		return this;
+	}
+	
+	public ServerBuilder setUserObject(Object userObject) {
+		this.userObject = userObject;
+		return this;
+	}
+	
+	public Server getServer() throws IOException {
+		return new Server(port, ip, backlog, logicFunction, userObject);
```
kaustubh-karkare: I asked about this in a different assignment, but what happens if you call this without setting any arguments? There's no validation here.

kaustubh-karkare:
Inactive for too long.

PR #37, opened 2017-04-14
assignment-2-3.py

humanoid2795:


solutions/assignment-2-3.py
```diff
@@ -0,0 +1,65 @@
...
+        self.port = port
+        threading.Thread.__init__(self)
+
+    def run(self):
+        self.start_serving()
+
+    def start_serving(self):
+        try:
+            server_socket = socket.socket()
+            server_socket.bind((self.host, self.port))
```
kaustubh-karkare: If you bind in the constructor, you can save the socket on the class instead of the host / port. Then you can start listening here.

solutions/assignment-2-3.py
```diff
@@ -0,0 +1,65 @@
...
+        self.server_listening = True
+        client_connections = []
+        # Assuming maximum simultaneous request to be 5
+        executor = concurrent.futures.ThreadPoolExecutor(max_workers=5)
+        while self.server_listening:
+            try:
+                client_socket, client_address = server_socket.accept()
+                client_connections.append(
+                    executor.submit(
+                        ReceiveAndProcessRequest,
```
kaustubh-karkare: This threw me off a bit. Capital letters indicate classes, but the concept of threading makes more sense in terms of functions, and although this works, I wouldn't want to conflate "create this object" and "do this thing". Use a `@classmethod`?

solutions/processhttprequest.py
```diff
@@ -0,0 +1,119 @@
...
+EOL = '\r\n'
+
+
+class ReceiveAndProcessRequest:
+
+    def __init__(self, client_socket, client_address):
+        self.client_socket = client_socket
+        self.client_address = client_address
+        self.default_file = 'index.html'
+        logging.info('Client connection received from:', self.client_address)
```
kaustubh-karkare: Instead of generic logging, use a specific logger object. That way, each request can log to a separate file.

solutions/processhttprequest.py
```diff
@@ -0,0 +1,119 @@
...
+        try:
+            http_request = self.client_socket.recv(BLOCK_SIZE)
```
kaustubh-karkare: You cant assume that the request will fit in 1 block.

solutions/httprequestparser.py
```diff
@@ -0,0 +1,34 @@
+import re
+from collections import namedtuple
+
+
+def parse_http_request(http_request):
+    request_line, separator, received_headers = (
+        http_request.partition(
+            '\r\n'
```
kaustubh-karkare: Put this in a constant.

solutions/httprequestparser.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+    http_request_data = (
+        request_line_tuple(
+            method=request_line_group.group(1),
+            path=request_line_group.group(2),
+            version=request_line_group.group(3),
+        )
+    )
+    headers = {}
+    header_pattern = (r'([A-Za-z-]+)\:\s+(.*)?')
```
kaustubh-karkare: I'm not sure you need a regex for this one. `.partition(": ")` should work, right?

solutions/processhttprequest.py
```diff
@@ -0,0 +1,119 @@
...
+            raise
+        self.process_request(http_request)
+        self.client_socket.close()
+
+    def process_request(self, http_request):
+        http_request_data, headers = parse_http_request(http_request)
+        http_request_data = (
+            http_request_data._replace(
+                path=os.path.join(
+                    os.getcwd(),
```
kaustubh-karkare: Dont assume that the current directory is the base path. This should be configurable.

http://www.tecmint.com/change-root-directory-of-apache-web-server/

solutions/processhttprequest.py
```diff
@@ -0,0 +1,119 @@
...
+                EOL +
+                'Content-Length: ' +
+                str(
+                    os.path.getsize(
+                        http_request_data.path
+                    )
+                ) +
+                EOL +
+                EOL
+            )
```
kaustubh-karkare: String concatenation is inelegant here. Maintain a headers dict, and use that to generate the final response.

solutions/processhttprequest.py
```diff
@@ -0,0 +1,119 @@
...
+            requested_file = open(http_request_data.path, 'r')
+            self.client_socket.send(requested_file.read())
```
kaustubh-karkare: Read and write in blocks.

solutions/processhttprequest.py
```diff
@@ -0,0 +1,119 @@
...
+
+        elif os.path.isdir(http_request_data.path):
+            file_list = NamedTemporaryFile()
```
kaustubh-karkare: Why do you need to have a temp file here?

solutions/assignment-2-3.py
```diff
@@ -0,0 +1,72 @@
...
+
+class Server(threading.Thread):
+
+    def __init__(self, host, port, base_directory):
+        try:
+            self.base_directory = base_directory
+            self.server_socket = socket.socket()
+            self.server_socket.bind((host, port))
+            self.server_socket.settimeout(1)
+        except:
```
kaustubh-karkare: Catch only the exceptions that you need. I've seen many problems caused by unnecessary suppression of exceptions.

solutions/assignment-2-3.py
```diff
@@ -0,0 +1,72 @@
+import socket
+import logging
+import argparse
+import threading
+import time
+from os import getcwd
+from processhttprequest import ReceiveAndProcessRequest
+
+
+class Server(threading.Thread):
```
kaustubh-karkare: I feel like this should be in its own file.
This one is the entry point which can collect CLI arguments, but in order to make the Server class truly reusable, you'd need to move this out.

solutions/assignment-2-3.py
```diff
@@ -0,0 +1,72 @@
...
+        self.server_listening = False
+        self.server_socket.close()
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--host', required=True, type=str, help='Server Host')
+    parser.add_argument('--port', required=True, type=int, help='Server port')
+    parser.add_argument(
+        '--base-directory',
```
kaustubh-karkare: Lets stick with the standard name: `--document-root`

solutions/httprequestparser.py
```diff
@@ -0,0 +1,35 @@
...
+
+EOL = '\r\n'
+
+
+def parse_http_request(http_request):
+    request_line, separator, received_headers = (
+        http_request.partition(
+            EOL
+        )
+    )
```
kaustubh-karkare: This is fine too:
```
request_line, _, received_headers = \
  http_request.partition(EOL)
```
`_` (or local variables starting with `_`) indicates unused variables.

solutions/httprequestparser.py
```diff
@@ -0,0 +1,35 @@
...
+    pattern = (r'([A-Z]+)(?:\s+)(\S+)(?:\s+)(.*)')
+    request_line_group = re.match(pattern, request_line)
+    request_line_tuple = (
```
kaustubh-karkare: Why does this have to be a namedtuple? Why not just put this in a dict? In fact, why not reuse the same dict, with special keys like "__method", etc.

solutions/processhttprequest.py
```diff
@@ -0,0 +1,127 @@
...
+import logging
+import socket
+import sys
+from httprequestparser import parse_http_request
+
+BLOCK_SIZE = 1024
+EOL = '\r\n'
+
+
+class ReceiveAndProcessRequest:
```
kaustubh-karkare: This feels weird to me. I think I commented on this before, but this name suggests a function, not a class. Could you fix it?

solutions/processhttprequest.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __init__(self, client_socket, client_address, base_directory):
+        logging.info('Logger for client request')
+        self.logger = logging.getLogger(__name__)
```
kaustubh-karkare: Lets use the specific client socket + time in the name, instead of __name__.

solutions/assignment-2-3.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def run(self):
+        self.start_serving()
+
+    def start_serving(self):
+        self.server_socket.listen(1)
+        self.server_listening = True
+        while self.server_listening:
+            try:
+                self.client_socket, self.client_address = (
```
kaustubh-karkare: Why do you need to save these on the class? Why cant these be local variables?

solutions/processhttprequest.py
```diff
@@ -0,0 +1,127 @@
...
+import sys
+from httprequestparser import parse_http_request
+
+BLOCK_SIZE = 1024
+EOL = '\r\n'
+
+
+class ReceiveAndProcessRequest:
+
+    def __init__(self, client_socket, client_address, base_directory):
```
kaustubh-karkare: What if I added an additional constraint like: "request_timeout" (ie - if it takes longer than 5 seconds to generate a response, kill the thread)? You'd need to add more arguments here. How about create an options object that you can pass, which contains an arbitrary number of configuration options.

solutions/httprequestparser.py
```diff
@@ -0,0 +1,35 @@
+import re
+from collections import namedtuple
+
+EOL = '\r\n'
+
+
+def parse_http_request(http_request):
```
kaustubh-karkare: Can you explain why this is a separate function in its own file instead of in the other?

solutions/processhttprequest.py
```diff
@@ -0,0 +1,127 @@
...
+        http_request_data, headers = parse_http_request(http_request)
+        http_request_data = (
+            http_request_data._replace(
+                path=os.path.join(
+                    self.base_directory,
+                    http_request_data.path[1:]
+                )
+            )
+        )
+        self.logger.info('Path request by client is', http_request_data.path)
```
kaustubh-karkare: Shouldn't this happen before you add the base directory?

solutions/processhttprequest.py
```diff
@@ -0,0 +1,127 @@
...
+            file_to_send = http_request_data.path
+            status_code = '200'
+            status_message = 'OK'
+            response_headers['Content-Length'] = (
+                str(
+                    os.path.getsize(
+                        file_to_send
+                    )
+                )
+            )
```
kaustubh-karkare:
```
response_headers['Content-Length'] = \
  str(os.path.getsize(file_to_send))

File: solutions/processhttprequest.py [hunk 23]
```diff
@@ -0,0 +1,127 @@
...
+
+        elif os.path.isdir(http_request_data.path):
+            default_file_path = (
+                self.search_in_directory(
+                    http_request_data.path,
+                    )
+                )
```
kaustubh-karkare: See previous formatting advice. Applies to other places in code too.

solutions/processhttprequest.py
```diff
@@ -0,0 +1,127 @@
...
+                while content_of_file:
+                    self.client_socket.send(content_of_file)
+                    content_of_file = send_content.read(BLOCK_SIZE)
+        elif status_code is '200':
+            for element in os.listdir(http_request_data.path):
+                self.client_socket.send(element + EOL)
+        else:
+            return
+
+    def search_in_directory(self, current_directory):
```
kaustubh-karkare: This seems unnecessary convoluted. Why not just do `os.path.isfile(os.path.join(current_directory, default_file))`?

solutions/processhttprequest.py
```diff
@@ -0,0 +1,127 @@
...
+                    http_request_data.path,
+                    )
+                )
+            status_code = '200'
+            status_message = 'OK'
+            if default_file_path is None:
+                response_headers['Content-Length'] = (
+                    str(
+                        sys.getsizeof(
+                            os.listdir(
```
kaustubh-karkare: Lets return proper html, instead of raw output like this.
Use <ul>, <li> and <a> tags so that you can nagivate around.

solutions/processhttprequest.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+
+        elif os.path.isdir(http_request_data.path):
+            default_file_path = (
+                self.search_in_directory(
+                    http_request_data.path,
+                    )
+                )
+            status_code = '200'
+            status_message = 'OK'
```
kaustubh-karkare: Create a dict from status_code to status_message, since that mapping is fixed, and you shouldn't need to repeat it.

solutions/processhttprequest.py
```diff
@@ -0,0 +1,127 @@
...
+                )
+            )
+
+        elif os.path.isdir(http_request_data.path):
+            default_file_path = (
+                self.search_in_directory(
+                    http_request_data.path,
+                    )
+                )
+            status_code = '200'
```
kaustubh-karkare: These should be numbers, not strings.

solutions/assignment-2-3.py
```diff
@@ -0,0 +1,28 @@
...
+import os
+import time
+
+from server import Server
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--host', required=True, help='Server host')
+    parser.add_argument('--port', required=True, type=int, help='Server port')
```
kaustubh-karkare: You can default this to 80 or 8080

solutions/handleclient.py
```diff
@@ -0,0 +1,116 @@
...
+EOL = '\r\n'
+EOH = EOL + EOL
+
+
+class HandleClient(threading.Thread):
+    status = {}
+    status[200] = 'OK'
+    status[400] = 'Bad Request'
+    status[404] = 'Not Found'
+    status[501] = 'Not Implemented'
```
kaustubh-karkare: Avoid some repetition by doing
```
status = {
  200: 'OK',

File: solutions/handleclient.py [hunk 30]
```diff
@@ -0,0 +1,116 @@
...
+
+    def __init__(self, client_socket, client_address, **kwargs):
+        self.client_socket = client_socket
+        self.client_socket.settimeout(1)
+        threading.Thread.__init__(self)
```
kaustubh-karkare: Can you use `super` here?

solutions/handleclient.py
```diff
@@ -0,0 +1,116 @@
...
+        client_request = dict()
+        headers = dict()
+        data = ''
+        while EOL not in data:
+            try:
+                data += self.client_socket.recv(BLOCK_SIZE)
+            except socket.timeout:
+                break
+        request_line, _, data = data.partition(EOL)
+        pattern = r'([A-Z]+)\s+/*(\S*)\s+(HTTP/\d+\.\d+)'
```
kaustubh-karkare: You can put this outside the function. As a constant.

solutions/handleclient.py
```diff
@@ -0,0 +1,116 @@
...
+        try:
+            request_line = re.match(pattern, request_line).groups()
```
kaustubh-karkare: Use a different variable. Dont store data of different types in the same one.

solutions/handleclient.py
```diff
@@ -0,0 +1,116 @@
...
+        except AttributeError:
+            return client_request, headers
+        client_request['method'] = request_line[0]
+        client_request['path'] = request_line[1]
+        client_request['version'] = request_line[2]
+        while EOH not in data:
+            try:
+                data += self.client_socket.recv(BLOCK_SIZE)
+            except socket.timeout:
+                break
```
kaustubh-karkare: Should this be continue?

solutions/handleclient.py
```diff
@@ -0,0 +1,116 @@
...
+        self.client_socket.settimeout(1)
+        threading.Thread.__init__(self)
+
+    def run(self):
+        client_request, request_headers = self.recieve_request()
+        response_headers = {}
+        if client_request.get('method') == 'GET':
+            requested_path = \
+                os.path.join(os.getcwd(), client_request.get('path'))
+            if client_request.get('path') is None:
```
kaustubh-karkare: Shouldn't this be done before the previous file?

solutions/handleclient.py
```diff
@@ -0,0 +1,116 @@
...
+                self.send_file(requested_path, response_headers)
+            elif os.path.isdir(requested_path):
+                default_path = \
+                    os.path.join(requested_path, 'index.html')
+                if os.path.isfile(default_path):
+                    self.send_file(default_path, response_headers)
+                else:
+                    self.send_directory_content(
+                        client_request['path'],
+                        response_headers
```
kaustubh-karkare: trailing comma

solutions/handleclient.py
```diff
@@ -0,0 +1,116 @@
...
+                    )
+            else:
+                self.send_status_and_headers(404, response_headers)
+        elif client_request.get('method') is None:
+            self.send_status_and_headers(400, response_headers)
```
kaustubh-karkare: You might want to get the 400 cases out of the way first, so the rest valid cases are easier to read.

solutions/handleclient.py
```diff
@@ -0,0 +1,116 @@
...
+            key, _, value = line.partition(': ')
+            headers[key] = value
+        return client_request, headers
+
+    def send_status_and_headers(self, status_code, headers):
+        self.client_socket.send(
+            'HTTP/1.1 %d %s%s' % (
+                status_code,
+                HandleClient.status[status_code],
+                EOL
```
kaustubh-karkare: trailing comma

solutions/handleclient.py
```diff
@@ -0,0 +1,116 @@
...
+                data = requested_file.read(BLOCK_SIZE)
+                if not data:
+                    break
+                self.client_socket.send(data)
+
+    def send_directory_content(self, directory_path, headers):
+        self.send_status_and_headers(200, headers)
+        for element in os.listdir(os.path.join(os.getcwd(), directory_path)):
+            self.client_socket.send(
+                "<a href=%s> %s </a> %s <br>" % (
```
kaustubh-karkare: Use proper `<ul>`, `<li>`, `<a>` tags. Put the property values inside quotes. No need for <br> or extraneous spacing.

solutions/handleclient.py
```diff
@@ -0,0 +1,116 @@
...
+                    (
+                        os.path.join(
+                            directory_path.replace(' ', '%20'),
+                            element.replace(' ', '%20')
```
kaustubh-karkare: There are methods that do this stuff for you. "url encoding" it is called. Use them.

solutions/server.py
```diff
@@ -0,0 +1,35 @@
+import os
+import socket
+
+from handleclient import HandleClient
+
+
+class Server:
+
+    def __init__(self, host, port, document_root):
+        os.chdir(document_root)
```
kaustubh-karkare: If you're changing directories at the start of the program, you should change back at the end of it.

solutions/server.py
```diff
@@ -0,0 +1,37 @@
...
+        self.server_socket.listen(5)
+        while self.server_listening:
+            try:
+                client_socket, client_address = self.server_socket.accept()
+                HandleClient(client_socket, client_address).start()
+            except socket.timeout:
+                pass
+            except KeyboardInterrupt:
+                self.stop_serving()
+        os.chdir(original_path)
```
kaustubh-karkare: In case of exceptions, this might not be invoked. Which is why you should prefer using try-finally blocks.

solutions/server.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def stop_serving(self):
+        self.server_listening = False
```
kaustubh-karkare: Since you've used this model before, for the sake of learning, use threading.Event.

solutions/handleclient.py
```diff
@@ -0,0 +1,133 @@
...
+        400: 'Bad Request',
+        404: 'Not Found',
+        408: 'Request Timeout',
+        501: 'Not Implemented',
+    }
+    pattern = r'([A-Z]+)\s+/*(\S*)\s+(HTTP/\d+\.\d+)'
+
+    def __init__(self, client_socket, client_address, **kwargs):
+        self.client_socket = client_socket
+        request_timeout = kwargs.get('request_timeout', 5)
```
kaustubh-karkare: You're not actually providing this.

solutions/handleclient.py
```diff
@@ -0,0 +1,133 @@
...
+                    self.send_directory_content(
+                        client_request['path'],
+                        response_headers,
+                    )
+        else:
+            self.send_status_and_headers(501, response_headers,)
+
+    def recieve_request(self):
+        client_request = dict()
+        headers = dict()
```
kaustubh-karkare: s/dict()/{}

solutions/handleclient.py
```diff
@@ -0,0 +1,133 @@
...
+        data = ''
+        while EOL not in data:
+            try:
+                data += self.client_socket.recv(BLOCK_SIZE)
+            except socket.timeout:
+                self.request_timed_out = True
```
kaustubh-karkare: Instead of state variables, throw exceptions, which can be caught from your run method.

solutions/handleclient.py
```diff
@@ -0,0 +1,133 @@
...
+                return client_request, headers
+        request_line, _, data = data.partition(EOL)
+        try:
+            client_request['method'], client_request['path'], \
+                client_request['version'] = \
```
kaustubh-karkare: Everything on one, or 1 item per line.

solutions/handleclient.py
```diff
@@ -0,0 +1,133 @@
...
+                    os.path.join(requested_path, 'index.html')
+                if os.path.isfile(default_path):
+                    self.send_file(default_path, response_headers)
+                else:
+                    self.send_directory_content(
+                        client_request['path'],
+                        response_headers,
+                    )
+        else:
+            self.send_status_and_headers(501, response_headers,)
```
kaustubh-karkare: Return early when possible to avoid indenting.

solutions/handleclient.py
```diff
@@ -0,0 +1,133 @@
...
+            not os.path.isfile(requested_path) and
+            not os.path.isdir(requested_path)
+        ):
+            self.send_status_and_headers(404, response_headers)
+        elif client_request.get('method') == 'GET':
+            if os.path.isfile(requested_path):
+                self.send_file(requested_path, response_headers)
+            else:
+                default_path = \
+                    os.path.join(requested_path, 'index.html')
```
kaustubh-karkare: `index.html` should probably be a constant.

solutions/handleclient.py
```diff
@@ -0,0 +1,133 @@
...
+                data = requested_file.read(BLOCK_SIZE)
+                if not data:
+                    break
+                self.client_socket.send(data)
+
+    def send_directory_content(self, directory_path, headers):
+        self.send_status_and_headers(200, headers)
+        self.client_socket.send('<html><body><ul>')
+        for element in os \
+                .listdir(os.path.join(os.getcwd(), directory_path)):
```
kaustubh-karkare: Create an `abs_path` variable so that you dont need to break this line.

solutions/handleclient.py
```diff
@@ -0,0 +1,133 @@
...
+            self.client_socket.send(
+                '<li><a href=' +
+                os.path.join(
+                    urllib.quote(directory_path),
+                    urllib.quote(element)
+                ) +
+                '>' +
```
kaustubh-karkare:
```
'<li><a href="' + urllib.quote(os.path.join(directory_path, element)) + '">'

kaustubh-karkare:
Inactive for too long.

PR #38, opened 2017-05-07
assignment-3.2

humanoid2795:


solutions/assignment-3.2/cppcompileutility.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def get_cpp_files_to_compile(self, directory):
+        valid_extensions = \
+            ['.C', '.cc', '.cpp', '.CPP', '.c++', '.cp', '.cxx']
+        files_to_compile = []
+        if not os.path.isfile(self.file_list_path):
+            file_list = open(self.file_list_path, 'w+')
+        else:
+            file_list = open(self.file_list_path, 'r+')
+        for file in os.listdir(directory):
```
kaustubh-karkare: Hmm. "file" is an existing type in python, which is why this is colored differently.
Lets use "file_name", which is more accurate.

solutions/assignment-3.2/cppcompileutility.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def __init__(self, directory):
+        self.file_list_path = os.path.join(directory, 'file_list')
+        files_to_compile = self.get_cpp_files_to_compile(directory)
+        self.update_file_list(directory)
+        self.compile_files(files_to_compile)
+
+    def get_cpp_files_to_compile(self, directory):
+        valid_extensions = \
+            ['.C', '.cc', '.cpp', '.CPP', '.c++', '.cp', '.cxx']
```
kaustubh-karkare: Does C++ force you to only ever use these extensions? Is it not possible to use anything else?

solutions/assignment-3.2/cppcompileutility.py
```diff
@@ -0,0 +1,83 @@
...
+        files_to_compile = []
+        if not os.path.isfile(self.file_list_path):
+            file_list = open(self.file_list_path, 'w+')
+        else:
+            file_list = open(self.file_list_path, 'r+')
```
kaustubh-karkare: I'm confused about this condition. Why do you need to open using "w+" ?

solutions/assignment-3.2/cppcompileutility.py
```diff
@@ -0,0 +1,83 @@
...
+        for file in os.listdir(directory):
+            extension = None
+            if file.rfind('.') != -1:
```
kaustubh-karkare: Prefer standard methods like "os.path.splitext"

solutions/assignment-3.2/cppcompileutility.py
```diff
@@ -0,0 +1,83 @@
...
+                extension = file[file.rfind('.'):]
+            else:
+                continue
+            is_acceptable_extension = False
+            for available_extension in valid_extensions:
+                if extension == available_extension:
+                    is_acceptable_extension = True
+                    break
+            if not is_acceptable_extension:
+                continue
```
kaustubh-karkare: I think the last 7 lines could be shortened to:
```
if extension not in valid_extensions:
  continue

File: solutions/assignment-3.2/cppcompileutility.py [hunk 6]
```diff
@@ -0,0 +1,83 @@
+import os
+import json
+
+
+class CompileCppFiles:
+
+    def __init__(self, directory):
+        self.file_list_path = os.path.join(directory, 'file_list')
```
kaustubh-karkare: Umm ... this wasn't exactly what I had in mind. This information should be stored in memory only.
Right now, it looks like I'd need to call your program each time after I update files, which will then possibly invoke the compiler. But that would be just as bad as having to compile it manually. Instead, I'd like to be able to start this program, and leave it running the background, and it automatically detects/compiles/tests on its own.

solutions/assignment-3.2/cppcompileutility.py
```diff
@@ -0,0 +1,83 @@
...
+            is_acceptable_extension = False
+            for available_extension in valid_extensions:
+                if extension == available_extension:
+                    is_acceptable_extension = True
+                    break
+            if not is_acceptable_extension:
+                continue
+            file_list.seek(0)
+            try:
+                decoded_file_list = json.load(file_list)
```
kaustubh-karkare: See previous comment. Your input isn't an existing file_list, it is the actual file system.

solutions/assignment-3.2/cppcompileutility.py
```diff
@@ -0,0 +1,83 @@
...
+                        break
+                if not file_exists_in_directory:
+                    del decoded_file_list[key]
+            file_list.seek(0)
+            file_list.truncate()
+            json.dump(decoded_file_list, file_list)
+
+    def compile_files(self, files_to_compile):
+        for file in files_to_compile:
+            os.system('g++ ' + file)
```
kaustubh-karkare: Umm ... it looks like each file is a distinct program? Can we use a better example? One which involves multiple files building a single final executable? Feel free to create a "test" directory or something showing me the files you're working with.

solutions/assignment-3.2/assignment-3.2.py
```diff
@@ -0,0 +1,34 @@
...
+import os
+import time
+
+from make import Make
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--file_name',
```
kaustubh-karkare: Lets stick to using dashes here. "--file-name". But is "--config-file" more accurate?

solutions/assignment-3.2/assignment-3.2.py
```diff
@@ -0,0 +1,34 @@
...
+    parser.add_argument(
+        '--directory',
+        '-d',
+        required=False,
+        type=str,
+        default=os.getcwd(),
+        help='Directory of Descriptor file'
+    )
+    parsed_argument = parser.parse_args()
+    descriptor_path = os.path.join(
```
kaustubh-karkare: Hmm. It seems odd that you have to provide the directory separately. Why not `--file-name=/path/to/.file` ?

solutions/assignment-3.2/make.py
```diff
@@ -0,0 +1,83 @@
...
+            args=()
+        ).start()
+        threading.Thread(
+            target=self.restart_utility,
+            args=(descriptor_stored_timestamp,)
+        ).start()
+
+    def validate_and_update_target(self, target, descriptor_list):
+        target_element = None
+        for element in descriptor_list:
```
kaustubh-karkare: Umm ... can you use better names here? It isn't obvious what descriptor or target means.

solutions/assignment-3.2/make.py
```diff
@@ -0,0 +1,83 @@
...
+        command_list = target_element['command']
+        target_path = os.path.join(os.path.dirname(self.path), target)
+        execute_commands = False
+        if os.path.isfile(target_path):
+            target_modified_timestamp = os.path.getmtime(target_path)
+        else:
+            target_modified_timestamp = 0
+            execute_commands = True
+        for element in dependency_list:
+            self.validate_and_update_target(element, descriptor_list)
```
kaustubh-karkare: Ah ... you're recursing into itself. I'm not sure that is needed though.

solutions/assignment-3.2/make.py
```diff
@@ -0,0 +1,83 @@
...
+        self.utility_to_run = False
+
+    def restart_utility(self, descriptor_stored_timestamp):
+        while self.utility_to_run:
+            descriptor_current_timestamp = \
+                os.path.getmtime(self.path)
+            if descriptor_current_timestamp - descriptor_stored_timestamp > 0:
+                descriptor_stored_timestamp = descriptor_current_timestamp
+                self.stop_utility()
+                self.start_utility()
```
kaustubh-karkare: Eh ... this is a rather convoluted way to basically write a loop. Simplify.

solutions/assignment-3.2/buildutility.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def run(self):
+        self.start_buildutility()
+
+    def start_buildutility(self):
+        interval_between_checks = 1
+        self.buildutility_is_alive = True
+        while self.buildutility_is_alive:
+            self.path_thread_is_alive = True
+            with ThreadPoolExecutor(max_workers=len(glob(self.watch)) + 1) \
```
kaustubh-karkare: Why does each path have its own thread?

solutions/assignment-3.2/buildutility.py
```diff
@@ -0,0 +1,78 @@
...
+                    as executor:
+                executor.submit(
+                    self.watch_paths_for_updation,
+                    os.path.dirname(self.watch),
+                    interval_between_checks
+                )
+                for element in glob(self.watch):
```
kaustubh-karkare: This is the 2nd time you're calling `glob(self.watch)`. Use results from first time?

solutions/assignment-3.2/buildutility.py
```diff
@@ -0,0 +1,78 @@
...
+import time
+
+from glob import glob
+from threading import Thread
+from concurrent.futures import ThreadPoolExecutor
+
+
+class BuildUtility(Thread):
+
+    def __init__(self, watch, action):
```
kaustubh-karkare: Use a better name that "watch"?

solutions/assignment-3.2/buildutility.py
```diff
@@ -0,0 +1,78 @@
...
+import logging
+import subprocess
+import time
+
+from glob import glob
+from threading import Thread
+from concurrent.futures import ThreadPoolExecutor
+
+
+class BuildUtility(Thread):
```
kaustubh-karkare: Watcher is probably a more accurate name :P

solutions/assignment-3.2/buildutility.py
```diff
@@ -0,0 +1,78 @@
...
+                self.path_thread_is_alive = False
+                break
+            time.sleep(interval_between_checks)
+
+    def build_files(self):
+        file_list = ''
+        for element in glob(self.watch):
+            file_list = file_list + element + ' '
+        subprocess.call(
+            self.action.replace(self.watch, file_list),
```
kaustubh-karkare: You shouldn't need to do this replacement.

solutions/assignment-3.2/buildutility.py
```diff
@@ -0,0 +1,78 @@
...
+            shell=True
+            )
```
kaustubh-karkare: Improper indenting.

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,39 @@
...
+
+def main():
+    logging.info('Logger for continuous_build_system')
+    logger = logging.getLogger('contiuous_build_system' + str(time.ctime()))
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--configfile',
+        '-f',
+        required=False,
+        default=os.path.join(os.getcwd(), '.configfile'),
```
kaustubh-karkare: configfile for what? You need a more useful filename, like ".bashrc" or ".gitconfig", etc.

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,39 @@
...
+        type=str,
+        help='Config file'
+    )
+    parsed_argument = parser.parse_args()
+    try:
+        with open(parsed_argument.configfile, 'r') as configfile:
+            try:
+                input_data = json.load(configfile)
+                watch = input_data["watch"]
```
kaustubh-karkare: Use .get, so that things are optional.

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,39 @@
...
+                action = input_data["action"]
+            except ValueError:
+                logger.error('Key("watch"/"action") is missing')
+                raise
+    except IOError:
+        logger.error('Config file not available')
+        raise
+    BuildUtility(watch, action).start()
+
```
kaustubh-karkare: What is the method to stop this?

solutions/assignment-3.2/test/buildutilitytest.py
```diff
@@ -0,0 +1,79 @@
+import os
+import unittest
+import time
+
+from calendar import timegm
+from tempfile import NamedTemporaryFile
+from buildutility import BuildUtility
+
+WATCH = "/tmp/test/*.cpp"
```
kaustubh-karkare: Use tempfile.mkdtemp. Otherwise I cant run 2 copies of this test in parallel.

solutions/assignment-3.2/buildutility.py
```diff
@@ -0,0 +1,78 @@
...
+                self.build_files()
+            time.sleep(interval_between_checks)
+
+    def watch_paths_for_updation(self, file_path, interval_between_checks):
+        try:
+            stored_modified_time = os.path.getmtime(file_path)
+        except IOError:
+            self.logger.warning('Path - ' + file_path + ' Not Available')
+            return
+        while self.path_thread_is_alive:
```
kaustubh-karkare: This variable doesn't feel right to me.
If the path thread throws an error, the build utility should stop, and accessible through a getError() method or something.

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,44 @@
...
+import json
+import logging
+import os
+import time
+
+from watcher import Watcher
+
+
+def main():
+    logging.info('Logger for continuous_build_system(cbs)')
```
kaustubh-karkare: What is this for? Remove?

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,44 @@
...
+    logger = logging.getLogger('contiuous_build_system' + str(time.ctime()))
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--cbsconfig',
+        '-f',
```
kaustubh-karkare: You want short names based on long ones. `-c` for config might be better.

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,44 @@
...
+        default=os.path.join(os.getcwd(), 'cbsconfig'),
+        type=str,
+        help='Continuous build system(cbs) config file'
+    )
+    parsed_argument = parser.parse_args()
+    try:
+        with open(parsed_argument.cbsconfig, 'r') as cbsconfig:
+            try:
+                input_data = json.load(cbsconfig)
+                paths_to_watch = input_data.get("watch")
```
kaustubh-karkare: `get` takes a 2nd argument, to use as the default if the key doesn't exist.
```
input_data.get("watch", os.getcwd())

File: solutions/assignment-3.2/continuous_build_system.py [hunk 28]
```diff
@@ -0,0 +1,44 @@
...
+                logger.error('Key("action") is missing')
+                raise
+    except IOError:
+        logger.error('Continuous build system config file not available')
+        raise
+    if paths_to_watch is None:
+        paths_to_watch = os.getcwd()
+    path_watcher = Watcher(paths_to_watch, action_to_execute)
+    path_watcher.start()
+    time.sleep(30)
```
kaustubh-karkare: Does something like this work?
```
try:
  while True:
    time.sleep(1)
except KeyboardInterrupt:
  pass

File: solutions/assignment-3.2/watcher.py [hunk 29]
```diff
@@ -0,0 +1,53 @@
...
+from threading import Thread
+
+
+class Watcher(Thread):
+
+    def __init__(
+            self,
+            paths_to_watch,
+            action_to_execute,
+            interval_between_checks=1
```
kaustubh-karkare: Overindented? Also, trailing comma.

solutions/assignment-3.2/watcher.py
```diff
@@ -0,0 +1,53 @@
...
+    ):
+        logging.info('Logger for  Watcher')
```
kaustubh-karkare: You already have a logger, do you need this?

solutions/assignment-3.2/watcher.py
```diff
@@ -0,0 +1,53 @@
...
+        self.action_to_execute = action_to_execute
+        self.interval_between_checks = interval_between_checks
+        self.watching = False
+        Thread.__init__(self)
+
+    def run(self):
+        previous_paths = {}
+        for element in glob(self.paths_to_watch):
+            try:
+                previous_paths[element] = os.path.getmtime(element)
```
kaustubh-karkare: `previous_paths` is inaccurate. `previous_modified_time` perhaps.

solutions/assignment-3.2/watcher.py
```diff
@@ -0,0 +1,53 @@
...
+            except IOError:
+                self.logger.info('Path - ' + element + 'Not Available')
+        self.watching = True
+        while self.watching:
+            self.start_watcher(previous_paths)
```
kaustubh-karkare: You're not starting it. You're running it.

solutions/assignment-3.2/watcher.py
```diff
@@ -0,0 +1,53 @@
...
+            time.sleep(self.interval_between_checks)
+
+    def start_watcher(self, previous_paths):
+        current_paths = {}
+        for element in glob(self.paths_to_watch):
+            try:
+                current_paths[element] = os.path.getmtime(element)
+            except IOError:
+                self.logger.info('Path - ' + element + 'Not Available')
+        if set(previous_paths.items()) != set(current_paths.items()):
```
kaustubh-karkare: This seems unnecessary. You could just compare the modified time for each file, and break out of the loop as soon as possible.

solutions/assignment-3.2/watcher.py
```diff
@@ -0,0 +1,66 @@
...
+            else:
+                action_executed = True
+                subprocess.call(self.action_to_execute, shell=True)
+                break
+        # Check if any previous path is deleted.
+        if not action_executed:
+            if previous_modified_time:
+                subprocess.call(self.action_to_execute, shell=True)
+        previous_modified_time.clear()
+        previous_modified_time.update(current_modified_time)
```
kaustubh-karkare: This works, but it might not be intuitive. In general, it isn't a good idea to modify arguments.
How about self.previous_modified_times = current_modified_times

solutions/assignment-3.2/watcher.py
```diff
@@ -0,0 +1,66 @@
...
+    ):
+        self.paths_to_watch = paths_to_watch
+        self.action_to_execute = action_to_execute
+        self.interval_between_checks = interval_between_checks
+        self.watching = False
+        self.logger = logger
+        Thread.__init__(self)
+
+    def run(self):
+        previous_modified_time = {}
```
kaustubh-karkare: times

solutions/assignment-3.2/watcher.py
```diff
@@ -0,0 +1,66 @@
...
+                current_modified_time[element] = os.path.getmtime(element)
+            except IOError:
+                self.logger.info('Path - ' + element + 'Not Available')
+        action_executed = False
+        # Check if new path is added or updated.
+        for current_path, current_time in current_modified_time.items():
+            if current_path in previous_modified_time:
+                if current_time - previous_modified_time[current_path] > 0:
+                    action_executed = True
+                    subprocess.call(self.action_to_execute, shell=True)
```
kaustubh-karkare: This repetition is inelegant. Just set `should_execute_action = True` and do this once later.

solutions/assignment-3.2/watcher.py
```diff
@@ -0,0 +1,66 @@
...
+                    break
+                else:
+                    del previous_modified_time[current_path]
+            else:
+                action_executed = True
+                subprocess.call(self.action_to_execute, shell=True)
+                break
+        # Check if any previous path is deleted.
+        if not action_executed:
+            if previous_modified_time:
```
kaustubh-karkare: Isn't this always true? I expected to see a loop here.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,100 @@
...
+from watcher import Watcher
+
+DIRECTORY = tempfile.mkdtemp(dir='/tmp/')
+WATCH = DIRECTORY + "/*.cpp"
+ACTION = "g++ -o " + DIRECTORY + "/output.exe " + DIRECTORY + "/*.cpp"
+FILECOUNT = 0
+
+
+def empty_cppfile():
+    global FILECOUNT
```
kaustubh-karkare: Eh ... in general, lets avoid mutable global variables. There's a better way to do this.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,100 @@
...
+    FILECOUNT = FILECOUNT + 1
+    temp_file = tempfile.NamedTemporaryFile(dir=DIRECTORY, delete=False)
+    new_file_name = \
+        os.path.join(DIRECTORY, 'testfile' + str(FILECOUNT) + '.cpp')
+    old_file_name = os.path.join(DIRECTORY, temp_file.name)
+    os.rename(old_file_name, new_file_name)
```
kaustubh-karkare: Why do you need to name it at all? This could simplify a bunch of code.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,100 @@
...
+    return new_file_name
+
+
+def create_cppfile():
+    temp_file = empty_cppfile()
+    with open(os.path.join(DIRECTORY, temp_file), 'w') as cppfile:
+        cppfile.write('#include <bits/stdc++.h>\n')
+        cppfile.write('using namespace std;\n')
+        cppfile.write('int main() {\n')
+        cppfile.write('return 0; }')
```
kaustubh-karkare: Use multiline strings.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class TestWatcher(unittest.TestCase):
+
+    def test_creatingfile(self):
+        path_watcher = \
+            Watcher(WATCH, ACTION)
```
kaustubh-karkare: Single line.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,100 @@
...
+        output_path = os.path.join(DIRECTORY, 'output.exe')
+        try:
+            if os.path.exists(output_path):
```
kaustubh-karkare: Make your program add 2 numbers, and then check the result.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,100 @@
...
+                os.remove(output_path)
+        except IOError:
+            pass
+        previous_block = {}
+        for element in glob.glob(WATCH):
+            previous_block[element] = os.path.getmtime(element)
+        temp_file = empty_cppfile()
+        path_watcher.start_watcher(previous_block)
+        self.assertTrue(os.path.exists(output_path))
+        os.remove(os.path.join(DIRECTORY, temp_file))
```
kaustubh-karkare: Use `with` statements to seamlessly create/delete these. Right now, in case of exceptions, this line might not be called.

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def main():
+    logger = logging.getLogger('contiuous_build_system ' + str(time.ctime()))
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--cbsconfig',
+        '-c',
+        required=False,
+        default=os.path.join(os.getcwd(), '.cbsconfig'),
```
kaustubh-karkare: This is new. Why cant you use a relative path here?

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,59 @@
...
+        with open(parsed_argument.cbsconfig, 'r') as cbsconfig:
+            try:
+                input_data = json.load(cbsconfig)
+                paths_to_watch = input_data.get("watch", os.getcwd())
+                action_to_execute = input_data["action"]
+            except ValueError:
+                logger.error('Key("action") is missing')
+                raise
+    except IOError:
+        logger.error('Continuous build system config file not available')
```
kaustubh-karkare: Hmm. You could avoid this nesting, if you return early if os.path.isFile is false or something.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,43 @@
+import os
+import tempfile
+import unittest
+
+from watcher import Watcher
+
+DIRECTORY = tempfile.mkdtemp(dir='/tmp/')
+OUTPUT_DIRECTORY = tempfile.mkdtemp(dir='/tmp/')
```
kaustubh-karkare: I'm pretty sure you dont need to provide the directory here.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,43 @@
...
+OUTPUT_PATH = os.path.join(OUTPUT_DIRECTORY, 'out.txt')
+WATCH = DIRECTORY
```
kaustubh-karkare: I think I had a different comment about using a better name for this.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,43 @@
...
+ACTION = "echo True> " + os.path.join(OUTPUT_DIRECTORY, "out.txt")
+
+
+class TestWatcher(unittest.TestCase):
+
+    def test_creatingpath(self):
+        path_watcher = Watcher(WATCH, ACTION)
```
kaustubh-karkare: > Why dont you need to call `.start()` here?

Update: As discussed on messenger, need a negative test case here.

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,58 @@
...
+import json
+import logging
+import os
+import time
+
+from watcher import Watcher
+
+
+def main():
+    logger = logging.getLogger('contiuous_build_system ' + str(time.ctime()))
```
kaustubh-karkare: Not asking you to change, but just curious ... what is the value of the time here?

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,58 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--cbsconfig',
+        '-c',
+        required=False,
+        default=os.path.join(os.getcwd(), 'cbsconfig'),
```
kaustubh-karkare: Replacing `default=os.path.join(os.getcwd(), 'cbsconfig')` with `default='cbsconfig'` makes no difference, right?

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,58 @@
...
+        '-i',
+        required=False,
+        default=1,
+        type=float,
+        help='Interval between consecutive checks',
+    )
+    parsed_argument = parser.parse_args()
+    if not os.path.isfile(parsed_argument.cbsconfig):
+        logger.error('Config file is not available')
+        return
```
kaustubh-karkare: `sys.exit(1)` ?

solutions/assignment-3.2/continuous_build_system.py
```diff
@@ -0,0 +1,58 @@
...
+    with open(parsed_argument.cbsconfig, 'r') as cbsconfig:
+        try:
+            input_data = json.load(cbsconfig)
+            paths_to_watch = input_data.get("watch", os.getcwd())
+            action_to_execute = input_data["action"]
```
kaustubh-karkare: You can take line 35-36 out of try-catch. And then use an if condition to replace it, which would be cheaper.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,54 @@
+import os
+import tempfile
+import unittest
+
+from watcher import Watcher
+
+DIRECTORY = tempfile.mkdtemp(dir='/tmp/')
```
kaustubh-karkare: Call this TEMP_DIRECTORY ?

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,54 @@
...
+OUTPUT_PATH = tempfile.mkstemp(dir=DIRECTORY)
+PATHS_TO_WATCH = DIRECTORY
+ACTION_TO_EXECUTE = "echo True> " + OUTPUT_PATH[1]
```
kaustubh-karkare: Why the `[1]`?

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,54 @@
+import os
+import tempfile
+import unittest
+
+from watcher import Watcher
+
+DIRECTORY = tempfile.mkdtemp(dir='/tmp/')
+OUTPUT_PATH = tempfile.mkstemp(dir=DIRECTORY)
```
kaustubh-karkare: Looks like this doesn't change across test cases. And so the file created in 1 test will be read from another test. You'd see the problem if you add a test cases where nothing changes, and so the file should never be created.

solutions/assignment-3.2/watcher.py
```diff
@@ -0,0 +1,67 @@
...
+            time.sleep(self.interval_between_checks)
+
+    def run_watcher(self):
+        current_modified_times = {}
+        for element in glob(self.paths_to_watch):
+            try:
+                current_modified_times[element] = os.path.getmtime(element)
+            except IOError:
+                self.logger.info('Path - ' + element + 'Not Available')
+        should_execute_action = False
```
kaustubh-karkare: Put an empty line before this. Makes it easier to read.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,81 @@
...
+        yield testing_path
+    finally:
+        os.chdir(original_path)
+        shutil.rmtree(testing_path)
+
+
+class TestWatcher(unittest.TestCase):
+
+    def test_creatingpath(self):
+        for testing_path in temporary_directory():
```
kaustubh-karkare: Not sure if you actually understood what context manager does, but you should be able to do "with temporary_directory():" instead of this.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,81 @@
...
+            output_path = 'out.txt'
+            paths_to_watch = testing_path
+            action_to_execute = "echo True> " + output_path
+            path_watcher = Watcher(paths_to_watch, action_to_execute)
+            time.sleep(0.1)
+            with tempfile.TemporaryFile(dir=testing_path) as _:
```
kaustubh-karkare: This doesn't need to be a `with` statement. Since shutils deletes the entire directory anyway, you can just create it.

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,81 @@
...
+
+class TestWatcher(unittest.TestCase):
+
+    def test_creatingpath(self):
+        for testing_path in temporary_directory():
+            output_path = 'out.txt'
+            paths_to_watch = testing_path
+            action_to_execute = "echo True> " + output_path
+            path_watcher = Watcher(paths_to_watch, action_to_execute)
+            time.sleep(0.1)
```
kaustubh-karkare: Why?

solutions/assignment-3.2/test/watchertest.py
```diff
@@ -0,0 +1,81 @@
...
+            temp_file = \
+                tempfile.NamedTemporaryFile(dir=testing_path, delete=False)
+            path_watcher = Watcher(paths_to_watch, action_to_execute)
+            current_modified_time = os.path.getmtime(temp_file.name)
+            os.utime(
+                temp_file.name,
+                (current_modified_time + 1, current_modified_time + 1),
+            )
+            path_watcher.run_watcher()
+            self.assertFalse(os.path.exists(output_path))
```
kaustubh-karkare: Why not?

PR #35, opened 2017-04-03
Solution 2.2 Added

sciencepal:
Please Review @kaustubh-karkare

PR #39, opened 2017-05-14
Solution 2.2 Added

sciencepal:
Hi Sir,
I have updated my previous code for assignment 2.2. I created a new pull request because I messed up while committing changes to the [old one](https://github.com/kaustubh-karkare/project-omega/pull/35). Following the review of the previous code, the new code has the following changes : 
1. The permissions of all files have been reverted to their original form.
2. Used re module instead of urlparse.
3. Function names and variable names have been made specific. 
4. Dictionary for header string has been created.
5. New code abides by the style guide (hopefully).
6. Exceptions have been raised instead of abrupt exit calls.
7. Tempfile module has been used.
8. Command Line Input has been separated from download logic.


solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+import socket
+import tempfile
+import logging
+import sys
+import os
+
+
+def get_headers(path, host, req_type, start_byte, end_byte):
+    header = ""
+    dict = {}
```
kaustubh-karkare: Lets not reuse identifiers which already have a different meaning.

solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+    dict["endline"] = "\r\n"
```
kaustubh-karkare: This should be a top-level constant.

solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+    dict["request"] = "%s %s HTTP/1.1" % (req_type, path)
+    dict["host"] = "Host: %s" % (host)
```
kaustubh-karkare: Do something like:
```
headers["Host"] = host
...
for key, value in headers.items():
  headers += "%s: %s%s" % (key, value, EOL)

File: solutions/aditya-pal/assignment2.2/download.py [hunk 4]
```diff
@@ -0,0 +1,148 @@
...
+            "GET",
+            self.start_byte,
+            self.end_byte
+        )
+        client_socket.send(headers)
+        received_data = ""
+        while True:
+            file_data = client_socket.recv(1024)
+            if not file_data:
+                break
```
kaustubh-karkare: You're assuming here that the entire file data will be stored in memory (in your `received_data` method). What if we're dealing with files as large as GBs, where this isn't a good idea? Assume that headers fit in memory, but flush every subsequent block as you get it.

solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+            received_data += file_data
+        file = tempfile.NamedTemporaryFile(delete=False)
+        self.files.append(file.name)
```
kaustubh-karkare: So this file contains the http headers too?
Also, is this threadsafe? Since different instance can finish at different times, the ordering is here is undefined.

solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+import os
+
+
+def get_headers(path, host, req_type, start_byte, end_byte):
+    header = ""
+    dict = {}
+    dict["endline"] = "\r\n"
+    dict["request"] = "%s %s HTTP/1.1" % (req_type, path)
+    dict["host"] = "Host: %s" % (host)
+    dict["range"] = "Range: bytes=%s-%s" % (str(start_byte), str(end_byte))
```
kaustubh-karkare: use %d so that you dont need to cast to string

solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+        self.thread_num = thread_num
+        self.files = files
+
+    def run(self):
+        client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        try:
+            client_socket.connect((self.host, self.port))
+        except socket.error:
+            logging.info("Could Not Establish Connection, Download Failed")
+            raise Exception("Error : Could Not Connect To Download Link")
```
kaustubh-karkare: A link is a string. The problem is that you "could not connect to server"

solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+        self.files = []
+
+    def download(self):
+        client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        try:
+            client_socket.connect((self.host, self.port))
+        except socket.error:
+            logging.info("Could Not Establish Connection, Download Failed")
+            raise Exception("Error : Could Not Connect To Download Link")
+        header_request = get_headers(self.path, self.host, "HEAD", 0, 0)
```
kaustubh-karkare: Since I've already seen the implementation above, I understand that the 0s are the start_index and end_index. But if I hadn't, it isn't obvious from this function call what they are. Either use defaults, or specify the argument names, but always try and make it obvious what it each argument means, without looking at the function implementation.

solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+        client_socket.send(header_request)
+        response_data = ""
+        while True:
+            response = client_socket.recv(1024)
+            if not response:
+                break
+            response_data += response
+        header_data = response_data
+        header_len = len(header_data)
+        first_index = response_data.find("Content-Length")
```
kaustubh-karkare: Dont do this kind of string manipulation, since it is extremely fragile. Create a headers dict like your request, then access the value corresponding to the "Content-Length" key.

solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+        response_data.strip()
+        content_length = 0
+        try:
+            content_length = int(response_data)
+            if content_length <= 0:
+                logging.info("No Meta Data")
+                raise Exception("Error : No Meta Data")
+        except ValueError:
+            logging.info("Incorrect Meta Data")
+            raise Exception("Error : Incorrect Meta Data")
```
kaustubh-karkare: It isn't incorrect. It is optional for the server to provide this header, and in that case, you should just download with 1 thread.

solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+            thread.join()
+        outfile = os.path.basename(os.path.basename(self.path))
+        output_file = open(outfile, 'w')
+        logging.info("Writing To File ...")
+        data = ""
+        for num in range(thread_num):
+            file = open(self.files[num])
+            file.seek(0)
+            for line in file:
+                data += line
```
kaustubh-karkare: Once again, you cant load all this data into memory.

solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+            file.close()
+            os.unlink(self.files[num])
+        x = (header_data[-8:])
+        p = data.rfind(x, header_len + 1)
```
kaustubh-karkare: How about you do this in each thread, before writing to tempfiles?

solutions/aditya-pal/assignment2.2/download.py
```diff
@@ -0,0 +1,148 @@
...
+        output_file.write(data[p + 8:])
+        logging.info("Download Completed Successfully !!!")
+
+
+def inputs(hostname, path, port, thread_count):
```
kaustubh-karkare: This doesn't seem particularly functional. Remove?

solutions/aditya-pal/assignment2.2/main.py
```diff
@@ -0,0 +1,34 @@
...
+        help="URL of File to be Downloaded"
+    )
+    arguments.add_argument(
+        "-t",
+        "--threads",
+        default=3,
+        required=False,
+        help="Number of Threads in which File will be downloaded"
+    )
+    arguments.add_argument(
```
kaustubh-karkare: Don't need this, port information is part of the URL.

solutions/aditya-pal/assignment2.2/main.py
```diff
@@ -0,0 +1,34 @@
...
+        required=False,
+        help="Port Number for file download"
+    )
+    args = arguments.parse_args()
+    url_elements = re.split('//', args.url, 1)[-1]
+    hostname = re.split('/', url_elements, 1)[0]
+    path = re.split('/', url_elements, 1)[-1]
+    path = '/' + path
+    thread_count = int(args.threads)
+    port = int(args.port)
```
kaustubh-karkare: This is not taking full advantage of regex. Use `re.match` to extract host, port, path in a single line.

kaustubh-karkare:
Inactive for too long.

PR #42, opened 2017-06-02
regular-expression-engine

humanoid2795:


solutions/regex.py
```diff
@@ -0,0 +1,273 @@
...
+    elif element == 's':
+        ascii_range_limits.append([ord(' '), ord(' ')])
+    elif element == 'w':
+        ascii_range_limits.append([ord('0'), ord('9')])
+        ascii_range_limits.append([ord('A'), ord('Z')])
+        ascii_range_limits.append([ord('a'), ord('z')])
+        ascii_range_limits.append([ord('_'), ord('_')])
+    elif element == 'D':
+        ascii_range_limits.append([0, ord('0') - 1])
+        ascii_range_limits.append([ord('9') + 1, 127])
```
kaustubh-karkare: Interesting. I might have used a 2-item tuple instead of a list here.

Instead of a list of ranges, how about an object that contains them, so that you can just set a "negative" flag so that you dont need to explicitly provide these inverse ranges?

solutions/regex.py
```diff
@@ -0,0 +1,273 @@
...
+    elif element == 'W':
+        ascii_range_limits.append([0, ord(' ') - 1])
+        ascii_range_limits.append([ord(' ') + 1, ord('0') - 1])
+        ascii_range_limits.append([ord('9') + 1, ord('A') - 1])
+        ascii_range_limits.append([ord('Z') + 1, ord('_') - 1])
+        ascii_range_limits.append([ord('_') + 1, ord('a') - 1])
+        ascii_range_limits.append([ord('z') + 1, 127])
+    else:
+        logging.error('Special character is invalid')
+        sys.exit(1)
```
kaustubh-karkare: This is almost always not what you want. Raise exceptions so that the users of your library can handle it differently.

solutions/regex.py
```diff
@@ -0,0 +1,273 @@
...
+    valid_ascii_codes = list()
+    for ascii_range in ascii_range_limits:
+        while ascii_range[0] <= ascii_range[1]:
+            valid_ascii_codes.append(ascii_range[0])
+            ascii_range[0] += 1
```
kaustubh-karkare: Why not compare limits directly. Less time/memory required.

solutions/regex.py
```diff
@@ -0,0 +1,273 @@
...
+import string
+import sys
+
+SPECIAL_CHARACTERS = ['d', 's', 'w', 'D', 'S', 'W']
+MAXIMUM_REPETITIONS = 1000
+
+
+def special_character_ascii_range(element):
+    ascii_range_limits = []
+    if element == 'd':
```
kaustubh-karkare: There is duplication here. You'd specified this "d" both here, and on line 5. Can you avoid it?

solutions/regex.py
```diff
@@ -0,0 +1,273 @@
...
+    index += 1
+    try:
+        if pattern[index] in SPECIAL_CHARACTERS:
+            valid_ascii_range = special_character_ascii_range(pattern[index])
+        else:
+            # Escape character will be treated as a literal
+            valid_ascii_range.append(ord(pattern[index]))
+    except IndexError:
+        logging.error('Escape character is missing')
+        raise
```
kaustubh-karkare: This will raise the index error again, which might not be what you want. Raise a more useful one?

solutions/regex.py
```diff
@@ -0,0 +1,273 @@
...
+    index += 1
+    return valid_ascii_range, index
+
+
+def parse_character_class(elements, index):
```
kaustubh-karkare: All this should be part of a "RegexParser" class. That way, you dont need to pass the index around.

solutions/regex.py
```diff
@@ -0,0 +1,273 @@
...
+            if match_text(current_state.next_state, text, index):
+                return True
+        else:
+            return match_text(current_state.next_state, text, index)
+    return False
+
+
+def check(pattern, text):
+    match_start_from_beginning = False
+    pattern = pattern.rstrip('$')
```
kaustubh-karkare: What about `match_to_ending` (based on your variable name below)?

solutions/regex.py
```diff
@@ -0,0 +1,273 @@
...
+            ord(text[index]) in current_state.valid_paths
+        ):
+            index += 1
+        else:
+            return False
+
+    if match_text(current_state.next_state, text, index):
+        return True
+
+    for ii in range(int(lower_limit), int(upper_limit)):
```
kaustubh-karkare: Interesting.

solutions/regex.py
```diff
@@ -0,0 +1,273 @@
...
+            for ii in range(ord('\r') + 1, 128):
+                self.valid_paths.append(ii)
+            index += 1
+        else:
+            self.valid_paths.append(ord(pattern[index]))
+            index += 1
+
+        if index < len(pattern):
+            # Check for repetition_qualifier
+            if pattern[index] == '?':
```
kaustubh-karkare: This stuff looks like it should part of the parsing phase, not matching phase.

solutions/tokenize.py
```diff
@@ -0,0 +1,99 @@
...
+                    last_index = last_index + 1
+                else:
+                    break
+            tokens += (('character-class', data[first_index:last_index]),)
+            ii = last_index
+
+        elif data[ii] == '{':
+            # Check if repitition range given or to treat as literal.
+            first_index = ii + 1
+            last_index = data.find('}', first_index)
```
kaustubh-karkare: Options:
1. What you're doing here: looking for the ending first, and then processing the contents.
2. Processing the contents as you increase the index, until you happen to reach the ending.

I'm not sure which is better.

solutions/tokenize.py
```diff
@@ -0,0 +1,99 @@
+import string
+
+UPPER_LIMIT = 1000
+
+
+def tokenizer(data):
+    tokens = ()
+    data_length = len(data)
+    ii = 0
+    while ii < data_length:
```
kaustubh-karkare: Alright, so the model you have here is much better, but instead of 1 large function like this, I'd suggest breaking down each case into a function that you can call. That way, you can recursively call things (like parse escape sequences inside character ranges, or character ranges inside groups, etc).

solutions/regex.py
```diff
@@ -0,0 +1,257 @@
...
+    def __init__(self):
+
+        self.state_type = 'initial-state'
+        self.next_state = None
+
+    def match(self, text, index):
+        return self.next_state.match(text, index)
+
+
+class StartAnchor:
```
kaustubh-karkare: You should make all of these extend a base class, with abstract methods (ie - `raise NotImplementedError()`) so that the API is known beforehand and consistent.

I say this because it is confusing why CharRange doesn't share the same methods as the other classes.

solutions/regex.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def match(self, text, index):
+        return self.next_state.match(text, index) if index == 0 else False
+
+
+class CharRange:
+
+    def __init__(self, char_ranges, acceptable_characters=True):
+        self.char_ranges = char_ranges
+        self.acceptable_characters = acceptable_characters
```
kaustubh-karkare: This is a confusing name. `inverse_match` might be better.

solutions/regex.py
```diff
@@ -0,0 +1,257 @@
...
+        self.next_state = None
+
+    def match(self, text, index):
+        return self.next_state.match(text, index)
+
+
+class StartAnchor:
+
+    def __init__(self):
+        self.state_type = 'start-anchor'
```
kaustubh-karkare: Use the enum I mention in the other file.

solutions/regex.py
```diff
@@ -0,0 +1,257 @@
...
+class ConsumeCharacter:
+
+    def __init__(
+        self,
+        char_range_items,
+        minimum_characters_to_consume=1,
+        maximum_characters_to_consume=1
+    ):
+        self.state_type = 'consume-character'
+        self.char_range_items = char_range_items
```
kaustubh-karkare: You shouldn't assume that the only thing here is a char range. It can be other groups too.
Example: `(a[bd]c){2}` matches `adcabc`

solutions/regex.py
```diff
@@ -0,0 +1,257 @@
...
+        self.maximum_characters_to_consume = maximum_characters_to_consume
+        self.next_state = None
+
+    def check_if_acceptable_character(self, character):
+        for char_range_item in self.char_range_items:
+            if char_range_item.check_if_character_in_range(character):
+                return True
+        return False
+
+    def match(self, text, index):
```
kaustubh-karkare: Nice!

Interesting sidenote: your implementation here tried to match as many characters as possible. Ie - it is greedy. So `/(a+)(a+)/.match("aaaaa")` will cause the first group to have 4 characters, and the 2nd one to match only 1 character.

In real regexes, you can make it non-greedy by adding a `?` after the `?`/`+`/`*`. So `/(a+?)(a+?)/.match("aaaaa")` will make the first group match only 1, and the 2nd one is forced to take 4 characters.

solutions/regex.py
```diff
@@ -0,0 +1,257 @@
+from tokenize import tokenizer
+
```
kaustubh-karkare: You have a file from step 1: tokenizer.
Create a separate file for step 2: parser, which takes in the list of tokens, and returns the linked list.
And finally a 3rd file which serves as the API (ie - puts the regex though the first 2 phases, and then calls the match.

solutions/regex.py
```diff
@@ -0,0 +1,257 @@
...
+                maximum_characters_to_consume
+            )
+        ii += 1
+        previous_state = previous_state.next_state
+    previous_state.next_state = FinalState()
+    return initial_state
+
+
+def check(pattern, text):
+    initial_state = compile(pattern)
```
kaustubh-karkare: This isn't state though. This is just a node in your list.

solutions/regex.py
```diff
@@ -0,0 +1,257 @@
...
+    if initial_state.match(text, 0):
+        return True
+    else:
+        for ii in range(1, len(text)):
```
kaustubh-karkare: If you start this loop from 0, you can get rid of the if condition.

PR #46, opened 2017-11-12
Command line options parser

Shivam-sopho:
Main library in - library.py
Driver Program = test.py


solutions/assignment1/library.py
```diff
@@ -0,0 +1,144 @@
+import re
```
kaustubh-karkare: Lets start using Python3, so that you can use specify typehints.

solutions/assignment1/test.py
```diff
@@ -0,0 +1,9 @@
+import library
+import sys
+if __name__ == '__main__':
+    printer = library.assignment1()
```
kaustubh-karkare: The filename / class name should be based on what it contains. Instead of "assignment1", call it Parser.

solutions/assignment1/test.py
```diff
@@ -0,0 +1,9 @@
+import library
```
kaustubh-karkare, line 1: Write a proper test using the unittest module provided by Python.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def __init__(self):
+       # print("class created")
+        self.copyargs = {}
+        self.requiredargs = {}
+        self.nonrequiredargs = {}
+        self.checking = "noerror"
+        self.userpassedargs = {}
+        self.userpassrequiredargs = {}
+        self.userpassnonrequiredargs = {}
```
kaustubh-karkare: Saving the results in the same class makes the class impossible to reuse.
You should split up the results from the parser.

solutions/assignment1/test.py
```diff
@@ -0,0 +1,9 @@
...
+import sys
+if __name__ == '__main__':
+    printer = library.assignment1()
+    ans = printer.createoptions(
+        key=(1234, "r", "local"), name=("shivam", "nr", "local"))
```
kaustubh-karkare: This API is definitely needs improvement.
Check out how the argparse module works: https://docs.python.org/3/library/argparse.html
What you're trying to do is to create your own simplified version of that.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,144 @@
...
+        Returns:
+            A dictionary of containing Command Line Options
+
+        """
+        # Making copy of required and nonrequiredargs
+        self.copyargs = wargs
+        for commands in self.copyargs:
+            temp = self.copyargs[commands][1]
+            if temp == "r":
+                self.requiredargs[commands] = self.copyargs[commands]
```
kaustubh-karkare: `self._required_args`

Check out the python style guide for the conventions you need to abide by.
The style guide is mentioned in the instructions for solving the problems here.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,144 @@
...
+        if self.checking == "noerror":
+            print(self.userpassedargs)
+
+    def errorlisting(self, errorcategory, command):
+        """ Function for generating the list of errors
+
+        """
+        print("Error: ")
+        if errorcategory == "mandatoryerror":
+            print("The \"--%s\" argument is required, but missing from input." % command)
```
kaustubh-karkare: Error messages usually go to stderr, not stdout.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,144 @@
...
+                self.errorlisting(self.checking, commands)
+                exit()
+        for commands in self.userpassedargs:
+            if commands not in self.copyargs:
+                self.checking = "overflow"
+                self.errorlisting(self.checking, commands)
+                exit()
+
+       # Checking for type for the argument values
+        for commands in self.userpassedargs:
```
kaustubh-karkare: Inconsistent indenting. Another thing covered by style guide.

kaustubh-karkare (changes requested):
- You can use something like https://www.pylint.org/
- The existing comments will result in significant refactoring, so I didn't review in much detail the code inside the Parser file.

Shivam-sopho:
@kaustubh-karkare . Taking your suggestions into consideration I have made the changes . 
1. For the python code formatting I am using pep8 in place of pylint
2. I have tried to setup the Unit Test using Unittest module of python

solutions/assignment1/test.py
```diff
@@ -0,0 +1,35 @@
+import library
+import unittest
+import sys
+
+
+class simple_test(unittest.TestCase):
```
kaustubh-karkare: Class names are typically camelCase, so SimpleTest

solutions/assignment1/test.py
```diff
@@ -0,0 +1,35 @@
...
+    def setUp(self):
+        self.printer = library.parser()
```
kaustubh-karkare: The names here are inconsistent.
Why is a "parser" stored in a variable called "printer"? It'll confuse people.
See point 3 of the [coding instructions](https://github.com/kaustubh-karkare/project-omega/blob/master/assignments/README.md).

solutions/assignment1/test.py
```diff
@@ -0,0 +1,35 @@
...
+        self.printer.add_options(key=(1234, "r", "local"))
+        self.printer.add_options(name=("Shivam", "nr", "local"))
+        self.commands = ['--key=123', '--name=s']
+        self.test = self.printer.parsecommands(self.commands)
+        pass
+
+    def test_print(self):
+        self.assertEqual(self.test, {'key': 123, 'name': 's'})
+
+    def test_required_args(self):
```
kaustubh-karkare, line 23: These tests are not comprehensive enough yet.
Ie - in order to test_required_args, you need to prove that an exception is raised when the required argument is not provided.

solutions/assignment1/test.py
```diff
@@ -0,0 +1,35 @@
...
+        self.assertEqual(self.test, {'key': 123, 'name': 's'})
+
+    def test_typeof_args(self):
+        self.commands = ['--key=123.7', '--name=s']
+        self.assertEqual(self.printer.parsecommands(
+            self.commands), {'key': 123, 'name': 's'})
```
kaustubh-karkare: Similarly, need to add case for what happens when the expected type (eg - number) != actual type (string like "abc").

solutions/assignment1/test.py
```diff
@@ -0,0 +1,35 @@
...
+
+if __name__ == '__main__':
+    unittest.main()
+    """
+    printer = library.parser()
+    printer.add_options(key=(1234, "r", "local"))
+    printer.add_options(name=("shivam", "nr", "local"))
+    commands = ['--key=123', '--name=s']
+    printer.parsecommands(commands)
+    """
```
kaustubh-karkare: Remove all necessary code.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,209 @@
...
+    def __init__(self):
+        # print("class created")
+        self.copyargs = {}
+        self.checking = "noerror"
+        self.userpassedargs = {}
+        self.type = ""
+        self.local = []
+        self.record = store_args_options()
+
+    def add_options(self, **wargs):
```
kaustubh-karkare: This variable is conventionally called "kwargs" for "KeyWord ARGumentS".

solutions/assignment1/library.py
```diff
@@ -0,0 +1,209 @@
...
+        """ Function for creating the command line options
+
+        Args:
+            **wargs : Keyword Arguments (List of Command line options passed)
+        Returns:
```
kaustubh-karkare: This method does not have a return statement.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,209 @@
...
+class parser:
+
+    def __init__(self):
+        # print("class created")
+        self.copyargs = {}
+        self.checking = "noerror"
+        self.userpassedargs = {}
+        self.type = ""
+        self.local = []
+        self.record = store_args_options()
```
kaustubh-karkare: Ah, so this separate class contains the options, both the expected and the value values.
I think you may have misunderstood me: we want a Parser class object which consists of a list of expected values, and then a separate object that contains actual values. Eg - 

```
parser = Parser()
```

solutions/assignment1/library.py
```diff
@@ -0,0 +1,209 @@
...
+                self.checking = "typeerror"
+                print(self.record.get_all_args()[commands])
+                self.type = typeuserdefined
+                self.errorlisting(self.checking, commands)
+                return self.checking
+
+        for commands in self.userpassedargs:
+            self.local.append(self.record.get_all_args()[commands][2])
+
+        if "local" in self.local and "remote" in self.local:
```
kaustubh-karkare: Hardcoding specific values is not allowed!
You must make the configurable as part of your API.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,209 @@
...
+        # User passed non required arguments and required argumengs
+        for commands in self.userpassedargs:
+            if commands in self.record.get_required_args():
+                self.record.add_user_required_args(
+                    commands, self.userpassedargs[commands])
+            else:
+                self.record.add_user_nonrequired_args(
+                    commands, self.userpassedargs)
+
+        return self.validateargs()
```
kaustubh-karkare: Why not do the validation while storing the values? It doesn't need to be a separate phase.

kaustubh-karkare (changes requested):
Okay, this is better, but still needs work.

solutions/assignment1/library.py
Shivam-sopho: If Validation is done in separate phase I thought it may make program more modular. Please let me know if I am heading in wrong direction.

kaustubh-karkare: Ah, a reasonable instinct. The reason I suggested doing both things together is that you kinda have a "single module of everything related to one value" in one place. Having separate phases add a bit more overhead, so having modules of "all actions on one value" is simpler.

solutions/assignment1/library.py
Shivam-sopho: One thing I have doubt is that In the assignment you have mentioned about local and remote . What does that actually mean.

kaustubh-karkare: They're just 2 random options that cant be used together.
Eg - you can either perform the action locally, or remotely.

Shivam-sopho: Every time the user passes the command line arguments for parsing, will that we every time treated as new or the last passed arguments will be retained with it ? Suppose the example in which we have called the parser.parse option 2 times.

kaustubh-karkare: Each time you call parse, it's a new thing.
You don't have to store values across calls.

Shivam-sopho:
1. I have taken your suggestions into consideration and made the changes.
2. Changes the API Structure to match the Python Arg Parser Module
3. Refined the UnitTests

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
+from re import search
+from sys import stderr
+
+
+class StoreArgsOptions:
```
kaustubh-karkare: Huh, not sure why you need this class at all, just looks like a straightforward dict.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+class Parser:
+
+    def __init__(self):
+        # print("class created")
+        self._expected_args = {}
+        self._required_args = {}
+        self.checking = "noerror"
+        self.type = ""
+        self.modes = []
+        self.record = StoreArgsOptions()
```
kaustubh-karkare: We should not store the results of parsing in the class.
Given the input, we can just parse it, and return the results immediately.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def add_option(self, *args, **kwargs):
+        """ Function for adding expected arguments info
+
+        Args:
+            **kwargs : Keyword Arguments with expected arguments information
+        """
+
+        for arg in args:
+            self._expected_args[arg] = kwargs
```
kaustubh-karkare: Interesting, this requires the "application developer" to understand internal details about your library, which is not good. I would recommend explicitly calling our the valid args/kwargs in your signature, so that the signature is enough to understand what is going on. And only allow 1 option to be added at a time.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+        Returns:
+            The User passed values if everything goes well , otherwise returns the error
+        """
+        # Storing required Args for Error Checking
+        for key, value in self._expected_args.items():
+            for i in value:
+                if i == 'required':
+                    self._required_args[key] = value
+
+        for keys in argslist:
```
kaustubh-karkare: s/keys/key, since it is one item.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+            keys = keys.split("--")
```
kaustubh-karkare:
```
assert key.startswith("--")
name, value = key[2:].partition("=")
```

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+            keys[1] = keys[1].split('=')
+            self.record.add_user_passed_arg(keys[1][0], keys[1][1])
+
+        # Checking error for mandatory and non-mandatory
+
+        for commands in self._required_args:
+            if commands not in self.record.get_user_passed_arg():
+                print(self.record.get_user_passed_arg())
```
kaustubh-karkare: Libraries aren't allowed to print stuff unless explicitly allowed :)

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+                self.checking = "mandatoryerror"
+                self.errorlisting(self.checking, commands)
+                return self.checking
+
+        for commands in self.record.get_user_passed_arg():
```
kaustubh-karkare: Instead of doing each type of check in phases, you can try to do all the checks for a given argument at a time.
Eg - after your splitting operation, you can immediately verify that it is expected, and then do the type checking etc, before adding it to the result dictionary. That way, you avoid all the extra loops :)

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+            if commands not in self._expected_args:
+                self.checking = "overflow"
+                self.errorlisting(self.checking, commands)
+                return self.checking
+
+        for commands in self.record.get_user_passed_arg():
+            try:
```
kaustubh-karkare: Looks like you're just trying to fit it into whatever type it goes in, which is non-ideal.
The specific type should be configurable (check out how the argparse module does it).

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+                            self.record.get_user_passed_arg()[commands] = str(
+                                self.record.get_user_passed_arg()[commands])
+                    except ValueError:
+                        return self.checking
+
+        # Type Checking(datatype checking)
+        for commands in self.record.get_user_passed_arg():
+
+            typeuserpassed = type(self.record.get_user_passed_arg()[commands])
+            typeuserdefined = self._expected_args[commands]['type']
```
kaustubh-karkare: I'm confused what this does, if you don't use it in the previous section.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+            if(typeuserdefined != typeuserpassed):
+                self.checking = "typeerror"
+                self.type = typeuserdefined
+                self.errorlisting(self.checking, commands)
+                return self.checking
+
+        for commands in self.record.get_user_passed_arg():
```
kaustubh-karkare: s/commands/command/

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+            self.modes.append(self._expected_args[commands]['mode'])
+
+        if "local" in self.modes and "remote" in self.modes:
```
kaustubh-karkare: Once again, this is hardcoding specific values in your code.
Ie - what if I wanted 5 different pairs of arguments to be mutually exclusive? This currently model would not support that.

How about making mode an arbitrary string, and you just ensure that the corresponding count is not more than one? Also, "mode" isn't the best name, maybe "unique_key" ?

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+            self.checking = "conflict"
+            self.errorlisting(self.checking, "both")
+            return self.checking
+
+        if self.checking == "noerror":
+            user_passed_args = self.record.get_user_passed_arg()
+            self.record.empty_user_passed_arg()
```
kaustubh-karkare: Given that you reset it at the end anyway, there is no reason to save it on the class in the first place.
It creates an unnecessary restriction that the parser cannot be used across multiple threads.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,154 @@
...
+            return user_passed_args
+        else:
+            return self.checking
+
+    def errorlisting(self, errorcategory, command):
```
kaustubh-karkare: This is just a library that will be called by a larger application, so you should raise exceptions instead. The caller can then choose to display the errors, or do something else.

I'd also recommend not having a separate method to generate error messages, just create the message from the place where you'd call this. That way, all the relevant logic is in one place.

solutions/assignment1/test.py
```diff
@@ -0,0 +1,51 @@
...
+    def test_print(self):
+        self.commands = ['--key=123', '--key2=1234.6', '--name1=Shivam']
+        self.assertEqual(self.parser.parse(self.commands),
+                         {'key': 123,  'key2': 1234.6, 'name1': 'Shivam'})
+
+    def test_required_args(self):
+        self.commands = ['--key=123']
+        self.assertEqual(self.parser.parse(self.commands), {'key': 123})
+        self.commands = ['--key2=123.6']
+        self.assertEqual(self.parser.parse(self.commands), 'mandatoryerror')
```
kaustubh-karkare: Right now, your function can either return a valid result, or this error string.
This means that all callsites have to check the 2 cases, which is non-ideal.
Throwing exceptions means that whoever is interested can catch them.

solutions/assignment1/test.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def test_overflow_args(self):
```
kaustubh-karkare: I think you mean "unexpected" instead of "overflow"?

kaustubh-karkare (changes requested):
Much better! Some more refactoring needed, but glad to see the progress!

Shivam-sopho:
As per suggestions I have made changes in the implementation
1. Make parsing in one pass instead of various passes.
2. Usage of Exceptions
3. Not storing the result of parsing directly returning it.
4. Make corrections in mutually exclusive keys logic.

solutions/assignment1/library.py
kaustubh-karkare: This is still an issue.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,103 @@
...
+        unique_key_nums = 0
+        present_unique_key = ""
+        previous_unique_key = ""
+        user_passed_args = {}
+
+        for keys in argslist:
+            if keys.startswith('--'):
+                x = keys[2:].partition('=')
+                name = x[0]
+                value = x[2]
```
kaustubh-karkare: name, _separator, value = keys[2:].partition('=')

solutions/assignment1/library.py
```diff
@@ -0,0 +1,103 @@
...
+            argslist : List of commands given by user
+
+        Returns:
+            The User passed values if everything goes well , otherwise returns the error
+        """
+        # Storing required Args for Error Checking
+        for keys, value in self._expected_args.items():
+            for i in value:
+                if i == 'required':
+                    self._required_args[keys] = value
```
kaustubh-karkare: You can do this in one line:
```
required_args = set([name for name, value in self._expected_args.items() if "required" in value])
```
Also note that `required_args` is a local variable, it does not need to be a class property, since it is never accessed outside this method.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,103 @@
...
+        user_passed_args = {}
+
+        for keys in argslist:
+            if keys.startswith('--'):
+                x = keys[2:].partition('=')
+                name = x[0]
+                value = x[2]
+                if name not in self._expected_args:
+                    try:
+                        raise ValueError("unexpected")
```
kaustubh-karkare: You can create your own exception types (ParserError) instead of existing ones (ValueError).
Makes it clearer when you write code for try-catch.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,103 @@
...
+                    except Exception as error:
+                        return str(error)
```
kaustubh-karkare: Wait, you're raising an exception and then immediately catching it? This is equivalent to not raising it at all.
That kinda defeats the point of raising exceptions ... you want to let the application/test code catch it, instead of doing it here.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,103 @@
...
+                try:
+                    value = int(value)
```
kaustubh-karkare: > Looks like you're just trying to fit it into whatever type it goes in, which is non-ideal.
The specific type should be configurable (check out how the argparse module does it).

kaustubh-karkare (changes requested):

Shivam-sopho:
As per suggestions I have made changes in the implementation
1. Improved the Usage of Exceptions
2. Included which options application programmer can pass
3. Replaced codes with one liners .

solutions/assignment1/test.py
```diff
@@ -0,0 +1,73 @@
...
+                         {'key': 123,  'key2': 1234.6, 'name1': 'Shivam'})
+
+    def test_required_args(self):
+        self.commands = ['--key=123']
+        self.assertEqual(self.parser.parse(self.commands), {'key': 123})
+        self.commands = ['--key2=123.6']
+        try:
+            self.parser.parse(self.commands)
+        except Exception as error:
+            self.assertEqual(str(error), 'mandatoryerror')
```
kaustubh-karkare: You can do:
```
with self.assertRaisesRegex('mandatoryerror'):
  self.parser.parse(self.commands)
```

solutions/assignment1/test.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class SimpleTest(unittest.TestCase):
+    def setUp(self):
+        self.parser = library.Parser()
+        self.parser.add_option('key', required=True,
+                               type=int, unique_key="local")
+        self.parser.add_option('name',  type=str, unique_key="remote")
+        self.parser.add_option('key2', unique_key="local", type=float)
+        self.parser.add_option('name1', unique_key="local", type=str)
```
kaustubh-karkare: Having to specify a unique_key in every case is inconvenient for users who don't actually care about this.
An easy fix option is to assume a default of None, which means there are no conflicts. That way, you only specify this for the options that care about it.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,95 @@
...
+    def __str__(self):
+        return(str(self.value))
+
+
+class Parser:
+
+    def __init__(self):
+        # print("class created")
+        self._expected_args = {}
+        self._required_args = {}
```
kaustubh-karkare: Not used anymore.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,95 @@
...
+        required_args = set(
+            [name for name, value in self._expected_args.items() if "required" in value])
+
+        unique_key_nums = 0
+        present_unique_key = ""
+        previous_unique_key = ""
+        user_passed_args = {}
+
+        for keys in argslist:
+            if keys.startswith('--'):
```
kaustubh-karkare: If this isn't true, what happens? You might want to raise an exception.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,95 @@
...
+                name, _separator, value = keys[2:].partition('=')
+                if name not in self._expected_args:
+                    raise ParserError("unexpected")
```
kaustubh-karkare: Okay, so you're using a custom exception, which is progress! You might also want to give detailed error messages in the exceptions, so that it is easy to debug. "unexpected option: {name}"

Shivam-sopho:
As per your suggestions I have added the changes . Please have a look and let me know if any new changes are required.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,93 @@
...
+                    try:
+                        value = int(value)
+                    except ValueError:
+                        raise ParserError("typeerror")
+                if typeuserdefined == float:
+                    try:
+                        value = float(value)
+                    except ValueError:
+                        raise ParserError("typeerror")
+                typeuserpassed = type(value)
```
kaustubh-karkare: You're still special casing `int` and `float` here.
```
value = self._expected_args[name]['type'](value) # raises ValueError if needed
```

solutions/assignment1/library.py
```diff
@@ -0,0 +1,93 @@
...
+                user_passed_args[name] = value
+
+                if(typeuserdefined != typeuserpassed):
+                    self.checking = "typeerror"
+                    raise ParserError("typeerror")
+
+                if 'unique_key' in self._expected_args[name].keys():
+                    previous_unique_key = present_unique_key
```
kaustubh-karkare: This logic depends on the specific ordering, which means that you can circumvent it easily, if you have multiple unique_key values.

solutions/assignment1/library.py
kaustubh-karkare: I would still highly recommend not using *args & **kwargs unless you're forwarding arguments, since it requires the "application developer" to understand internal details about your library.
```
def add_option(self, name, *, required=true, type=str, unique_key=None):
```
The above will prevent the application developer from making mistakes.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,93 @@
...
+                if previous_unique_key != "" and previous_unique_key != present_unique_key:
+                    raise ParserError("conflict")
+            else:
+                raise ParserError("Unexpected Format")
+        for commands in required_args:
+            if commands not in user_passed_args:
+                raise ParserError("mandatoryerror")
+        return user_passed_args
+
+    def isstring(self, string):
```
kaustubh-karkare: This will not exist in the final form, since this is also a form of special casing for strings, which is not allowed.

kaustubh-karkare (changes requested):

Shivam-sopho:
I have added the changes as per your suggestions. Please have a look on them.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,85 @@
...
+                        uniquekey_count = uniquekey_count + 1
+
+                if uniquekey_count > 1:
+                    raise ParserError("conflict")
+            else:
+                raise ParserError("Unexpected Format")
+
+        for commands in required_args:
+            if commands not in user_passed_args:
+                raise ParserError("mandatoryerror")
```
kaustubh-karkare: I would recommend more detailed error messages instead of a single word.
"Missing required argument: {name}"

solutions/assignment1/library.py
```diff
@@ -0,0 +1,85 @@
...
+                    else:
+                        conflict_checker[uniquekey] = 1
+                        uniquekey_count = uniquekey_count + 1
+
+                if uniquekey_count > 1:
+                    raise ParserError("conflict")
+            else:
+                raise ParserError("Unexpected Format")
+
+        for commands in required_args:
```
kaustubh-karkare: `for name in required_arg_names`

As usual, we should maintain a high bar for names of variables.

solutions/assignment1/library.py
```diff
@@ -0,0 +1,85 @@
...
+
+                if uniquekey is not None:
+                    if uniquekey in conflict_checker:
+                        conflict_checker[uniquekey] = 2
+                    else:
+                        conflict_checker[uniquekey] = 1
+                        uniquekey_count = uniquekey_count + 1
+
+                if uniquekey_count > 1:
+                    raise ParserError("conflict")
```
kaustubh-karkare: I don't think this logic is correct.
If any conflict_checker[uniquekey] == 2, then you have to raise an exception.
uniquekey_count > 1 is allowed, if you're using 2 separate uniquekeys

kaustubh-karkare (approved):
I think at this point, you've completed the goals I had in mind.
There's still some items left to fix in the code, so check out the comments but once you've fixed it, you can close the PR.

Shivam-sopho:
Made the above changes .

PR #47, opened 2019-08-15
Add submission for Argument Parsing

ankurdubey521:
Taking your feedback into consideration, I'm using the Google Styling guide and ESLint for checking code conventions

### DIFF
--- a/solutions/ankurdubey521/Argument Parser/parser.js
+++ b/solutions/ankurdubey521/Argument Parser/parser.js
@@ -1,40 +1,6 @@
 /**
- * Exception Class
- * @param {string} message
+ * Argument Definition Class
  */
-function InvalidArgumentException(message) {
-  this.message = message;
-}
-/**
- * Exception Class
- * @param {string} message
- */
-function InvalidArgumentTypeException(message) {
-  this.message = message;
-}
-/**
- * Exception Class
- * @param {string} message
- */
-function MalformedArgumentException(message) {
-  this.message = message;
-}
-/**
- * Exception Class
- * @param {string} message
- */
-function MissingRequiredArgumentException(message) {
-  this.message = message;
-}
-/**
- * Exception Class
- * @param {string} message
- */
-function MutuallyExclusiveArgumentsPassedException(message) {
-  this.message = message;
-}
-
-// eslint-disable-next-line require-jsdoc
 class Argument {
   /**
    * @constructor
@@ -56,6 +22,43 @@ class Argument {
 };
 
 module.exports = class Parser {
+
+  /**
+   * Exception Class
+   * @param {string} message
+   */
+  InvalidArgumentException(message) {
+    this.message = message;
+  }
+  /**
+   * Exception Class
+   * @param {string} message
+   */
+  InvalidArgumentTypeException(message) {
+    this.message = message;
+  }
+  /**
+   * Exception Class
+   * @param {string} message
+   */
+  MalformedArgumentException(message) {
+    this.message = message;
+  }
+  /**
+   * Exception Class
+   * @param {string} message
+   */
+  MissingRequiredArgumentException(message) {
+    this.message = message;
+  }
+  /**
+   * Exception Class
+   * @param {string} message
+   */
+  MutuallyExclusiveArgumentsPassedException(message) {
+    this.message = message;
+  }
+
   /**
    * @constructor
    */
@@ -181,11 +184,11 @@ module.exports = class Parser {
         arg.replace('-', '').split('').forEach((smallArg) => {
           const index = this.argnameIndexMap[smallArg];
           if (index === undefined) {
-            throw new InvalidArgumentException(
+            throw new this.InvalidArgumentException(
                 'Error: "' + arg + '" is not a valid argument');
           }
           if (!this.indexedArgs[index]['validator'](value)) {
-            throw new InvalidArgumentTypeException(
+            throw new this.InvalidArgumentTypeException(
                 'Error: The value for the "-' + smallArg.toString() +
                 '" argument must be a ' + this.indexedArgs[index]['type'] +
                 '.');
@@ -197,11 +200,11 @@ module.exports = class Parser {
         const largeArg = arg.replace('--', '');
         const index = this.argnameIndexMap[largeArg];
         if (index === undefined) {
-          throw new InvalidArgumentException(
+          throw new this.InvalidArgumentException(
               'Error: "' + arg + '" is not a valid argument');
         }
         if (!this.indexedArgs[index]['validator'](value)) {
-          throw new InvalidArgumentTypeException(
+          throw new this.InvalidArgumentTypeException(
               'Error: The value for the "-' + largeArg.toString() +
               '" argument must be a ' + this.indexedArgs[index]['type'] +
               '.');
@@ -209,7 +212,7 @@ module.exports = class Parser {
         argValues[this.indexedArgs[index]['largeArg']] = value;
       } else {
         // Malformed Argument
-        throw new MalformedArgumentException(
+        throw new this.MalformedArgumentException(
             'Error: Malformed argument ' + arg);
       }
     });
@@ -225,7 +228,7 @@ module.exports = class Parser {
 
   /**
    * Sets arguments passed to be mutually exclusive
-   * @param {list} argList
+   * @param {list} argList: list of argument names (large version only)
    */
   setMutuallyExclusive(argList) {
     this.exclusiveGroups.push(argList);
@@ -244,7 +247,7 @@ module.exports = class Parser {
       const arg = this.indexedArgs[key];
       if (arg.isRequired && argsValue[arg['largeArg']] === undefined) {
         // Check if required=True arg has been set
-        throw new MissingRequiredArgumentException(
+        throw new this.MissingRequiredArgumentException(
             'Error: The \'--' + arg.largeArg +
             '\' argument is required, but missing from input.');
       }
@@ -262,7 +265,8 @@ module.exports = class Parser {
         }
       });
       if (setArgCount > 1) {
-        throw new MutuallyExclusiveArgumentsPassedException(
+        throw new
+        this.MutuallyExclusiveArgumentsPassedException(
             'Error: The arguments "' + setArgs.toString() + '" cannot be passed together.'
         );
       }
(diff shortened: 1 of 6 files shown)


ankurdubey521:

solutions/ankurdubey521/Argument Parser/parser.js
```diff
@@ -0,0 +1,234 @@
...
+  }
+};
+
+const Parser = require('./parser');
+
+const parser = new Parser();
+parser
+    .option('-k', '--key', 'key value, must be + integer'
+        , undefined, 'positive-integer', true)
+    .option('-n', '--name', '', undefined, 'string', false)
```
kaustubh-karkare: You want to make it easy to read function calls.
In this case, there is no way I can figure out what the `undefined` and `false` mean, without reading your code.
In case of large codebases, I do not even have the time to read your code.

So how about something like:

```
.option({
  smallArg: '-n',
  largeArg: '--name',
  type: 'string'
})
```

Check out [destructured function parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Unpacking_fields_from_objects_passed_as_function_parameter).

solutions/ankurdubey521/Argument Parser/parser.js
```diff
@@ -0,0 +1,234 @@
...
+        this.indexedArgs[this.indexOfNewArg].validator = (arg) => {
+          return typeof (arg) === 'boolean';
+        };
+        break;
+      }
+    }
+
+    // Store in argIndices
+    this.argIndices[smallArg] = this.indexOfNewArg.toString();
+    this.argIndices[largeArg] = this.indexOfNewArg.toString();
```
kaustubh-karkare: The name `argIndices` is now a lie, since the keys are not just indices. Lets fix that :)

https://github.com/kaustubh-karkare/project-omega/blob/master/assignments/README.md

Coding instruction number 3.

solutions/ankurdubey521/Argument Parser/parser.js
```diff
@@ -0,0 +1,234 @@
...
+   * @param {string} type
+   * @param {boolean} isRequired
+   * @return {this}
+   */
+  option(smallArg, largeArg, description,
+      defaultValue, type = 'string', isRequired = false) {
+    smallArg = smallArg.replace('-', ''); // single character version of arg
+    largeArg = largeArg.replace('--', ''); // multi-character version of arg
+
+    this.indexedArgs[this.indexOfNewArg] = {
```
kaustubh-karkare: Prefer using proper class instances instead of maps like this, where the structure is not enforced.

solutions/ankurdubey521/Argument Parser/parser.js
```diff
@@ -0,0 +1,234 @@
...
+          throw new InvalidArgumentException(
+              'Error: "' + arg + '" is not a valid argument');
+        }
+        if (!this.indexedArgs[index]['validator'](value)) {
+          throw new InvalidArgumentTypeException(
+              'Error: The value for the "-' + largeArg.toString() +
+              '" argument must be a ' + this.indexedArgs[index]['type'] +
+              '.');
+        }
+        this.indexedArgs[index]['value'] = value;
```
kaustubh-karkare: Storing the values in the Parser objects is not good, it makes the object non-reusable. What if I wanted to parse multiple sets of `sys.argv`?

solutions/ankurdubey521/Argument Parser/parser.js
```diff
@@ -0,0 +1,234 @@
...
+    .option('-k', '--key', 'key value, must be + integer'
+        , undefined, 'positive-integer', true)
+    .option('-n', '--name', '', undefined, 'string', false)
+    .option('-l', '--local', '', undefined, 'boolean', false)
+    .option('-r', '--remote', '', undefined, 'boolean', false);
+
+try {
+  parser.parse(process.argv);
+  if (parser.opts().local && parser.opts().remote) {
+    console.log(
```
kaustubh-karkare: Oh, and this is cheating :P
You want the parser to support this.
Check out [this example](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group).

solutions/ankurdubey521/Argument Parser/parser.js
ankurdubey521: https://github.com/kaustubh-karkare/project-omega/pull/47/commits/62e7c9f9ccbb7382810acd953cf7a183e0819b56

solutions/ankurdubey521/Argument Parser/parser.js
ankurdubey521: https://github.com/kaustubh-karkare/project-omega/pull/47/commits/81e08b472f0216ba0a884229fa4e92407bb4ae0f

solutions/ankurdubey521/Argument Parser/parser.js
ankurdubey521: https://github.com/kaustubh-karkare/project-omega/pull/47/commits/91f918c036c8b0a80ffdabfa7640da8e445cbdfb

solutions/ankurdubey521/Argument Parser/parser.js
ankurdubey521: https://github.com/kaustubh-karkare/project-omega/pull/47/commits/6a127f034ec0f067d36bb4a573fda97cdc103d35

solutions/ankurdubey521/Argument Parser/parser.js
ankurdubey521: https://github.com/kaustubh-karkare/project-omega/pull/47/commits/dd1776874fe90df6eaccadeb721589301a82cfee

solutions/ankurdubey521/Argument Parser/parser.js
```diff
@@ -0,0 +1,277 @@
...
+    largeArg = largeArg.replace('--', ''); // multi-character version of arg
+
+    this.indexedArgs[this.indexOfNewArg] = new Argument(smallArg, largeArg,
+        defaultValue, description, type, isRequired);
+
+
+    // Set the validator function, which checks if value is of correct type
+    switch (type) {
+      case 'positive-integer': {
+        this.indexedArgs[this.indexOfNewArg].validator = this.isPositiveNumber;
```
kaustubh-karkare: [Is this slow?](https://v8.dev/blog/fast-properties). Quoting the part I'm wondering about:

> Every time a new property is added, the object's HiddenClass is changed.

solutions/ankurdubey521/Argument Parser/parser.js
```diff
@@ -0,0 +1,277 @@
...
+
+  /**
+   * Populates arg values and checks for syntactic and type correctness
+   * @param {object} argList
+   * @return {Object} args
+   */
+  parse(argList) {
+    const argValues = {};
+
+    argList = argList.slice(2); // Remove node and executable name from list
```
kaustubh-karkare: I think this part can be the caller's responsibility, since you could be using this class to parse log files, etc, right?

solutions/ankurdubey521/Argument Parser/parser.js
```diff
@@ -0,0 +1,277 @@
...
+
+      if (this.isSmallArg(arg)) {
+        // Small version of Arg passed
+        arg.replace('-', '').split('').forEach((smallArg) => {
+          const index = this.argnameIndexMap[smallArg];
+          if (index === undefined) {
+            throw new this.InvalidArgumentException(
+                'Error: "' + arg + '" is not a valid argument');
+          }
+          if (!this.indexedArgs[index]['validator'](value)) {
```
kaustubh-karkare: I think this is better for static analysis:
```
this.indexedArgs[index].validator(value)
```

solutions/ankurdubey521/Argument Parser/parser.js
```diff
@@ -0,0 +1,277 @@
...
+    const argValues = {};
+
+    argList = argList.slice(2); // Remove node and executable name from list
+
+    argList.forEach((arg) => {
+      // Check if arg has a value explicitly defined
+      // if yes extract it else interpret as a boolean true
+      let value = '';
+      if (arg.indexOf('=') != -1) {
+        value = arg.slice(arg.indexOf('=') + 1);
```
kaustubh-karkare:
```
[a, b] = "a=b".split("=")
```

solutions/ankurdubey521/Argument Parser/parser.js
```diff
@@ -0,0 +1,277 @@
...
+        arg = arg.slice(0, arg.indexOf('='));
+      } else {
+        value = true;
+      }
```
kaustubh-karkare: Here's a complication to mess with you:
What if we wanted to allow using both `=` and space as a separator between keys and values?
Eg - 
```
--key=value
--key value
-k=v
-k v
``` 
Assume all of the above are valid.

solutions/ankurdubey521/Argument Parser/parser.js
```diff
@@ -0,0 +1,277 @@
...
+      if (arg.isRequired && argsValue[arg['largeArg']] === undefined) {
+        // Check if required=True arg has been set
+        throw new this.MissingRequiredArgumentException(
+            'Error: The \'--' + arg.largeArg +
+            '\' argument is required, but missing from input.');
+      }
+    });
+
+    // Check for mutually exclusive arguments
+    this.exclusiveGroups.forEach((list) => {
```
kaustubh-karkare: Much better.
Check out how [python's argparse does this](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group). What do you think of that approach compared to this one?

solutions/ankurdubey521/Argument Parser/parser.test.js
```diff
@@ -0,0 +1,94 @@
...
+parser.setMutuallyExclusive(['--local', '--remote']);
+
+test('PassKeyNameResultsInOutput', () => {
+  const argv = ['node', 'parser.js', '--key=12345', '--name=kaustubh'];
+  expect(parser.parseOpts(argv)).toEqual(
+      {
+        key: '12345',
+        name: 'kaustubh',
+        local: undefined,
+        remote: undefined,
```
kaustubh-karkare: Hmm, in your place, I wouldn't even set the `local` and `remote` keys in this object.
Why? In almost all cases, it doesn't matter.

```
> a = {b: undefined}
{b: undefined}
> a.b
undefined
> a.c
undefined
> a.hasOwnProperty('b')
true
> a.hasOwnProperty('c')
false
```

solutions/ankurdubey521/Argument Parser/parser.test.js
```diff
@@ -0,0 +1,94 @@
...
+      {
+        key: '12345',
+        name: 'kaustubh',
+        local: true,
+        remote: undefined,
+      }
+  );
+});
+
+test('PassInvalidArgumentResultsException', () => {
```
kaustubh-karkare: This is a useful test name. I didn't get that kind value from previous test names.

solutions/ankurdubey521/Argument Parser/parser.js
ankurdubey521: Compared to mine, Arg-parse allows modification of a previously created group by returning a reference to it. It also allows to enforce that at least a single arg from the group is present, which is something I could implement in future

solutions/ankurdubey521/Argument Parser/parser.js
ankurdubey521: Not sure about this one. How should I gauge the impact, considering this happens only once per program run?

ankurdubey521:
>     * Printing helptext automatically if you input `-h` or `--help`.
What does automatically mean? Should the object write directly to the console? What happens if stdout is not console but some file? Also, you mentioned that the library might be used to parse log files, what happens in that case?

ankurdubey521:
> Positional arguments.
Could you elaborate on this too? I looked around but haven't been able to figure out if a case may arise in which multiple positional arguments are defined.

PR #48, opened 2019-08-19
Solution of assignment 1

Hemant-Chowdhury:
Please review the pull request.
To run the code in command line type:
python Main.py --key=-1234 --name=hjvbs5dv


solutions/assignment-1/Main.py
```diff
@@ -0,0 +1,48 @@
...
+import sys
+import json
+
+commands = {}  # Contains the list of allowed commands
+"""Inserting Commands"""
+"""Command is of the type (command_name, command_description, regex_allowed, type_allowed, no_of_arg, conflicting_commands, required_commands, can_overwrite)"""
+commands['--key'] = Command('--key', 'Stores key value pair', r'\+?\d+', "positive integer", 1, None, None, False)
+commands['--name'] = Command('--name', 'Defines name', r'[a-zA-Z]+', 'alphabetic', 1, None, ['key'], True)
+commands['--local'] = Command('--local', 'Defines id of local machine', r'\w+', 'word', 1, ['remote'], None, False)
+commands['--remote'] = Command('--remote', 'Defines id of remote machine', r'\w+', 'word', 1, ['local'], None, False)
```
kaustubh-karkare: You probably want to specify the argument names, in addition to values, I have no idea what the `1` or `None` or `False` mean, without looking at your implementation, which is a sign that your API isn't great.

https://blog.codinghorror.com/avoiding-booleans/

Yes, the comment just above explains what each argument means, but writing that comment was a cost we don't want. If I created commands from different places, you're forcing me to copy the comment to different places.

You want the code to be self-documenting.

solutions/assignment-1/Main.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+error_message = None
+result_space = {}
+
+for arg in sys.argv[1:]:
```
kaustubh-karkare: The logic that lives here should be in its own class, so that it is re-useable for different inputs.

solutions/assignment-1/Main.py
```diff
@@ -0,0 +1,48 @@
...
+        if commands[cmd].no_of_arg == 1:
+            if len(arg_list) > 2:
+                error_message = f'{arg} has invalid syntax'
+            elif len(arg_list) < 2:
+                error_message = f'{arg} needs one argument which is missing'
+            else:
+                try:
+                    commands[cmd].validate(result_space, arg_list[1:])
+                    result_space[cmd[2:]] = arg_list[1]
+                except Exception as e:
```
kaustubh-karkare: Use a custom exception for validation error, instead of something that catches everything (including things that you don't expect).

solutions/assignment-1/Command.py
```diff
@@ -0,0 +1,46 @@
+import re
+
+
+class Command(object):
```
kaustubh-karkare: You probably could not have known, this concept is called an "Option".
But I like that it is represented by its own class.

solutions/assignment-1/Command.py
```diff
@@ -0,0 +1,46 @@
...
+    """Command contains the features related to particular command. It also has validators to validate conditions while parsing."""
+
+    def __init__(self, name, usage, regex, type_allowed, no_of_arg, conflicting_commands, required_commands, overwrite):
+        self.name = name  # Command name
+        self.usage = usage  # What is the use of this command
+        self.regex = r'^' + regex + r'$'  # Regular expression which can distinguish the correct type of argument
+        self.type_allowed = type_allowed  # Type of arguments allowed for the command
```
kaustubh-karkare: If you have `self.regex`, do you really need this?

solutions/assignment-1/Command.py
```diff
@@ -0,0 +1,46 @@
...
+        self.no_of_arg = no_of_arg  # Number of arguments required for the command
+        self.conflicting_commands = conflicting_commands  # Commands which should not apprear together with self.name
```
kaustubh-karkare: Interesting, so you need to specify in more than one place what the conflicts are.
Ie - you have to mention it in both the --local and --remote command objects.
That isn't ideal.

solutions/assignment-1/Main.py
```diff
@@ -0,0 +1,48 @@
+from Command import Command
+import sys
+import json
+
+commands = {}  # Contains the list of allowed commands
+"""Inserting Commands"""
+"""Command is of the type (command_name, command_description, regex_allowed, type_allowed, no_of_arg, conflicting_commands, required_commands, can_overwrite)"""
```
kaustubh-karkare: I don't quite understand why you need required_commands and can_overwrite. Can you help explain?

solutions/assignment-1/Command.py
Hemant-Chowdhury: By adding a regex we can specify what type of patterns arguments should have. For example in the case of "--key" we only need to allow positive integers so providing inputs like "--key=-123" should raise an exception. If I don't use the pattern I cannot generalize the class Commands because then I am only left with options "int", "str" etc.
In other words, using regex gives me a lot of flexibility.

solutions/assignment-1/Main.py
Hemant-Chowdhury: required_commands is to specify those commands which cannot appear alone(It was just a thought, like cases just opposite to conflicting_commands). For eg, suppose "--name" always requires a "--key" to be meaningful, so in that case, we can specify it in required_commands.

can_overrite is to handle cases like "--key=123 --name=abc --name=ABC", in this case, we can just see if "--name" can overwrite itself, if it is allowed then the input is valid else it is an error.

kaustubh-karkare: Yes, this part I understood. The reason I was confused was because these features weren't strictly necessary for the sample input/output I had given.

solutions/assignment-1/Command.py
kaustubh-karkare: You're only using this as part of the error message. You could show the regex in place of that, right?

Hemant-Chowdhury: Yes, self.type_allowed is used only in the error message in case self.regex does not match with the argument. I can instead use self.regex itself to print the error message, but then I thought that the end-user may not be familiar with regex representation, so using some basic English phrase (which represents the regex) may be helpful. For ex, "--key" only allows positive integers. The error message can be represented as "123aaa is not a positive integer. A positive integer type argument is expected" or "123aaa is not of the pattern "^\+?\d+$". A "^\+?\d+$" type argument is expected". Here the first type of representation is easier to understand. To store that basic English phrase I am using self.type_allowed.

Hemant-Chowdhury:
Please review the changes @kaustubh-karkare

Hemant-Chowdhury:
Please review the changes @kaustubh-karkare

solutions/assignment-1/argparser.py
```diff
@@ -0,0 +1,179 @@
...
+    def validate_mutually_exclusive_group(self, arg_parse_mutex_group: Dict[str, List[str]]) -> None:
+        for identifier in self.mutually_exclusive_group_identifier:
+            if len(arg_parse_mutex_group[identifier]) > 1:
+                raise ParsingError(f'{arg_parse_mutex_group[identifier]} cannot appear together')
+            elif len(arg_parse_mutex_group[identifier]) == 0:
+                mutex_list = []
+                for option in self.mutually_exclusive_group:
+                    if self.mutually_exclusive_group[option] == identifier:
+                        mutex_list.append(option)
+                raise ParsingError(f'Atleast one option among {mutex_list} is expected')
```
kaustubh-karkare: This is an assumption you've made: that at least one of the mutually exclusive options must be provided.
In theory, that should be configurable for the group, right?

solutions/assignment-1/argparser.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def validate_option(self, argument: str, arg_parse_options: dict, arg_parse_mutex_group: dict) -> Option:
```
kaustubh-karkare: IMO this method and the next shouldn't need to be separate methods, you could have just specified the logic inline.
Why? Each time you use a method, as a reader of the code, I have to jump to a different location to understand what happened, which makes things a bit harder to read. In case the method name was good enough that I don't have to look deeper, this is fine, but these method names do not tell me what is happening inside.
Eg - Why does `validate_option` need any arguments other than the option to validate? Why is it taking `arg_parse_mutex_group`?
Sometimes it is possible to rename the method to solve the problem, but in this case, since you're doing a bunch of things, I think it would be easier to just inline the code.
Btw, we cant do this all the time (ie - the original method might be too long, adding more complexity would be a bad idea then), but in this case, my subjective opinion is that inlining makes sense.

solutions/assignment-1/argparser.py
```diff
@@ -0,0 +1,179 @@
...
+    def insert_value(self, option: Option, value: str, arg_parse_options: dict)-> None:
+        value = option.validate_type(value)
+        (arg_parse_options[option.name]).append(value)
+
+    def arg_parse(self, arguments: List[str]) -> str:
+
+        if len(arguments) == 1 and (arguments[0] == '--help'):
+            return self.print_help()
+
+        arg_parse_mutex_group: Dict[str, List[str]] = {}  # It is used to store the options of a mutually exclusive group together in the form {identifier: List[str]}, the size of the List[str] should always be 1 to be valid
```
kaustubh-karkare: I don't think you need this particular map.
Ie - you could do the mutually exclusive validation check entirely based on the next map.

solutions/assignment-1/argparser.py
```diff
@@ -0,0 +1,179 @@
...
+                else:
+                    self.insert_value(option, arg, arg_parse_options)
+
+        self.validate_mutually_exclusive_group(arg_parse_mutex_group)
+
+        for arg in arg_parse_options:
+            self.options[arg].validate_nargs(len(arg_parse_options[arg]))
+            output.update(self.options[arg].output(arg_parse_options[arg]))
+
+        return json.dumps(output)
```
kaustubh-karkare: Translating to string isn't something that the library class should be doing.
Instead, if the application wants to print the output, that is where this happens.

PR #49, opened 2019-08-22
solution added for 1st assignment

AbhiTaker:


solutions/abhitaker/src/argparse/main.go
```diff
@@ -0,0 +1,110 @@
...
+func main() {
+
+	// keystore stores and maps the permissible keys to it's value's property
+	keystore := make(map[string]validate.KeyProperty)
+	// jsonkey stores and maps the key, value pair having correct format
+	jsonkey := make(map[string]string)
+	keystore = AddKey("--id", AddKeyProperty("Integer", true, false), keystore)
+	keystore = AddKey("--name", AddKeyProperty("String", true, false), keystore)
+	keystore = AddKey("--age", AddKeyProperty("Integer", false, false), keystore)
+	keystore = AddKey("--roll", AddKeyProperty("AlphaNumeric", false, false), keystore)
```
kaustubh-karkare: Adding keys needs to be entirely configurable, you're not allowed to assume that these are the only 4.

We still want to be able to do something like this from your tests:
```
* parser = Parser()
* parser.add_option('key', type=int, required=True)
* parser.add_option('name', type=str)
* self.assertEqual(parser.parse(['--key=123', '--name=abc']), {'key': 123, 'name': 'abc'})
* self.assertEqual(parser.parse(['--key=456']), {'key': 456})
```

solutions/abhitaker/src/argparse/main.go
```diff
@@ -0,0 +1,110 @@
...
+
+	args := os.Args[1:]
+	for id := 0; id < len(args); id++ {
+
+		argCheck := validate.FormatCheck(args[id])
+		if argCheck == false {
+			ErrorPrint(args[id], "FormatError")
```
kaustubh-karkare: Not sure if Go has custom exceptions, but that is what I'd use here.
 If not, I'd still avoid using raw strings with significant values, and try to use enums instead.

solutions/abhitaker/src/validate/regexitem_test.go
```diff
@@ -0,0 +1,78 @@
+package validate
+
```
kaustubh-karkare: Interesting, so this file tests the various checks with multiple inputs, which is good.
It is missing a complete integration test that tests the code in your main file.

solutions/abhitaker/src/argparse/main_test.go
```diff
@@ -0,0 +1,64 @@
...
+	parser := new(Parser)
+	parser.keystore = make(map[string]validate.KeyProperty)
+	parser.AddKey("--id", AddKeyProperty("Integer", true, false))
+	parser.AddKey("--name", AddKeyProperty("String", true, false))
+	parser.AddKey("--age", AddKeyProperty("Integer", false, false))
+	parser.AddKey("--roll", AddKeyProperty("AlphaNumeric", false, false))
+
+	testInput := []string{"--id=1", "--name=abhishek", "--roll=be1058416", "--age=21"}
+	testExpect := "{\"age\":\"21\",\"id\":\"1\",\"name\":\"abhishek\",\"roll\":\"be1058416\"}"
+	jsonResult, _ := parser.ParseApp(testInput)
```
kaustubh-karkare: This method should return a Map, not a string.
Btw, note that I started reviewing your code my looking at the API inside your tests, since as a user that part is most important to me.

solutions/abhitaker/src/argparse/main_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"testing"
+	"validate"
+)
+
+func TestParserJSON(t *testing.T) {
+
+	parser := new(Parser)
+	parser.keystore = make(map[string]validate.KeyProperty)
```
kaustubh-karkare: This is interesting, since the caller has to initialize your object.
Is it possible to create a method like `initParser` that returns a parser object, and does all the necessary setup?

solutions/abhitaker/src/argparse/main.go
```diff
@@ -0,0 +1,129 @@
...
+
+	var prop validate.KeyProperty
+	prop.Datatype = datatype
+	prop.Mandatory = mandatory
+	prop.Found = found
+	return prop
+}
+
+// ErrorPrint return the appropriate error
+func ErrorPrint(errorItem, errorType string) error {
```
kaustubh-karkare: Given that each of the errors is generated from exactly one place, why not generate the string + error at that location? That will let you avoid this errorItem variable that contains different things based on context, which can use confusion. Ie - what is the value of going through this method?

solutions/abhitaker/src/validate/regexitem.go
```diff
@@ -0,0 +1,52 @@
...
+	Mandatory, Found bool
+}
+
+// JSONItem is Container for storing JSON item
+type JSONItem struct {
+	key, value string
+}
+
+// FormatCheck performs format check on argument
+func FormatCheck(itemArg string) bool {
```
kaustubh-karkare: It's odd that your format check, and then the split method are not part of a single logical unit, since they're so related.

AbhiTaker:
Changes Made:

1.  Created a new Package **argparse**
2.  **main.go** showcase how one can use the **argparse** library
3.  **ParseApp** API returns the JSON Map and **JSONConvert** API converts map into JSON format.
4.  **InitParser** returns a new instance of Parser after initializing the Parser object
5.  Left ErrorPrint as it is (Personal Design Decision)
6.  Removed **Found** flag from the structure
7.  Improved the format for **AddKey**

solutions/abhitaker/src/argparse/argparse.go
```diff
@@ -0,0 +1,126 @@
...
+	parser.keystore[key] = prop
+}
+
+// AddKeyProperty adds argument properties to KeyProperty container
+func AddKeyProperty(datatype string, mandatory, found bool) validate.KeyProperty {
+
+	var prop validate.KeyProperty
+	prop.Datatype = datatype
+	prop.Mandatory = mandatory
+	prop.Found = found
```
kaustubh-karkare: In one of the previous comments, I said this:

> Btw, note that the found flag is specific to one call to your ParseApp not across multiple calls, so should not be part of your AddKeyProperty structure anyway.

I still think you should do this, because there is no value in storing "found" outside of your "ParseApp" method.

solutions/abhitaker/src/validate/regexitem.go
```diff
@@ -0,0 +1,58 @@
...
+	} else {
+		return "", "", matched
+	}
+}
+
+// ExistenceCheck check if given key is present in argument list
+func ExistenceCheck(itemKey string, keystore map[string]KeyProperty) bool {
+
+	_, found := keystore[itemKey]
+	return found
```
kaustubh-karkare: Btw, you already have a different thing called "found", reusing names like this is usually a recipe for confusion.

solutions/abhitaker/src/validate/regexitem.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// KeyProperty is a Container for storing Permissible Key's property
+type KeyProperty struct {
+	Datatype         string
+	Mandatory, Found bool
+}
+
+// JSONItem is Container for storing JSON item
+type JSONItem struct {
```
kaustubh-karkare: This is unused, right?

solutions/abhitaker/src/validate/regexitem.go
```diff
@@ -0,0 +1,58 @@
...
+
+	switch dataType {
+
+	case "String":
+		matched, _ := regexp.MatchString("^([a-zA-Z]+)$", itemValue)
+		return matched
+
+	case "Integer":
+		matched, _ := regexp.MatchString("^([0-9]+$)", itemValue)
+		return matched
```
kaustubh-karkare: In other languages, I would have cast this to an Int, but that's not possible here :)

PR #50, opened 2019-08-23
Added solution to Command Line Parser with Unit Tests

pranjalrai:
I have added solution for assignment 1 along with unit tests. I have used pylint for uniform syntax. Please suggest changes, if required. Thanks.


solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,100 @@
...
+REMOTE_COMMAND = {
+    'command_name': '--remote',
+    'conflicting_with': LOCAL_COMMAND
+}
+
+COMMANDS['--key'] = KEY_COMMAND
+COMMANDS['--name'] = NAME_COMMAND
+COMMANDS['--local'] = LOCAL_COMMAND
+COMMANDS['--remote'] = REMOTE_COMMAND
+
```
kaustubh-karkare: You've hardcoded all the expectations, which wasn't the intent :)
This should be configurable, each testcase can set different expectations with the parser.

```
parser = Parser()

parser = Parser()
try:
  parser.parse(['--key=789'])
catch ArgumentParserException as e:
  self.assertTrue("unrecognized arguments: --key" in e.message)
```

Keep in mind [coding instruction 4.2](https://github.com/kaustubh-karkare/project-omega/blob/master/assignments/README.md):

> Additionally, make sure that your code is resuable: that it solves more than just the exact problem you are working on right now. In real life, requirements are always changing, and you want to make sure you can handle that with minimal effort. To simulate that, I will always evaluate your code at a higher standard than the problem statement given to you.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,100 @@
...
+    'required_command': KEY_COMMAND
+}
+
+LOCAL_COMMAND = {
+    'command_name': '--local',
+}
+
+REMOTE_COMMAND = {
+    'command_name': '--remote',
+    'conflicting_with': LOCAL_COMMAND
```
kaustubh-karkare: https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group

Can you implement something like this?

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,100 @@
...
+    return 1
+
+
+def validate_name(name_value):  # to validate the argument provided to --name command
+    if name_value.isalpha():
+        return 1
+    return -1
+
+
+def main(argv):
```
kaustubh-karkare: Lets use some OOPs concepts :)
All this should be in a class.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,100 @@
...
+    results = {}
+    commands_found = []
+    error_message = None
+    for arg in argv[1:]:
+        argument = arg.split('=')
+        command = argument[0]
+        commands_found.append(command)
+        if command in ('--local', '--remote'):
```
kaustubh-karkare: You cant hardcode any specific arguments in your logic that is supposed to be general.
Applies to various places below also.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,100 @@
...
+        return 1
+    return -1
+
+
+def main(argv):
+    results = {}
+    commands_found = []
+    error_message = None
+    for arg in argv[1:]:
+        argument = arg.split('=')
```
kaustubh-karkare: This works. But suggestion:
```
name, value = arg.partition('=')
```

solutions/Pranjal Rai/src/Assignment 1/TestCommandLParser.py
```diff
@@ -0,0 +1,48 @@
+import unittest
+import CommandLineParser as CLP
+
+
+class TestCommandLParser(unittest.TestCase):
```
kaustubh-karkare: Good tests!
However I would recommend that your class return proper json objects (not strings), since the actual use of an argument parse is to serve as an input to something else.
Also, if something goes wrong, the class could raise exceptions, allowing the caller to catch them and ignore if needed, instead of having to check what is inside the return value.

kaustubh-karkare (changes requested):

pranjalrai (commented):
1. Created a class and encapsulated all the functions inside it.
2. Raised exceptions in a more generalized way.
3. Adding a new command is more generalized now.
4. Added validation through regular expression. Each command doesn't need a separate function to validate its argument.

kaustubh-karkare (changes requested):
Excellent progress!

pranjalrai (commented):
1. Separated the previous __init__ and get_arguments methods in 2 different classes.
2. COMMANDS is now being accessed using self.COMMANDS in CommandLineParser class.
3. Added a generalized implementation for flag variables/commands.

Please review the changes.

pranjalrai (commented):
Created a separate main file that will control/call the application logic of Command-Line Parser and the unit test file.
Please review the changes.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,76 @@
...
+    COMMANDS = {}
+
+    def __init__(self, command_name, command_type, regular_expression, required_command, conflicting_command, is_local):
+        self.command_name = command_name
+        self.command_type = command_type
+        self.regular_expression = regular_expression
+        self.required_command = required_command
+        self.conflicting_command = conflicting_command
+        self.is_local = is_local
+        self.COMMANDS[self.command_name] = self
```
kaustubh-karkare: I don't think you need to have static properties like this.
Why? Because I'd like to be able to create separate `CommandLineParser` objects, each with a separate list of valid options.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+
+class CommandLineParser(AddCommand):
+
+    def __init__(self):
+        self.COMMANDS = super().COMMANDS
```
kaustubh-karkare: all caps variable names are usually reserved for constants, so here, I would just call it "self.commands" (lower case).

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,76 @@
...
+        else:
+            for command in commands_found:
+                conflicting_command = self.COMMANDS[command].conflicting_command
+                if conflicting_command is not None and conflicting_command in commands_found:
+                    error_message = 'The ' + command + ' and ' + conflicting_command + ' arguments cannot be used together'
+                required_command = self.COMMANDS[command].required_command
+                if required_command is not None and required_command not in commands_found:
+                    error_message = 'The ' + required_command + ' argument is required, but missing from input'
+
+        final_response = error_message
```
kaustubh-karkare: Instead of returning an error message, you should raise an exception, so that the application/testing layer can catch it and do whatever it wants with it (print it and exit, or log it and exit, or use some default fallback behavior).
Parsing a string return value (which is what this method generates) is more inconvenient. The API should be: the method either returns a Dict, or raises an exception.

solutions/Pranjal Rai/src/Assignment 1/TestCommandLParser.py
```diff
@@ -0,0 +1,55 @@
+import unittest
+from CommandLineParser import CommandLineParser
+
+
+class TestCommandLParser(unittest.TestCase):
+    def test_basic_usage(self):
+        command_line_param = ['./test', '--key=123', '--name=pranjal']
+        result = CommandLineParser().get_arguments(command_line_param)
+        print(result)
+        print(type(result))
```
kaustubh-karkare: You can remove all these random print statements in your code, once you're done running it manually.

solutions/Pranjal Rai/src/Assignment 1/main.py
```diff
@@ -0,0 +1,12 @@
...
+import sys
+
+KEY_COMMAND = CLP.AddCommand('--key', 'positive integer', r'\d+', None, None, False)
+NAME_COMMAND = CLP.AddCommand('--name', 'albhapets only', r'[a-zA-Z]+', '--key', None, False)
+LOCAL_COMMAND = CLP.AddCommand('--local', None, r'/\A\z/', None, '--remote', True)
+REMOTE_COMMAND = CLP.AddCommand('--remote', None, r'/\A\z/', None, '--local', True)
+
+if __name__ == '__main__':
+    CLP.CommandLineParser().get_arguments(sys.argv)
+    TEST.TestCommandLParser().call_functions()
```
kaustubh-karkare: This shouldn't be here, and the call_functions method shouldn't need to exist. The unit testing file is independent.

https://docs.python.org/3.6/library/unittest.html#unittest.main

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,76 @@
...
+class AddCommand:
+    COMMANDS = {}
+
+    def __init__(self, command_name, command_type, regular_expression, required_command, conflicting_command, is_local):
+        self.command_name = command_name
+        self.command_type = command_type
+        self.regular_expression = regular_expression
+        self.required_command = required_command
+        self.conflicting_command = conflicting_command
+        self.is_local = is_local
```
kaustubh-karkare: Based on your usage of this, I think `is_flag` is a better name.
The existence (or non-existence) of the option is equivalent to whether the flag is set or not set.

solutions/Pranjal Rai/src/Assignment 1/main.py
```diff
@@ -0,0 +1,12 @@
+import CommandLineParser as CLP
+import TestCommandLParser as TEST
+import sys
+
+KEY_COMMAND = CLP.AddCommand('--key', 'positive integer', r'\d+', None, None, False)
+NAME_COMMAND = CLP.AddCommand('--name', 'albhapets only', r'[a-zA-Z]+', '--key', None, False)
+LOCAL_COMMAND = CLP.AddCommand('--local', None, r'/\A\z/', None, '--remote', True)
+REMOTE_COMMAND = CLP.AddCommand('--remote', None, r'/\A\z/', None, '--local', True)
```
kaustubh-karkare: Instead of calling a separate method, you should just make it an API on your parser object.
```
parser = CLP.CommandLineParser()
parser.add_command(...)
parser. get_arguments(sys.argv)
```

kaustubh-karkare (changes requested):

pranjalrai (commented):
1. Removed static dictionary 'command'. Each Command-Line Parser object will have its own command dictionary.
2. is_local -> is_flag
3. Separated test file.
4. If something goes wrong, instead of returning a string, now raising an exception.
5. Removed print statements.
6. Replaced uppercase variable names.

Please review the changes.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,82 @@
...
+        
+        if error_message is None: # return json object if everything goes fine
+            final_response = json.dumps(results, sort_keys=True)
+            print(final_response)
+            return final_response
+        else: # raise an exception if some error occurred
+            try:
+                raise Exception(final_response)
+            except Exception as exception:
+                print(exception)
```
kaustubh-karkare: Btw, there are still print statements left in here.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,82 @@
...
+                if conflicting_command is not None and conflicting_command in commands_found:
+                    error_message = 'The ' + command + ' and ' + conflicting_command + ' arguments cannot be used together'
+                required_command = self.commands[command].required_command
+                if required_command is not None and required_command not in commands_found:
+                    error_message = 'The ' + required_command + ' argument is required, but missing from input'
+
+        final_response = error_message
+        
+        if error_message is None: # return json object if everything goes fine
+            final_response = json.dumps(results, sort_keys=True)
```
kaustubh-karkare: Returning a `dict` is more useful than returning a string.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,82 @@
...
+class AddCommand:
+
+    def __init__(self, command_name, command_type, regular_expression, required_command, conflicting_command, is_flag, commands):
+        self.command_name = command_name
+        self.command_type = command_type
+        self.regular_expression = regular_expression
+        self.required_command = required_command
+        self.conflicting_command = conflicting_command
+        self.is_flag = is_flag
+        commands[self.command_name] = self
```
kaustubh-karkare: This class shouldn't be responsible for populating the dict. Instead, the `CommandLineParser` class should be the one to create the dict, and add items to it. It will result in a much better API ... check out how python's argparse module is supposed to be used.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,82 @@
...
+                if re.fullmatch(regular_expression, value):
+                    results[command] = value
+                else:
+                    error_message = 'invalid argument to ' + command
+            
+            if error_message:
+                break
+
+        if error_message is not None:
+            """some error occurred already"""
```
kaustubh-karkare: If you raise an exception as soon as you found a problem, you wouldn't need to do this type of check at all.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,82 @@
...
+            pass
+        else:
+            for command in commands_found:
+                conflicting_command = self.commands[command].conflicting_command
+                if conflicting_command is not None and conflicting_command in commands_found:
+                    error_message = 'The ' + command + ' and ' + conflicting_command + ' arguments cannot be used together'
+                required_command = self.commands[command].required_command
+                if required_command is not None and required_command not in commands_found:
```
kaustubh-karkare: I'm not sure that your "required_command" check works as intended.
Shouldn't you be looping over `self.commands` instead of commands_found?

kaustubh-karkare (changes requested):

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
pranjalrai: This print statement was intended to let the user know about the exception raised. Should I remove this print statement?

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
pranjalrai: Actually commands_found contains all the commands received from the user in command line input, so we need to check if a required command is present in the list or not.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,77 @@
...
+import re
+
+
+class CommandLineParser():
+
+    def __init__(self):
+        self.commands = []
+
+    def add_command(self, command_name, format = None, required_command = None, conflicting_command = None, is_flag = False):
+        command = {'command_name': command_name, 'format': format, 'required_command': required_command, 'conflicting_command': conflicting_command, 'is_flag': is_flag}
```
kaustubh-karkare: Saving complex data in dictionaries makes it harder to perform typechecking, so the usual recommendation is to prefer storing it in a class.
```
command = Command(name=name, format=format, ...)
self.commands[command.name] = command
```
Note that `self.commands` is a dict, to avoid the O(n) complexity of your search_command method.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def search_command(self, name_of_command):
+        for command in self.commands:
+            if command['command_name'] == name_of_command:
+                    return command
+        return None
+
+    def get_arguments(self, argv):
+        results = {}
+        commands_found = []
```
kaustubh-karkare: Instead of a list, a set is more appropriate, right?

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,77 @@
...
+            command = None
+            value = None
+            argument = arg.split('=')
+
+            if len(argument) == 2:
+                command, value = argument
+            elif len(argument) == 1:
+                command = argument[0]
+            elif len(argument) > 2:
+                error_message = 'invalid number of arguments passed to ' + argument[0]
```
kaustubh-karkare: raise Exception('invalid number of arguments passed to ' + argument[0])

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,77 @@
...
+            instruction = self.search_command(command)
+            conflicting_command = instruction['conflicting_command']
+            if conflicting_command is not None and conflicting_command in commands_found:
+                try:
+                    raise Exception('The ' + command + ' and ' + conflicting_command + ' arguments cannot be used together')
+                except Exception as exception:
+                    return exception
+
+            required_command = instruction['required_command']
+            if required_command is not None and required_command not in commands_found:
```
kaustubh-karkare: This is still the pairwise requirements. Lets just simplify and do individual requirements.
So required=True means that if it is missing from input, you raise an exception.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,77 @@
...
+                try:
+                    raise Exception('The ' + required_command + ' argument is required, but missing from input')
+                except Exception as exception:
+                    return exception
+        
+        final_response = json.dumps(results, sort_keys=True)
```
kaustubh-karkare: The library should either returns a dict (instead of a string), or raise an exception (instead of returning it).

solutions/Pranjal Rai/src/Assignment 1/TestCommandLParser.py
kaustubh-karkare: ^

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,58 @@
...
+                command, value = argument
+            elif len(argument) == 1:
+                command = argument[0]
+            elif len(argument) > 2:
+                raise Exception('invalid number of arguments passed to ' + argument[0])
+
+            commands_found.add(command)
+
+            if command not in self.commands: # commands with at least one argument
+                    raise Exception(command + ' is not a recognized command')
```
kaustubh-karkare: extra indent

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,58 @@
...
+            if len(argument) == 2:
+                command, value = argument
+            elif len(argument) == 1:
+                command = argument[0]
+            elif len(argument) > 2:
+                raise Exception('invalid number of arguments passed to ' + argument[0])
+
+            commands_found.add(command)
+
+            if command not in self.commands: # commands with at least one argument
```
kaustubh-karkare: Not sure what the comment means.

solutions/Pranjal Rai/src/Assignment 1/CommandLineParser.py
```diff
@@ -0,0 +1,58 @@
...
+            if conflicting_command is not None and conflicting_command in commands_found:
+                raise Exception('The ' + command + ' and ' + conflicting_command + ' arguments cannot be used together')
+
+        for command in self.commands:
+            instruction = self.commands[command]
+            if flag is False and instruction['is_required'] == True and command not in commands_found:
+                raise Exception('The ' + command + ' argument is required, but missing from input')
+        
+        final_response = json.dumps(results, sort_keys=True)
+        return {'final_response': final_response}
```
kaustubh-karkare: I was suggesting that instead of:
```
final_response = json.dumps(results, sort_keys=True)
return {'final_response': final_response}
```
You could just do:
```
return final_response
```

kaustubh-karkare (approved):

kaustubh-karkare (changes requested):
Unit test needs to be updated.

solutions/Pranjal Rai/src/Assignment 1/TestCommandLParser.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_invalid_usage(self): #all invalid instructions here
+        parser = CLP.CommandLineParser()
+        parser.add_command('--key', format = r'\d+', is_required = True)
+        parser.add_command('--name', format = r'[a-zA-Z]+')
+        parser.add_command('--local', conflicting_command = '--remote', is_flag = True)
+        parser.add_command('--remote', conflicting_command = '--local', is_flag = True)
+
+        command_line_param = ['./test', '--key=1a3', '--name=pranjal'] #invalid key
+        with self.assertRaisesRegexp(Exception, 'invalid argument to --key'):
+            parser.get_arguments(command_line_param)
```
kaustubh-karkare: You might as well have done:
```
parser.get_arguments(['./test', '--key=1a3', '--name=pranjal'])
```

kaustubh-karkare (approved):

kaustubh-karkare:
Use separate branch for different problems. The code for build automation tool should not have appeared here.

PR #51, opened 2019-08-24
eskimo--assignment1 (Command Line Parser)

eskimo-01:


solutions/eskimo/assignment1/test.js
```diff
@@ -0,0 +1,45 @@
+var command_line=require("./command_line");
```
kaustubh-karkare: Lets use a framework like Jest to test your code properlu.

solutions/eskimo/assignment1/command_line.js
```diff
@@ -0,0 +1,341 @@
...
+    constructor()
+    { 
+        //Stores all commands
+        this.commandList=[];
+
+        //Stores the argument in JSON format
+        this.argument_JSON={};
+
+        //To check two Exclusive commands doesn't get together 
+        this.isSetExclusive= new Array(100).fill("0");
```
kaustubh-karkare: I always thought that the linter wanted spaces around operations like '='.

solutions/eskimo/assignment1/test.js
```diff
@@ -0,0 +1,45 @@
...
+
+command_line.command({
+    command:"key",
+    describe:"Input the key",
+    demandOption:true,
+    type:'number',
+    handler: function(){
+        console.log("Key");
+    }
```
kaustubh-karkare: Looks like the helper is called when the this command is provided?
That can be useful, but it isn't strictly necessary.
Ie - when I'm using a CLI argument parser, each command does not map to a specific method.
It might change something small, like a condition in an if-statement in some other part of the code.
But I guess it's fine now that you've added it.

solutions/eskimo/assignment1/command_line.js
```diff
@@ -0,0 +1,341 @@
...
+
+        //help command :- print how object needs to be created and all commands with descroption
+        var help_obj=
+        {
+            command:"help",
+            describe:"HELP",
+            handler:()=>{
```
kaustubh-karkare: These instructions are useful if you're a developer using the library, but not if you're a user of the larger program.
And so showing them in CLI output is not really useful I think. It would have been fine to explain all this in a comment next to the method that you're supposed to call, or in a readme file.

solutions/eskimo/assignment1/command_line.js
```diff
@@ -0,0 +1,341 @@
...
+
+        //Adds help to commandList
+        this.commandList.push(help_obj);
+    }
+
+    /**
+     * Add commands to commandList
+     * @param {JSON_object} add_command
+     */
+    command(add_command)
```
kaustubh-karkare: You could also provided default values like this:
```
add_option({command, type = 'string', required = false, ...})
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

Btw, note that I renamed the method. Because a function named "command" does not tell me what it does without looking inside (ie - is it get_command? or does_command_exist?).

solutions/eskimo/assignment1/command_line.js
```diff
@@ -0,0 +1,341 @@
...
+            if(val.startsWith("--"))
+            {
+                //Silces the "--" from argument
+                val=val.slice(2);
+
+                //Stores the current argument in a var
+                var input_command=val;
+
+                //Checks that a command exists in the commandList
+                var command_exist=0;
```
kaustubh-karkare: Instead of using this flag and iterating over the list (which is O(n)), why not put the commands in a map and look it up in O(1)?

solutions/eskimo/assignment1/test.js
```diff
@@ -0,0 +1,45 @@
...
+    type:"string",
+    handler: function(){
+        console.log("name");
+    }
+})
+
+command_line.command({
+    command:"local",
+    describe:"Set local",
+    ExclusiveIndex: 1,
```
kaustubh-karkare: This is not great API. Can you do something like:
```
group = command_line.add_mutually_exclusive_group()
group.add('local', ...)
group.add('remote', ...)
```

solutions/eskimo/assignment1/command_line.js
```diff
@@ -0,0 +1,341 @@
...
+                        }
+
+                        //Checks for command demandOption
+                        if(command_available.demandOption)
+                        {
+                            //Checks for '=' in argument if demandOption is true
+                            if(!val.includes("="))
+                            {
+                                //Error if argument doesn't have a value
+                                console.log("The '--"+command_available.command+"' argument is required, but missing from input.");
```
kaustubh-karkare: Instead of logging, you should throw errors here too, right?

solutions/eskimo/assignment1/command_line.js
```diff
@@ -0,0 +1,341 @@
...
+                console.log();
+                console.log("Commands:- ")
+                this.commandList.forEach(command_available => {
+                    console.log(command_available.command+": "+command_available.describe);
+                });
+            }
+        }
+
+        //Adds help to commandList
+        this.commandList.push(help_obj);
```
kaustubh-karkare: This shouldn't be a class property.
Ie - it is relevant only for a single call to your `execute` method, so you can create it there, return it, and then forget it.

solutions/eskimo/assignment1/command_line.js
```diff
@@ -0,0 +1,341 @@
...
+                                return;
+                            }
+
+                            //Takes the value of argument
+                            val=val.split("=")[1];
+
+                            //Checks the command type
+                            var command_type=command_available.type;
+
+                            if(command_type=='string')
```
kaustubh-karkare: How about a map?
```
const validators = {
  string: value => value.match("^[A-z]+$"),
  number: value => value.match("^[0-9]+$"),
  ...
}
```
You'll save a lot of lines.

solutions/eskimo/assignment1/command_line.js
```diff
@@ -0,0 +1,341 @@
...
+                //Error if argument is wrong
+                console.log("Wrong Argument: ",val);
+            }
+        });
+    }
+
+    /**
+     * Prints the JSON
+     */
+    display()
```
kaustubh-karkare: `this.argument_JSON` shouldn't be a class property.
Ie - it is relevant only for a single call to your execute method, so you can create it there, return it, and then forget it. If the caller wants, they can remember it, but that's not the job of this module.

solutions/eskimo/assignment1/command_line.js
eskimo-01: I used the linter of visual studio code.

solutions/eskimo/assignment1/command_line.js
eskimo-01: I checked that yargs have such type of property of providing command details when --help is typed, that's why I included it.

solutions/eskimo/assignment1/test.js
eskimo-01: I think of creating a commad line parser like yargs... so I used its format of input.

solutions/eskimo/assignment1/command_line.js
kaustubh-karkare: I'm not opposed to have a help.

There are 3 roles here:
* the developer of the library (you)
* the developer of the application (using this library)
* the user of the application

This output will be seen by "the user of the application", but is actually relevant only to "the developer of the application". If you want to show something for "the user of the application", you could create a help message based on the exact commands that were specified. This is where the description field comes in handy :)

solutions/eskimo/assignment1/command_line.js
```diff
@@ -0,0 +1,339 @@
...
+var commandList=[];
+
+//Stores Command Name
+var commandName={};
+
+//Stores smallCommand Name
+var commandsmallName={};
+
+//Stores the argument in JSON format
+var argument_JSON={};
```
kaustubh-karkare: There shouldn't be any global variables like this. This violates OOPs :)
If your parser class needs anything, it can create class properties.

PR #52, opened 2019-08-24
Solution for assignment 1

AkashAsthana:


solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,188 @@
...
+
+/*
+* Defining valid arguments as objects 
+* with properties like isReq,
+* type and cantBeUsedWith
+*/
+var key = new Arguments(true, "number");
+var name = new Arguments(false, "string");
+var local = new Arguments(false, "boolean", "remote");
+var remote = new Arguments(false, "boolean", "local");
```
kaustubh-karkare: In order to be truly reusable ([coding instruction 4.2](https://github.com/kaustubh-karkare/project-omega/blob/master/assignments/README.md)), you can create a class like this (where we can add different options for each test):
```
parser = Parser()
```

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,188 @@
...
+
+
+
+/*
+* Function for extracting out 
+* argument's name from 
+* the commands array
+*/
+function extractArgumentName(argumentName) {
```
kaustubh-karkare: If the method takes `argumentName` as input, why does it have to do any process to extract stuff.
Ie - the variable name is not accurate ([coding instruction 3](https://github.com/kaustubh-karkare/project-omega/blob/master/assignments/README.md)), you can just use `argument` or `option` if you wish.

solutions/akash-asthana/Assignment-1/assignment1.test.js
```diff
@@ -0,0 +1,28 @@
+//USED JEST FOR TESTING THE PROJECT
+const { extractArgumentName, extractArgumentValue, typeCheck } = require("./assignment1");
+
```
kaustubh-karkare: What you have here are some "unit tests", which is a good start.
Can you also write some "integration tests"?

AkashAsthana: Yes, I'll do that!

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,236 @@
...
+var errorPresent = false;
+
+
+class Parser {
+
+    addOption(name, isReq, type, cantBeUsedWith) {
+        this.options.push({ name: name, isReq: isReq, type: type, cantBeUsedWith: cantBeUsedWith });
+    }
+}
+Parser.prototype.options = [];
```
kaustubh-karkare: When you set the array directly on the prototype, it becomes a static variable, which is incorrect.
```
p = new Parser();
p.options.length; // 0
p.addOption('test')
p.options.length; // 1
q = new Parser();
q.options.length // 1, which is wrong
```
Lets not deal with prototypes in this problem, that concept isn't needed here.

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,236 @@
...
+
+
+
+var parser = new Parser();
+
+parser.addOption("key", true, "number");
+parser.addOption("name", false, "string");
+parser.addOption("local", false, 'boolean', "remote");
+parser.addOption("remote", false, "boolean", "local");
```
kaustubh-karkare: [Boolean arguments are bad in general](https://softwareengineering.stackexchange.com/a/370801), since it is not possible for me to figure out what the true/false means by looking at these calls alone. You're forcing me to look at your implementation to figure it out, which hurts readability.

Instead, you can do something like [this](https://davidwalsh.name/destructuring-function-arguments):
```
parser.addOption({name: 'key', isRequired: true, type: "number"});
```
That way, it is clear.

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,236 @@
...
+            return false;
+        }
+    }
+    return true;
+}
+
+
+
+//MAIN FUNCTION
+function parseArgumentsIntoJSON(commands) {
```
kaustubh-karkare: This should be a method in your class (along with the previous ones). That way,
```
const p = new Parser();
p.addOption(...)
result = p. parseArgumentsIntoJSON(commands);
```
Btw, you're not allowed to create the Parser object in this file, you can only define the class.
You should only create class instances outside the library, in your test (this is how your integration tests will look like).

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,236 @@
...
+
+    for (var i = 0; i < commands.length; i++) {
+
+        var argumentName = extractArgumentName(commands[i]);
+        var argumentValue = extractArgumentValue(commands[i]);
```
kaustubh-karkare: Do you really need separate functions for this? How about:
```
var [name, value] = commands[ii].split("=")
```

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,236 @@
...
+    }
+    return true;
+}
+
+
+
+//MAIN FUNCTION
+function parseArgumentsIntoJSON(commands) {
+
+    for (var i = 0; i < commands.length; i++) {
```
kaustubh-karkare: [Use `ii` instead of `i`.](https://stackoverflow.com/a/101071)

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,236 @@
...
+}
+
+
+
+/*
+* Function to check if arguments
+* given by the user are defined
+*/
+function checkIfArgumentDefined(argumentName) {
+    for (var i = 0; i < parser.options.length; i++) {
```
kaustubh-karkare: To avoid looping over the list, you can store the options in a map<name, option> instead.

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,236 @@
...
+            argumentNameToBeChecked = parser.options[i].name;
+            break;
+        }
+    }
+
+    if (parser.options[i].type === "number") {
+        return !isNaN(argumentValueToBeChecked);
+    }
+    if (parser.options[i].type === "string") {
+        return /^[a-zA-Z]+$/.test(argumentValueToBeChecked);
```
kaustubh-karkare: Why does a string have restrictions like this?
All characters can be valid inside a string.

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,236 @@
...
+* Function to check if
+* the mandatory arguments are
+* present in the commands
+* given by the user
+*/
+function areRequiredArgumentsPresent(argumentNameArray) {
+    for (var i = 0; i < parser.options.length; i++) {
+        if (parser.options[i].isReq === true) {
+            if (!argumentNameArray.includes(parser.options[i].name)) {
+                errorStatement += `Error :The argument "--${parser.options[i].name}" is required\n`;
```
kaustubh-karkare: https://www.facebook.com/groups/856311934764144/permalink/861189610943043/

> no global variables, no standalone functions, etc.

errorStatement & parser are global variables inside this library file.

I'd recommend raising exceptions here.

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,236 @@
...
+
+
+
+/*
+* Function to check if arguments 
+* given by the user can be used together
+*/
+function ifArgumentsCanBeUsedTogether(argumentNameArray) {
+    for (var i = 0; i < argumentNameArray.length; i++) {
+        if (argumentNameArray.includes(parser.options[i].cantBeUsedWith)) {
```
kaustubh-karkare: You're assuming that only 2 arguments can be mutually exclusive.
What if we wanted to allow exactly one of 5 options?

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,236 @@
...
+    * Printing out the JSON output
+    * if no error is found
+    */
+    if (!errorPresent) {
+        var finalResult = {};
+        for (var i = 0; i < argumentNameArray.length; i++) {
+            finalResult[argumentNameArray[i]] = argumentValueArray[i];
+        }
+
+        console.log(JSON.stringify(finalResult));
```
kaustubh-karkare: In order to make this library actually useful, you should return the JSON object, and then let the application print it (or compare against expected value in test, or whatever else it wants to do).

solutions/akash-asthana/Assignment-1/assignment1.js
```diff
@@ -0,0 +1,236 @@
...
+
+        if (!typeCheck(argumentName, argumentValue)) {
+            errorStatement += `Error : Type of the argument "--${argumentName}" is invalid\n`;
+            errorPresent = true;
+        }
+
+
+
+        argumentNameArray.push(argumentName);
+        argumentValueArray.push(argumentValue);
```
kaustubh-karkare: Why not use a map here? Instead of these 2 arrays which represent keys and values of that map.

solutions/akash-asthana/Assignment-1/assignment1.js
AkashAsthana: I intended a "string" to consist of only letters so that the option "--name" can have a proper name like "akash"  instead of something like "akash123". Anyway i'll change it.

PR #53, opened 2019-08-24
scheleon - Assignment - 1

scheleon:
Command Line parser using nodejs


solutions/scheleon/argument-parser-assignment-1/parser.js
```diff
@@ -0,0 +1,265 @@
...
+    getValueType() {
+        return this.valueType;
+    }
+
+    getDefaultValue() {
+        return this.defaultValue;
+    }
+
+    getValue() {
+        return this.value;
```
kaustubh-karkare: There's no reason to save the value on this class.
Ie - this class just defines the expectations that the value must meet.
Once the value is confirmed to meet them, it can be return to the calling code.

solutions/scheleon/argument-parser-assignment-1/parser.js
```diff
@@ -0,0 +1,265 @@
...
+     * set value after type casting to {valueType}
+     * @param {*} value 
+     */
+    setValue(value){
+        if (this.valueType == "String") { //
+            try { //
+                this.value = String(value);
+            } catch (err) {
+                var valueType = typeof value;
+                throw "Expected : " + this.valueType + ", found "
```
kaustubh-karkare: How about throwing custom exceptions?

solutions/scheleon/argument-parser-assignment-1/parser.js
```diff
@@ -0,0 +1,265 @@
...
+    * @param {*} defaultValue 
+    */
+    addArgument(shortLabel, largeLabel, isValueRequired, valueType, defaultValue) {
+        for (var i = 0; i < this.arguments.length; i++) {
+            if(shortLabel == this.arguments[i].getShortLabel() 
+                || largeLabel == this.arguments[i].getLargeLabel()) {
+                throw "Argument with the same label already exists";
+                return;
+            }
+        }
```
kaustubh-karkare: This loop is unnecessary O(n) computation.
Can you make it O(1)?

solutions/scheleon/argument-parser-assignment-1/parser.js
```diff
@@ -0,0 +1,265 @@
...
+        /**
+        * For Argument type : --key
+        */
+        var shortInputArgv = "^-[\\w]+=[\\w ]+$"
+        /**
+         * For Argument type : -key=value
+         */
+        var largeInputArgv = "^--[\\w]+=[\\w ]+$";
+        /**
+         * Reuired if argument is of type : -key value
```
kaustubh-karkare: Not sure why you're supporting this "single dash" case.

solutions/scheleon/argument-parser-assignment-1/parser.js
```diff
@@ -0,0 +1,265 @@
...
+         */
+        var inputValue = "^[\\w -]+$";
+
+        for(var itr=2; itr < process.argv.length; itr++) {
```
kaustubh-karkare: `process.argv` should be given as input to this class.

solutions/scheleon/argument-parser-assignment-1/parser.js
```diff
@@ -0,0 +1,265 @@
...
+
+                var label = process.argv[itr].match("[\\w]+")[0];
+                var argumentIndex = this.findArgumentIndexByLabel(label);
+                
+                /**
+                 * Set { "value" : true }, if isValueRequired == false, else
+                 * increase the iterator and perform a check for value on next argument 
+                 */
+                if (this.arguments[argumentIndex].getIsValueRequired() == true) { //
+                    itr++;
```
kaustubh-karkare: Does `itr` stand for iterator?
This isn't really an iterator, it is just an index, so using `ii` is fine.
https://stackoverflow.com/a/101071

solutions/scheleon/argument-parser-assignment-1/test-arguments.js
```diff
@@ -0,0 +1,6 @@
+const parser = require("./parser");
+
+parser.addArgument("w", "weight", true, "String", "hello");
+parser.addArgument("j", "jojo", false, "Boolean", false);
+
+console.log(parser.listArgsProvided());
```
kaustubh-karkare: You should use a proper testing framework like Jest to test your code.

solutions/scheleon/argument-parser-assignment-1/parser.js
scheleon: Yes, I can populate the hashmap parallelly and check if an option with given label exists or not.

solutions/scheleon/argument-parser-assignment-1/parser.js
scheleon: Sorry, I the order of comment for description was wrong. And if argument is of type 
-key value
argument[key] should be set to value as many command line programs use this format of implementation including nmap.

solutions/scheleon/argument-parser-assignment-1/parser.js
scheleon: Exactly, the single variable always makes me lose my concentration so I thought of using itr. But will change it to ii.

solutions/scheleon/argument-parser-assignment-1/parser.js
scheleon: Yes, I can populate the hashmap parallelly and check if an option with given label exists or not.

solutions/scheleon/argument-parser-assignment-1/parser.js
scheleon: Will make necessary changes for i/o.

solutions/scheleon/argument-parser-assignment-1/parser.js
scheleon: I just thought that user would be entering his own custom options, so what if he needs an option with a default value which he won't use much? So I'm was storing those values to add them later if not provided in arguments.

solutions/scheleon/argument-parser-assignment-1/parser.js
scheleon: Should both large and small labels be considered for an option?

solutions/scheleon/argument-parser-assignment-1/test-arguments.js
```diff
@@ -0,0 +1,8 @@
+const parser = require("./parser");
+
+parser.addArgument("w", "weight", true, "String", "hello");
+parser.addArgument("j", "jojo", true, "Boolean", true);
+//parser.addArgument("j", "jojo", false, "Boolean", false);
+
```
kaustubh-karkare: This is fine for testing things yourself, but you need to use something like https://jestjs.io/ to write proper tests too.

solutions/scheleon/argument-parser-assignment-1/parser.js
```diff
@@ -0,0 +1,284 @@
...
+                if (this.arguments[argumentIndex].getIsValueRequired() == true) { //
+                    ii++;
+
+                    /**
+                     * Check if the iterator reaches the end of the passed arguments array
+                     */
+                    if(ii == providedArgumentList.length) {
+                        throw new EmptyArgumentValueException(label);
+                    }
+                    var argvValueProvided = providedArgumentList[ii].match(inputValue);
```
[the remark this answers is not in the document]
scheleon: I am first verifying the arguments then extracting the values from them
-key=value then I'm extracting key and value pairs using different regex expressions.
I can actually check for regex only once and then split the strings using '=', will that be okay?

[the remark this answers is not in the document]
kaustubh-karkare: Yep.

PR #54, opened 2019-08-24
Solution to assignment-1 added

shahayush457:


solutions/ayush-shah/argument-parser/README.md
```diff
@@ -0,0 +1,27 @@
+# Examples
```
kaustubh-karkare: You're using this readme file to prove that the command works.
Instead, just use a proper testing framework like Jest.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,99 @@
+import sys
+Integer, String, Others, Required = set(), set(), set(), set()
+errors, Json = dict(), dict()
```
kaustubh-karkare: You shouldn't put these outside the class, they can all be class properties / local variables.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,99 @@
...
+    parse.add_argument('--name', 'no', 'string')
+    parse.add_argument('--local', 'no', 'others')
+    parse.add_argument('--remote', 'no', 'others')
+    length_of_arguments = len(argument)
+    if length_of_arguments == 1:
+        print("Error: no arguments given in input")
+        return 0
+    check = False
+    keys = list()
+    for arguments in range(1, length_of_arguments):
```
kaustubh-karkare: This stuff should also be inside the class.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,99 @@
...
+    response, key = parse.check_required(keys)
+    if not response:
+        parse.store_error(key, "required")
+    if bool(errors):
+        parse.display_error()
+    else:
+        parse.display_result()
+
+if __name__ == '__main__':
+    main(sys.argv)
```
kaustubh-karkare: This is intended to be a library, so I'd recommend not doing any work in this file.
Just implement the class, and then from the test file, you can use it.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,99 @@
...
+    def check_required(self, keys):
+        for check in Required:
+            flag = False
+            for key in keys:
+                if key == check:
+                    flag = True
+            if not flag:
+                return False, check
+        return True, "ok"
+    def display_error(self):
```
kaustubh-karkare: I like that you're displaying errors, but in general, displaying isn't enough.
This is just a library that will be called by a larger application, so you should raise exceptions instead.
The caller can then choose to display the errors, or do something else.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,99 @@
...
+                    print("Error: The value for argument '"+key+"' is missing")
+                if error == "required":
+                    print("Error: The argument '"+key+"' is required, but missing from input")
+                if error == "local and remote":
+                    print("Error: The '--local' and '--remote' arguments cannot be used together")
+                if error == "not-int":
+                    print("Error: The value for argument '"+key+"' must be integer")
+                if error == "not-string":
+                    print("Error: The value for argument '"+key+"' must be string")
+    def display_result(self):
```
kaustubh-karkare: This is not true JSON.
Check out the `json` module.

kaustubh-karkare (changes requested):

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -1,99 +1,94 @@
...
+
+    def display_result(self):
+        to_json = json.dumps(self.Json)
+        return to_json
+
+    def main(self, argument):
+        self.add_argument('--key', 'yes', 'integer')
+        self.add_argument('--name', 'no', 'string')
+        self.add_argument('--local', 'no', 'others')
+        self.add_argument('--remote', 'no', 'others')
```
kaustubh-karkare: You shouldn't be adding arguments in this file at all.
Each test method can add its own arguments, but the code in this file needs to be very generalized and reusable.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -1,99 +1,94 @@
...
-    if bool(errors):
-        parse.display_error()
-    else:
-        parse.display_result()
+        response, key = self.check_required(keys)
+        if not response:
+            return "Error : argument '" + key + "' is required but missing"
+        return self.display_result()
 
 if __name__ == '__main__':
```
kaustubh-karkare: Don't need a main section at all, since this is supposed to be a reusable library.

kaustubh-karkare (changes requested):

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,85 @@
+import json
+
+class Parser:
+    """
+    Class for adding the command line arguments
+    storing errors, storing results and displaying them
+    """
+    Integer, String, Others, Required = set(), set(), set(), set()
```
kaustubh-karkare: - These are static properties on the class which prevents multiple parser objects from being created and used simultaneously. We shouldn't need these, we can create normal properties.
- We shouldn't create a separate set for each type. Instead, you should store type information as part of an argument object, and a list of argument objects can be stored on the parser object.

kaustubh-karkare (changes requested):

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,104 @@
...
+    Class for adding the command line arguments
+    storing errors, storing results and displaying them
+    """
+    
+    def __init__(self):
+        self.arguments = []
+        self.Json = {}
+
+    def add_argument(self, argument, required, types):
+        add = Options(argument, types, required)
```
kaustubh-karkare, line 32: You want the code you write to be easily readable.
This change in ordering of arguments is random and therefore confusing.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,104 @@
...
+        super(ParseError, self).__init__(message)
+
+class Options(object):
+    """
+    Class for storing argument information
+    """
+    
+    def __init__(self, argument, types, required):
+        self.argument = argument
+        self.types = types
```
kaustubh-karkare, line 18: The variable name "types" suggests a list of things, not a single string, which is what you're providing.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,104 @@
...
+        for arguments in range(1, length_of_arguments):
+            args = argument[arguments]
+            key = args.partition('=')[0]
+            if key == '--local' and not check:
+                check = True
+            elif key == '--remote' and not check:
+                check = True
+            elif check and key in ('--local', '--remote'):
+                return True
+        return False
```
kaustubh-karkare, line 62: This is not allowed.
You cannot assume that --local and --remote are the only mutually exclusive options. Instead, you need to make that configurable, just like `add_option`

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,104 @@
...
+        
+class Parser:
+    """
+    Class for adding the command line arguments
+    storing errors, storing results and displaying them
+    """
+    
+    def __init__(self):
+        self.arguments = []
+        self.Json = {}
```
kaustubh-karkare, line 29: Don't save the result on the class, just return it.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,104 @@
...
+        length_of_arguments = len(argument)
+        if length_of_arguments == 1:
+            raise ParseError("Error: no arguments given in input")
+        if self.check_local_and_remote(argument):
+            raise ParseError("Error: The '--local' and '--remote' arguments cannot be used together")
+        keys = list()
+        for arguments in range(1, length_of_arguments):
+            args = argument[arguments]
+            key = args.partition('=')[0]
+            value = args.partition('=')[2]
```
kaustubh-karkare, line 78: More concise:
```
key, _, value = args.partition('=')

File: solutions/ayush-shah/argument-parser/parser.py [hunk 15]
```diff
@@ -0,0 +1,104 @@
...
+            keys.append(key)
+            if key not in ("--remote", "--local"):
+                self.store_result(key, value)
```
kaustubh-karkare, line 81: You shouldn't save the result before doing the validation below.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,104 @@
...
+            for obj in self.arguments:
+                if obj.argument == key:
+                    flag = True
+            if not flag:
+                raise ParseError("Error: invalid argument '" + key + "'")
+            if '=' not in args:
+                raise ParseError("Error: The value for argument '" + key + "' is missing")
+            flag = False
+            for obj in self.arguments:
+                if obj.types == "integer" and obj.argument == key:
```
kaustubh-karkare, line 92: Using strings like "integer" in your code is bad practice, because it is easy to make typing mistakes. Use constants defined at the top of your file, or enums, etc.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def add_argument(self, argument, required, types):
+        add = Options(argument, types, required)
+        self.arguments.append(add)
+
+    def store_result(self, key, value):
+        self.Json[key] = value
+
+    def check_required(self, keys):
+        for check in self.arguments:
```
kaustubh-karkare, line 39: "check" is not a useful variable name, since I have no idea what it contains without looking at the usage. The proper name here is "option" or something.

solutions/ayush-shah/argument-parser/parser.py
```diff
@@ -0,0 +1,104 @@
...
+            flag = False
+            if not check.required:
+                continue
+            for key in keys:
+                if key == check.argument and check.required:
+                    flag = True
```
kaustubh-karkare, line 45: This is an unnecessary O(n) operation. You can optimize it to O(1).

kaustubh-karkare (changes requested):

solutions/ayush-shah/argument-parser/README.md
kaustubh-karkare: You can get rid of this file now, since (a) the test proves that it works, and (b) these examples do not work anymore.

PR #56, opened 2019-08-25
Solution for Assignment 1

Abhilasha06:


kaustubh-karkare:
Btw, here's an example of what I'm hoping to see:
```
def test_simple(self):
    parser = Parser()
```

solutions/Abhilasha06/Assignment1/test.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        ans = MyParser.check_argument("./test --key=123")
+        self.assertEqual('{"--key": "123"}', ans)
+    def test_simple1(self):
+        """
+        key and name defined
+        """
+        ans = MyParser.check_argument("./test --key=12345 --name=kaustubh")
+        self.assertEqual('{"--key": "12345", "--name": "kaustubh"}', ans)
+    def test_simple2(self):
```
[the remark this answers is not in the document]
Abhilasha06: I'll change the names of the tests... but pylint says to have a docstring for every method.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,94 @@
...
+@author: Abhilasha
+"""
+class MyParser:
+    """
+    command line parser
+    """
+    @classmethod
+    def add_option(cls, option_name, option_type, reqirement):
+        """
+        adding the options
```
kaustubh-karkare: These comments aren't particularly useful though.
I mean, you don't need to add comments for the sake of adding comments.

solutions/Abhilasha06/Assignment1/test.py
```diff
@@ -0,0 +1,43 @@
...
+import unittest
+from my_parser import MyParser
+class Tests(unittest.TestCase):
+    """
+    some test cases
+    """
+    MyParser.add_option('--key', 'int', 'True')
+    MyParser.add_option('--name', 'str', 'True')
+    MyParser.add_option('--local', 'str', 'False')
+    MyParser.add_option('--remote', 'str', 'False')
```
kaustubh-karkare: In general, boolean arguments aren't good for code readability.

https://softwareengineering.stackexchange.com/a/370801

In python, fortunately you can specify the argument name, so:
```
MyParser.add_option('--local', type='str', required=False)
```
PS. True/False don't need to be strings, right?

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,94 @@
...
+    @classmethod
+    def check_argument(cls, argument):
+        """
+        to parse the given input and give error if it is invalid
+        """
+        import json
+        data = {}
+        list_of_fields = ["--key", "--local", "--remote", "--name"]
+        argument = list(argument.split(" "))
+        no_of_args = (int)(format(len(argument)))
```
kaustubh-karkare: Why do you need to use `format` and then cast to `int`?

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,94 @@
...
+        if no_of_args == 1:
+            ans = "Error: The '--key' argument is required, but missing from input."
```
kaustubh-karkare: This message about `--key` makes an assumption that one of the arguments is always `--key`, which isn't true.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,94 @@
...
+        for args in argument[1:]:
+            splitted_word = args.split('=')
+            size = int(len(splitted_word))
```
kaustubh-karkare: len already returns an int.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,94 @@
...
+            if size > 2:
+                ans = "Error: Invalid argument"
+            elif splitted_word[0] not in list_of_fields:
+                return"Error: Field not defined"
```
kaustubh-karkare: When you encounter cases like this, people usually raise exceptions, instead of returning.
That way, the caller of your library can assume that if it returned value, everything is fine, but if it didn't, then an exception would be raised that can be caught.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,94 @@
...
+        to check if remote is valid or not
+        """
+        if size != 1:
+            mssg2 = "Error: --remote takes no value"
+        elif "--local" in data:
+            mssg2 = "Error: --local and --remote can't occur together"
+        else:
+            data["--remote"] = "true"
+            mssg2 = "successful"
+        return mssg2
```
kaustubh-karkare: These 2 methods (check_local & check_remote) make assumptions that --local and --remote are valid. This is not true. Ie - the test should be able to configure that only one of --key and --name is allowed. Everything should be flexible here.

solutions/Abhilasha06/Assignment1/test.py
```diff
@@ -0,0 +1,43 @@
...
+        ans = MyParser.check_argument("./test --local --name=abcd")
+        self.assertEqual('Error: The \'--key\' argument is required,'
+                         ' but missing or has invalid value.', ans)
+    def test_undefined_field(self):
+        ans = MyParser.check_argument("./test --age=19")
+        self.assertEqual('Error: Field not defined', ans)
+    def test_local_remote_together(self):
+        ans = MyParser.check_argument("./test --key=19 --local --remote")
+        self.assertEqual('Error: --local and --remote can\'t occur together', ans)
+    def test_key_and_local_defined(self):
```
kaustubh-karkare: It would help to add some empty lines between your tests to make them easier to read.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,94 @@
...
+
+"""
+Created on Sun Aug 25 15:21:34 2019
+@author: Abhilasha
+"""
+class MyParser:
+    """
+    command line parser
+    """
+    @classmethod
```
kaustubh-karkare: These shouldn't be @classmethods, you should be able to add separate options in each test, that are all independent of each other.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,94 @@
...
+    def add_option(cls, option_name, option_type, reqirement):
+        """
+        adding the options
+        """
+        list_of_fields = []
+        list_of_fields.append(option_name)
```
kaustubh-karkare: I'm confused here. `list_of_fields` is a local variable whose value will be lost at the end of the method, right?

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,148 @@
...
+                    elif(splitted_word[0]=="--remote"):
+                        if opt.nargs == size and size == 2:
+                            self.check_key(opt.dtype, splitted_word[1], opt.nargs, size)
+                        elif size == 1:
+                            self.data["--remote"] = "true"
+                            
+            if flag==0:
+                raise Error("Invalid field given.")
+          
+        return(json.dumps(self.data))
```
kaustubh-karkare: You shouldn't translate the result into string. You can return a dict, which in general is more useful, and then let your application or test convert to string if it needs to. However, based on your tests, it isn't needed.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,148 @@
...
+        else:
+            self.data["remote"] = given_type
+            
+            
+    def check_argument(self, argument):
+        """
+        to parse the given input and give error if it is invalid
+        """
+        import json
+        argument = list(argument.split(" "))
```
kaustubh-karkare: Given that the typical input to libraries like this is `sys.argv`, this method could take a list of strings (already split), instead of a single one.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,148 @@
...
+            
+        else:
+            self.data["remote"] = given_type
+            
+            
+    def check_argument(self, argument):
+        """
+        to parse the given input and give error if it is invalid
+        """
+        import json
```
kaustubh-karkare: Why not put these imports at the top of the file?

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,148 @@
...
+        
+        if no_of_args == 1:
+            raise Error("No arguments.")
+            
+        for args in argument[1:]:
+            splitted_word = args.split('=')
+            size = len(splitted_word)
+            flag=0
+            
+            for opt in self.options:
```
kaustubh-karkare: Instead of looping over all items in the list, you can just create a dict right?
```
matching_option = self.option_name_to_option.get(splitted_word[0])
```

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,148 @@
...
+            
+                if (splitted_word[0] == opt.field_name) :
+                    flag=1
+                    if(splitted_word[0] == "--key"):
```
kaustubh-karkare: Ah, this is still not generalized, right?
Yes, each test can configure which options it can take (which is progress), but the choice is limited, only one of these 4 is supported. You want to allow arbitrary options in the tests, so that I can create any random string and use it as an option.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,148 @@
+# -*- coding: utf-8 -*-
+
+"""
+Created on Sun Aug 25 15:21:34 2019
+@author: Abhilasha
+"""
+class Error(Exception):
```
kaustubh-karkare: How about calling this something more specific? Like MyParserError?

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,148 @@
...
+    command line parser
+    """
+    def __init__(self):
+        self.options=[]
+        self.data={}
+        
+
+    def add_option(self, field_name, dtype, nargs):
+        option_being_added = Option(field_name, dtype, nargs)
+        self.options.append(option_being_added)
```
kaustubh-karkare: Yep, this is in the right direction.

solutions/Abhilasha06/Assignment1/test.py
```diff
@@ -9,62 +9,76 @@ class Tests(unittest.TestCase):
     """
     
     def setUp(self):
-        self.st=MyParser()
+        self.Parser=MyParser()
```
kaustubh-karkare: Naming conventions: I think `self.parser` is a better name, since capital first letter is usually reserved for class names, not instances names.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,111 @@
...
+        
+    def check_size(self, field_size, given_size):
+        """
+        to check if the no of arguments given is valid
+        """      
+        if field_size != given_size:
+            raise MyParserError("Invalid number of arguments.")
+        
+    
+    def check_dtype(self,field_name, given_type, given_value, size):
```
kaustubh-karkare: Did you run the linter? I think convention says that there should be a space after `self,` for consistency with other arguments.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,111 @@
...
+        if given_type == "str":
+            
+            if given_value.isalpha():
+                self.data[field_name] = given_value  
+            else:
+                raise MyParserError("The field has invalid value.")
+            
+        elif given_type == "int":
+            
+            if bool(re.match('^[0-9]+$',given_value)):
```
kaustubh-karkare: `given_value.isdigit()`

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,111 @@
...
+            raise MyParserError("Invalid number of arguments.")
+        
+    
+    def check_dtype(self,field_name, given_type, given_value, size):
+        """
+        to check if the datatype of the given arguments are valid
+        """
+        if given_type == "str":
+            
+            if given_value.isalpha():
```
kaustubh-karkare: I don't think you need this. You can have any characters in a string.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,111 @@
...
+            splitted_word = args.split('=')
+            size = len(splitted_word)
+            flag = 0
+            
+            for opt in self.options:
+                if (splitted_word[0] == opt.name):
+              
+                    flag = 1
+                
+                    self.check_size(opt.nargs, size)
```
kaustubh-karkare: I think the conventional definition of `nargs` is number of values excluding the key name. So `size - 1` is probably more accurate. But for nargs = 3, we'd have to do something like `--name=abc=def=ghi`, which can be confusing, right? Ie - `--name abc def ghi` is how it usually works (they're separate values, not equal to each other), but supporting both spaces and `=` to separate key and value is a more complex problem.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,111 @@
...
+                    if size == 1:
+                         self.data[opt.name] = "True"
+                    else:
+                        self.check_dtype(splitted_word[0], opt.dtype, splitted_word[1],size)
```
kaustubh-karkare: ^ Just to call out the problems here, you've assumed that size = 2 only. What if there were more? We'd want to check the types of all values?

solutions/Abhilasha06/Assignment1/test.py
```diff
@@ -0,0 +1,85 @@
...
+    def test_key_and_name_defined(self):
+        self.Parser.add_option('--key', dtype='int', nargs = 2)
+        self.Parser.add_option('--name', dtype='str', nargs = 2)
+        ans = self.Parser.check_argument(['./test', '--key=12345', '--name=kaustubh'])
+        self.assertEqual({'--key': '12345', '--name': 'kaustubh'}, ans)
+        
+        
+    def test_invalid_key(self):
+        self.Parser.add_option('--key', dtype='int', nargs = 2)
+        with self.assertRaises(Exception) as context:
```
kaustubh-karkare: self. assertRaises(MyParserError)

solutions/Abhilasha06/Assignment1/test.py
```diff
@@ -0,0 +1,85 @@
...
+            self.Parser.check_argument(['./test', '--key=cat'])
+        self.assertEqual('The field has invalid value.', str(context.exception))
+        
+        
+    def test_local_remote(self):
```
kaustubh-karkare: test_unexpected_option

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,105 @@
...
+                self.data[field_name] = given_value
+            else:
+                raise MyParserError("The field has invalid value.")
+            
+        
+    def check_options(self, argument):
+        """
+        to parse the given input and give error if it is invalid
+        """
+        no_of_args = len(argument)
```
kaustubh-karkare: Since this is a list, we should try to make it clear based on the name. "arguments"?

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,105 @@
...
+        
+        if no_of_args == 1:
+            raise MyParserError("No options given.")
+            
+        for args in argument[1:]:
+            splitted_word = args.split('=')
+            size = len(splitted_word)
+            flag = 0
```
kaustubh-karkare: This kind of variable is a flag, true, but we usually want a more accurate name, like "found_option"

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,105 @@
...
+            
+            for opt in self.options:
+                if splitted_word[0] == opt.name:
+              
+                    flag = 1
+
+                    if size > 2:
+                        raise MyParserError("Too many arguments.")
+                        
+                    elif not opt.is_required and size == 1:
```
kaustubh-karkare: Hmm, this was slightly difficult to read, since each branch of the "if" clause has different types of conditions. It is usually easier to read if there is only 1 type of condition. Eg - in this case, this outer if condition should only focus on size, and then you could have inner ones that are checking is_required. That way, it is very easy to verify that all cases have been covered just by skimming through the code.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,105 @@
...
+        self.is_required = is_required
+
+       
+class MyParser:
+    """
+    command line parser
+    """
+    def __init__(self):
+        self.options = []
+        self.data = {}
```
kaustubh-karkare: You don't need to save this on the class at all.
You can just create the dict in your `check_options` method, and return it without saving.

solutions/Abhilasha06/Assignment1/test.py
```diff
@@ -0,0 +1,74 @@
...
+        with self.assertRaises(MyParserError) as context:
+            self.parser.check_options(['./test', '--key=cat'])
+        self.assertEqual('The field has invalid value.', str(context.exception))
+        
+        
+    def test_unexpexted_option(self):
+        self.parser.add_option('--local', dtype='str', is_required=False)
+        with self.assertRaises(MyParserError) as context:
+            self.parser.check_options(['./test', '--local', '--remote'])
+        self.assertEqual("Invalid field given.", str(context.exception))
```
kaustubh-karkare: Instead of "invalid", you can use "undefined" or "unexpected" which is a bit more accurate.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,105 @@
...
+                        raise MyParserError("Too many arguments.")
+                        
+                    elif not opt.is_required and size == 1:
+                        self.data[opt.name] = "True"
+                         
+                    else:
+                        if not opt.is_required and size == 2:
+                            raise MyParserError("Too many arguments.")
+                        elif opt.is_required and size == 1:
+                            raise MyParserError("Too less arguments.")
```
kaustubh-karkare: is_required and does it take any values are 2 different things that you're combining?
Why not have a separate setting called is_flag? that determines whether the value should be provided.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,107 @@
...
+                            data[opt.name] = "True"
+                        elif opt.is_flag:
+                            raise MyParserError("Too less arguments.")
+                            
+                         
+                    elif size == 2:
+                        if not opt.is_flag:
+                            raise MyParserError("Too many arguments.")
+                        elif opt.is_flag:
+                            self.check_dtype(splitted_word[0], opt.dtype, splitted_word[1], data)
```
kaustubh-karkare: The name of your function is `check_dtype`. It sounds like it performs a check, and then returns true/false. Setting a value in the `data` map is unexpected, based on the name.
```
if self.check_dtype(splitted_word[0], opt.dtype, splitted_word[1]):
    data[splitted_word[0]] = splitted_word[1]
```

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,107 @@
...
+            for opt in self.options:
+                if splitted_word[0] == opt.name:
+              
+                    found_option = True
+
+                    if size > 2:
+                        raise MyParserError("Too many arguments.")
+                        
+                    elif size == 1:
+                        if not opt.is_flag:
```
kaustubh-karkare: Ah, you misunderstood my explanation, but that's okay.
is_flag = don't need extra value, and !is_flag = need 2nd value.
What you have here is the opposite, but this is a fairly easy fix.

solutions/Abhilasha06/Assignment1/my_parser.py
```diff
@@ -0,0 +1,107 @@
...
+                    elif size == 2:
+                        if not opt.is_flag:
+                            raise MyParserError("Too many arguments.")
+                        elif opt.is_flag:
+                            self.check_dtype(splitted_word[0], opt.dtype, splitted_word[1], data)
+                
+                            
+            if not found_option:
+                raise MyParserError("Unexpected field given.")
+
```
kaustubh-karkare: So much easier to read :)

solutions/Abhilasha06/Assignment1/test.py
```diff
@@ -0,0 +1,74 @@
...
+        self.parser.add_option('--key', dtype='int', is_flag=True)
+        self.parser.add_option('--name', dtype='str', is_flag=True)
+        ans = self.parser.check_options(['./test', '--key=12345', '--name=kaustubh'])
+        self.assertEqual({'--key': '12345', '--name': 'kaustubh'}, ans)
+        
+        
+    def test_three_options_defined(self):
+        self.parser.add_option('--key', dtype='int', is_flag=True)
+        self.parser.add_option('--local', dtype='str', is_flag=False)
+        self.parser.add_option('--remote', dtype='str', is_flag=False)
```
kaustubh-karkare: Hmm, this just made me realize that flag arguments should always have a dtype of "bool", right? Because they will be set to "True".

PR #58, opened 2019-09-04
Solution to assignment-4 (assignment-2.1 on hold)

Hemant-Chowdhury:
Please review the pull request.
Commands used in the assignment are compatible with the windows operating system.
The code is taking 3-5s to execute test.py


solutions/assignment-2/build.py
```diff
@@ -0,0 +1,108 @@
...
+        self.message: str = message
+
+
+class Option(object):
+    """Option stores all the possible build options provided in build configuration files"""
+
+    def __init__(self, properties: Dict[str, Any], command_path: str):
+        self.name: str = properties['name']
+        self.command: str = 'cd {} && {}'.format(command_path, properties['command'])
+        self.file: Union[None, Vector] = None
```
kaustubh-karkare: These variable names aren't clear. The bar is: "can I figure out what is in this by looking at just the name?" In this case, the answer is obviously "no", because there is no way to figure out that it is a "vector" based on the name.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,108 @@
...
+        super(BuildAutomationError, self).__init__(message)
+        self.message: str = message
+
+
+class Option(object):
+    """Option stores all the possible build options provided in build configuration files"""
+
+    def __init__(self, properties: Dict[str, Any], command_path: str):
+        self.name: str = properties['name']
+        self.command: str = 'cd {} && {}'.format(command_path, properties['command'])
```
kaustubh-karkare: You can use "pushd" and "popd".

solutions/assignment-2/test1/test.cpp
```diff
@@ -0,0 +1,11 @@
+#include<bits/stdc++.h>
```
kaustubh-karkare: Interesting, looks like this is specifically designed for programming contests. I don't really mind since this is all part of your testing code, but in general, I confirmed with a friend who works with C++ as part of his daily job, and this is non-ideal when building actual applications.

solutions/assignment-2/test.py
```diff
@@ -0,0 +1,14 @@
+import unittest
+import subprocess
+
+
+class TestBuildAutomation(unittest.TestCase):
+
+    def test_test1(self):
+        cmd = 'cd test1 && python build_auto.py clean run'
+        result = subprocess.check_output(cmd, shell=True)
```
kaustubh-karkare: Instead of doing it in the same directory (which limits you to one test at a time), you can copy the necessary files to /tmp/ (I think all languages have some API to create temporary directories) and run the test there. That way, I can run multiple copies of the same test in parallel.

You could even avoid creating the separate files, and just save their content in this file. Eg -
```
FILE_STRUCTURE = {
  'algorithms/sort_bubble.cpp': """<code>""",
  ...
}
# loop through dict, and create files in temp directory
```

solutions/assignment-2/test1/build_auto.py
```diff
@@ -0,0 +1,9 @@
+"""Triggers build automation commands in the directory where this file exists"""
+import sys
+sys.path.append('../')
+from build import Build
+import os
+
+
+build = Build(os.getcwd())
```
kaustubh-karkare: Use argparse to make the current directory configurable.
```
cd / && build --root=~/build_test/ clean run
```

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,108 @@
...
+
+class Build(object):
+    """Build scans a given directory and stores all the possible build options """
+
+    def __init__(self, path: str):
+        self.build_options: Dict[str, Option] = dict()
+        self.scan(path=path, relative_path='')
+
+    def scan(self, path: str, relative_path: str) -> None:
+        for file in os.listdir(path):
```
kaustubh-karkare: As we can see based on the syntax highlighting, `file` is a commonly used existing python function.
You want to call this `filename` to be more accurate and avoid conflicts.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,108 @@
...
+            file_path: str = os.path.join(path, file)
+            if os.path.isdir(file_path):
+                self.scan(path=file_path, relative_path=f'{relative_path}{file}\\')
+            else:
+                if file == 'build.json':
```
kaustubh-karkare: This should be a constant property on your class.
```
class Build:
  BUILD_FILENAME = "build.json"
  ...
```

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,108 @@
...
+                    json_file: Any = open(file_path)
+                    json_dict: Any = json.loads(json_file.read())
```
kaustubh-karkare: https://docs.python.org/3/library/json.html#json.JSONDecodeError

What happens in that case?

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,108 @@
...
+                    for properties in json_dict:
+                        properties['name'] = relative_path + properties['name']
+                        if properties['name'] not in self.build_options:
+                            self.build_options[properties['name']] = Option(properties=properties, command_path=path)
+                        else:
+                            name = properties['name']
+                            raise BuildAutomationError(f'{name} has multifile entities in different build configuration files which conflicts')
```
kaustubh-karkare: Good use of exceptions :)

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def execute(self, options: Vector) -> None:
+        for option in options:
+            try:
+                self.build_options[option].execute(self.build_options)
```
kaustubh-karkare: So if I give multiple commands, you'd have to run `topological_sort` for each of them? Isn't that duplicate computation?

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,108 @@
...
+        except KeyError:
+            raise BuildAutomationError(f'Invalid option "{source}" found')
+        if option.deps is not None:
+            for child in option.deps:
+                if child not in visited:
+                    Graph.depth_first_search(child, visited, indegree, build_options)
+                indegree[child] += 1
+
+    @staticmethod
+    def topological_sort(source: str, build_options: Dict[str, Option]) -> Vector:
```
kaustubh-karkare: In this case, your graph can be thought of as a library, which is independent of your build logic, and so should not be aware of concepts like "build_options". You should only deal with general concepts like nodes and edges.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,108 @@
...
+            except KeyError:
+                raise BuildAutomationError(f'Invalid option "{source}" found')
+            topo_sort.append(next_option.name)
+            if next_option.deps is not None:
+                for child in next_option.deps:
+                    indegree[child] -= 1
+                    if indegree[child] == 0:
+                        que.put(child)
+        for option in indegree:
+            if indegree[option] != 0:   # All the values should have indegree == 0 if the dependencies form a DAG
```
kaustubh-karkare: - Hmm, do you really need to do a toposort of all the whole graph? Why not only focus on the subset of "build rules" (ie - nodes in the graph) that are direct dependencies of the thing you care about?
- I think that will simplify your code right? You wouldn't need both `visited` and `indegree`, you just need one of them.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class Option(object):
+    """Option stores all the possible build options provided in build configuration files"""
+
+    def __init__(self, properties: Dict[str, Any], command_path: str):
+        self.name: str = properties['name']
+        self.command: str = 'cd {} && {}'.format(command_path, properties['command'])
+        self.file: Union[None, Vector] = None
+        self.deps: Union[None, Vector] = None
```
kaustubh-karkare:
```
self.deps: Optional[List[str]] = None
```

solutions/assignment-2/test.py
```diff
@@ -0,0 +1,14 @@
...
+        self.assertEqual(result, b'Tested sort_bubble sort_merge sort_quick')
+
```
kaustubh-karkare: Given the above, you should have multiple tests, for the various failure cases too.
Show what happens when the build.json file is invalid, or there is a circular dependency, or if a deps list contains an unknown name, etc.

kaustubh-karkare (changes requested):
Really good start!
Now that I've seen how you've solved part1, you should start working on part2 of the problem too.

solutions/assignment-2/build.py
Hemant-Chowdhury: I think we need to do toposort because if we only care about the direct dependencies either we will not be able to find the correct order of execution or if we only use one of the `visited `
or `indegree`, we may not be able to identify circular dependencies.
Examples: 
image

solutions/assignment-2/build.py
Hemant-Chowdhury: I did not quite understand how to use `pushd` in my case.
I think since I am using `os.system()` to run each option's command `cd` and `pushd` will give the same results as `os.system` always executes keeping the current directory(where build.py is located) as the active directory.

kaustubh-karkare: Fair enough. 

https://stackoverflow.com/questions/35277128/why-does-not-os-systemcd-mydir-work-and-we-have-to-use-os-chdirmydir-ins

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+        self.stop_running_build_process()
+        self.scan(path=self.build_arguments['root'], relative_path='')
+        self.set_option_dependents_and_file_dependents()
+        self.execute()
+        self.set_watch()
+
+    def parse(self, arguments: List[str]) -> Dict[str, Any]:
+        parser = argparse.ArgumentParser(description='Process build options')
+        parser.add_argument('options', nargs='*', type=str, help='A list of build options')
+        parser.add_argument('--root', type=str, default='\\', help='Root directory where build is to be triggered')
```
kaustubh-karkare: I would think that the default root is the current directory, right?

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+        parser.add_argument('--watch', default='false', type=str, help='Status to run build option on file change')
+        parser.add_argument('--poll', type=float, default=1, help='Time interval for checking the file modification status')
+        parser.add_argument('--ignore', nargs='*', type=str, help='List of files not to be watched')
+        """Breaking the arguments for any '=' attached with key and value"""
```
kaustubh-karkare: Doesn't argparse handle this?

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+            for arg in key_value.split('='):
+                simplified_arguments.append(arg)
+        return vars(parser.parse_args(simplified_arguments))
+
+    def find_path(self, relative_path: str) -> str:
+        path: str = relative_path
+        if os.path.exists(path):
+            return path
+        else:
+            if path.startswith('~'):
```
kaustubh-karkare: Does `os.path.realpath` take care of this?

solutions/assignment-2/graph.py
```diff
@@ -0,0 +1,38 @@
...
+    def depth_first_search(self) -> List[Any]:
+        processed: Dict[Any, int] = {key: -1 for key in self.adjacency_list}
+        dfs_list: List[Any] = list()
+        for node in processed:
+            if processed[node] == -1:
+                self.dfs(node, processed, dfs_list)
+        dfs_list.reverse()
+        return dfs_list
+
+    def dfs(self, node: Any, processed: Dict[Any, int], dfs_list: List[Any]) -> None:
```
kaustubh-karkare: Looks like this is a private method. Add an underscore at the beginning of the name.

solutions/assignment-2/graph.py
```diff
@@ -0,0 +1,38 @@
...
+        self.message: str = message
+
+
+class Graph(object):
+    """Graph supports static methods like depth first search and topological sorting to resolve the dependencies"""
+
+    def __init__(self, adjacency_list: Dict[Any, List[Any]]):
+        self.adjacency_list = adjacency_list
+
+    def depth_first_search(self) -> List[Any]:
```
kaustubh-karkare: Well, this is a "topological sort using draft first traversal", not a "depth first search".

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+        self.build_arguments: Dict[str, Any] = self.parse(arguments)
+        self.build_options: Dict[str, Option] = dict()
+        self.build_files: Dict[str, File] = dict()
+        self.dependency_list: Dict[str, List[str]] = dict()
+        self.build_arguments['root'] = self.find_path(self.build_arguments['root'])
+        self.stop_running_build_process()
+        self.scan(path=self.build_arguments['root'], relative_path='')
+        self.set_option_dependents_and_file_dependents()
+        self.execute()
+        self.set_watch()
```
kaustubh-karkare: It is unexpected that an __init__ method is doing so much more than initialization.
Can you call execute/set_watch from the application part of this file?
```
build = Build(...)
for name in ...:
  build.execute(name)

File: solutions/assignment-2/build.py [hunk 21]
```diff
@@ -0,0 +1,263 @@
...
+                raise BuildAutomationError(f'Invalid relative path: {relative_path}, doesn\'t exists')
+
+    def stop_running_build_process(self) -> None:
+        with open(self.PROCESS_FILENAME, 'r') as process_file:
+            try:
+                process_dict: Any = json.loads(process_file.read())
+            except ValueError:
+                with open(self.PROCESS_FILENAME, 'w') as process_write_file:
+                    process_write_file.write('{}')
+                process_dict = json.loads(process_file.read())
```
kaustubh-karkare: Reading from the file is unnecessary in this case, right? Why not just use `process_dict = {}`?

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+            process = self.build_arguments['root']
+            if process in process_dict:
+                os.system(f'TASKKILL /PID {process_dict[process]} /F')
+
+    def scan(self, path: str, relative_path: str) -> None:
+        for filename in os.listdir(path):
+            file_path: str = path + filename
+            if os.path.isdir(file_path):
+                self.scan(path=f'{file_path}\\', relative_path=f'{relative_path}{filename}\\')
```
kaustubh-karkare: Instead of adding slashes manually, you can use `os.path.sep`

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+            else:
```
kaustubh-karkare: Instead of an else case, you could just add a continue statement to the if-block, and then avoid an extra layer of indenting.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+                if filename == self.BUILD_FILENAME:
```
kaustubh-karkare: Similarly,
```
if filename != self.BUILD_FILENAME:
  continue
```

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+                    try:
+                        json_file: Any = open(file_path)
+                        json_dict: Any = json.loads(json_file.read())
```
kaustubh-karkare:
```
with open(file_path) as json_file:
  json_dict = json.loads(json_file.read())
```

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+                    except ValueError:
+                        raise BuildAutomationError(f'Decoding json build configuration file "{file_path}" failed')
+                    for properties in json_dict:
+                        properties['name'] = os.path.join(relative_path, properties['name'])
+                        if properties['name'] not in self.build_options:
+                            self.build_options[properties['name']] = Option(properties=properties, command_path=path)
+                            if 'deps' not in properties:
+                                self.dependency_list[properties['name']] = []
+                            else:
+                                self.dependency_list[properties['name']] = properties['deps']
```
kaustubh-karkare: This should should be done in the Option.__init__ method, instead of here.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+class Option(object):
+    """Option stores all the possible build options provided in build configuration files"""
+
+    def __init__(self, properties: Dict[str, Any], command_path: str):
+        self.name: str = properties['name']
+        self.command: str = 'cd {} && {}'.format(command_path, properties['command'])
+        self.all_dependent_options: List[Option] = list()   # the order of all the options which needs to be executed for execution of this option
+        self.all_supporter_options: List[Option] = list()
+
+    def execute(self) -> None:
```
kaustubh-karkare: Wait, why not execute the command for this current option too?

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+            except KeyError:
+                raise BuildAutomationError(f'Invalid option "{option}" provided')
+
+    def set_watch(self) -> None:
+        ignored_files: List[str] = list()
+        if self.build_arguments['ignore']:
+            for files in self.build_arguments['ignore']:
+                ignored_files.append(self.find_path(files))
+            for files in ignored_files:
+                print(files)
```
kaustubh-karkare: - `files` is a single filename right? So this name is inaccurate.
- Also, instead of printing to stdout directly, you can do something like:
```
import logging
class Build:
  def __init__(self):
    self.logger = logging.getLogger()
  def set_watch(self):
    ...
    self.logger.info(f"Ignoring file: {filename}")
```
And then the application part can connect the logger to stdout if it wants.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+        self.scan(path=self.build_arguments['root'], relative_path='')
+        self.set_option_dependents_and_file_dependents()
+        self.execute()
+        self.set_watch()
+
+    def parse(self, arguments: List[str]) -> Dict[str, Any]:
+        parser = argparse.ArgumentParser(description='Process build options')
+        parser.add_argument('options', nargs='*', type=str, help='A list of build options')
+        parser.add_argument('--root', type=str, default='\\', help='Root directory where build is to be triggered')
+        parser.add_argument('--watch', default='false', type=str, help='Status to run build option on file change')
```
kaustubh-karkare: default=False, action="store_true", type=bool

so that you don't need to provide a value

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+        elif self.build_arguments['watch'].upper() == 'TRUE':
+            self.mark_running_build_process()
+            while True:  # To be stopped by giving a command line argument --watch=false with --root being the same directory
+                try:
+                    sleep(self.build_arguments['poll'])
+                    for filename in self.build_files:
+                        if filename not in ignored_files:
+                            if self.build_files[filename].is_modified():
+                                self.build_files[filename].rebuild()
+                except KeyboardInterrupt:
```
kaustubh-karkare: You're assuming that KeyboardInterrupt is the only way to stop this.
What if I wanted to do something like:
```
builder = Build()
builder.start_watchng_on_separate_thread(...)
sleep(10)
builder.stop_watching()
```
This will eliminate the need for your process.json complexity, that I guess you needed to make the test pass.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+                print(files)
+        if self.build_arguments['watch'].upper() == 'FALSE':
+            self.cleanup_running_process()
+            return
+        elif self.build_arguments['watch'].upper() == 'TRUE':
+            self.mark_running_build_process()
+            while True:  # To be stopped by giving a command line argument --watch=false with --root being the same directory
+                try:
+                    sleep(self.build_arguments['poll'])
+                    for filename in self.build_files:
```
kaustubh-karkare: This model assumes that the graph structure is fixed.
What if I modified the build.json file to add another dependency?

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+class File(object):
+    """docstring for File"""
+
+    def __init__(self, filename: str):
+        self.filename: str = filename
+        self.last_modified: float = os.path.getmtime(filename)
+        self.dependent_options: List[Option] = list()   # list of options which are directly dependent on this file
+        self.all_dependent_options: List[Option] = list()   # the order of all the options which needs to be executed if this file modifies
```
kaustubh-karkare: Can you make your code work without `self.dependent_options` and `self.all_dependent_options`?
I ask because there are too many copies of the same data in multiple places.

Since each option (which is generally called a "BuildRule") is associated with a file, you can look up the dependencies from the options class, don't need to store a copy here.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+class Option(object):
+    """Option stores all the possible build options provided in build configuration files"""
+
+    def __init__(self, properties: Dict[str, Any], command_path: str):
+        self.name: str = properties['name']
+        self.command: str = 'cd {} && {}'.format(command_path, properties['command'])
+        self.all_dependent_options: List[Option] = list()   # the order of all the options which needs to be executed for execution of this option
+        self.all_supporter_options: List[Option] = list()
```
kaustubh-karkare: I would actually recommend not having this `self.all_supporter_options` list either.
Compute what you need only when needed.

solutions/assignment-2/test.py
```diff
@@ -0,0 +1,632 @@
...
+              },
+              {
+                "name": "run",
+                "deps": ["test", "algorithms\\sort_bubble", "algorithms\\sort_merge", "algorithms\\sort_quick"],
+                "command": "g++ algorithms\\sort_bubble.o algorithms\\sort_merge.o algorithms\\sort_quick.o test.o -o test.exe && .\\test.exe"
+              }
+            ]
+            """
+        }
+        self.create_file_structure(FILE_STRUCTURE, root='tmp')
```
kaustubh-karkare: Instead of `tmp`, use https://docs.python.org/3/library/tempfile.html

solutions/assignment-2/test.py
```diff
@@ -0,0 +1,632 @@
...
+        cmd = r'python build.py --root=~\tmp\ clean run'
+        result = subprocess.check_output(cmd, shell=True)
+        self.assertEqual(result, b'Tested sort_bubble sort_merge sort_quick\r\n')
+        assert(os.path.exists(r".\tmp\test.o"))
```
kaustubh-karkare: self.assertTrue

solutions/assignment-2/test.py
```diff
@@ -0,0 +1,632 @@
...
+        self.delete_file_structure(root='tmp')
+
+    def test_circular_dependency(self):
+        FILE_STRUCTURE = {
+            'dirA\\build.json':
+            r"""
+            [
+              {
+                "name": "A",
+                "deps": ["dirB\\B"],
```
kaustubh-karkare: It isn't clear whether this is an absolute path or a relative one.
Eg - You could do "../dirB/B" or "//dirB/B" (where // signifies root)

solutions/assignment-2/test.py
```diff
@@ -0,0 +1,632 @@
...
+                "command": "echo parent"
+              }
+            ]
+            """
+        }
+        self.create_file_structure(FILE_STRUCTURE, root='tmp')
+
+        cmd = r'python build.py --root=~\tmp\ parent'
+        output = subprocess.check_output(cmd, shell=True)
+        self.assertEqual(output, b'Error: Undefined dependencies found "B".\r\n')
```
kaustubh-karkare: The message can explain which build rule tried to depend on it.

solutions/assignment-2/test.py
```diff
@@ -0,0 +1,632 @@
...
+    def test_multilevel_dependency(self):
+        """
+                parent
+                /   \
+               A     B
+                \   /
+                  C
+                /   \
+               D     E
+        """
```
kaustubh-karkare: Nice!

solutions/assignment-2/test.py
```diff
@@ -0,0 +1,632 @@
+import os
```
kaustubh-karkare: Btw, this file is too large. You can split it down into smaller ones.

solutions/assignment-2/test.py
```diff
@@ -0,0 +1,632 @@
...
+                """)
+        sleep(10)
+        with open(r'tmp\algorithms\sort_merge.cpp', 'w') as sort_merge:
+            sort_merge.write(r"""
+                #include<iostream>
+                void sort_merge(){
+                std::cout<<"sort_merge_modified";
+                }
+                """)
+        sleep(10)
```
kaustubh-karkare: Tests with sleep statements are non-ideal, since they make you wait.
Can you modify your code so that something like this is possible?
```
# assuming something like:
def watch():
  with True:
    self.check_and_maybe_execute()
```
And then you call `check_and_maybe_execute` from this file, eliminating any wait.

solutions/assignment-2/test.py
```diff
@@ -0,0 +1,632 @@
...
+        build_2.start()
+
+        build_1.join()
+        build_2.join()
+
+        self.assertEqual(build_1.output, b'Tested sort_bubble sort_merge sort_quick\r\nTested sort_bubble_modified sort_merge sort_quick\r\nTested sort_bubble_modified sort_merge_modified sort_quick\r\n<Modified> Tested sort_bubble_modified sort_merge_modified sort_quick\r\n')
+        self.delete_file_structure(root='tmp')
+
+    def test_ignore_files(self):
+        FILE_STRUCTURE = {
```
kaustubh-karkare: Since this has been repeated a few times, make it a class level variable and share it? You can use it like a template, and add modifications on top of a deepcopy.

kaustubh-karkare (changes requested):
Excellent progress!

solutions/assignment-2/build.py
Hemant-Chowdhury: I can make these changes but then I will not be able to run the program through command line like what I did in the tests `os.system('python build.py --root=.\tmp\ clean run')`. 
Actually, I tried to create build automation which is able to work from end to end. So, supposing the tests being done by a user, it only needs to give a command to start the build, and in case if he wants to stop it just needs to give another command to stop, a user cannot have access to the build object. Now suppose the user does not know what build processes are running in the first place, he will blindly give another command to build and watch a directory which is already in the process, so in my implementation, I had to take case to stop the already running process and start the new process as demanded by the user, which I did by keeping track of running process in process.json file.
I am a little bit confused, should I take care of these implementations( Required from a user point of view ) or just create a library which has methods to control the operation using the build object. I am curious how actually it is implemented.

solutions/assignment-2/build.py
Hemant-Chowdhury: `self.dependent_options` and `self.all_dependent_options` can be removed but then I need to compute this again whenever a file is modified. Actually, I introduced this to avoid duplicate computation (as previously suggested). So, the question is how to decide between the two possible implementations in real projects, both come with its advantages and disadvantages?

solutions/assignment-2/build.py
kaustubh-karkare: There's no requirement to run it through `os.system` though. You can think of your Build class as a library, which is separate from the application part (which basically just connects the library to the CLI input & keyboard interrupts). And then you can test the Build class directly, without going though CLI/keyboard/kill commands stuff.

> I am curious how actually it is implemented.

The example I've seen is that multiple processes try to acquire the same lock, so the 2nd one is blocked until the first one terminates. It was kinda beyond the scope of what I intended for you to solve though, and while I appreciate it, don't feel obligated to do it :)

solutions/assignment-2/build.py
kaustubh-karkare: > Actually, I introduced this to avoid duplicate computation (as previously suggested)

Hmm, I had to look to look up what you were referring to, and I think you meant the comment that said "So if I give multiple commands, you'd have to run topological_sort for each of them? Isn't that duplicate computation?" Good point, I am going back on my word, yes, but it is because the probably statement has changed (ie - now you're solving part2 also).

The basic issue is that any time you cache data, it is going to become inconsistent, and so you'd need to handle that. 

> but then I need to compute this again whenever a file is modified

This is a good thing, because the dependencies might have changed, right?

> how to decide between the two possible implementations in real projects, both come with its advantages and disadvantages?

It varies. My preference is to avoid premature optimization, focus on simplicity of code, so that it is easy to maintain, and only add complexity if it is truly needed.

solutions/assignment-2/file_structure.py
```diff
@@ -0,0 +1,267 @@
...
+
+    @staticmethod
+    def get_file_structure_having_circular_dependency():
+        FILE_STRUCTURE = {
+            '.\\dirA\\build.json':
+            r"""
+            [
+              {
+                "name": "runA",
+                "deps": ["dirB\\runB"],
```
Hemant-Chowdhury: Earlier you mentioned that `"dirB\\runB"` is quite confusing and it does not tell whether it is an absolute path or relative path.
Here, I have assumed that all the entries which come in the `"deps"`  is a relative path, and `\\` is used to handle the escape character `\`, since I am using Windows os, separators for os.path is `\` 
Any changes to `"dirB\\runB"` like `".\dirB\\runB"` or `"\\dirB\runB"` was difficult because entries in `"deps"` should match with some `"name"`.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,191 @@
...
+    """docstring for File"""
+
+    def __init__(self, filename: str, is_BUILD_FILE: bool=False):
+        self.filename: str = filename
+        self.last_modified: float = os.path.getmtime(filename)
+        self.is_BUILD_FILE: bool = is_BUILD_FILE
+
+    def is_modified(self) -> bool:
+        if os.path.getmtime(self.filename) != self.last_modified:
+            self.last_modified = os.path.getmtime(self.filename)
```
kaustubh-karkare: Call `os.path.getmtime` only once, since it is unnecessary duplication computation.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,191 @@
...
+class Rule(object):
+    """rule stores all the possible build rules provided in build configuration files"""
+
+    def __init__(self, properties: Dict[str, Any], command_path: str):
+        self.name: str = properties['name']
+        self.command: str = 'cd {} && {}'.format(command_path, properties['command'])
+        self.filenames: List[str] = list()
+        self.dependent_rulenames: List[str] = list()
+        if 'deps' in properties:
+            self.dependent_rulenames = properties['deps']
```
kaustubh-karkare: self.dependent_rulenames = properties.get('deps', [])

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,191 @@
...
+        self.stop_watch = True
+
+    def _parse(self, arguments: List[str]) -> Dict[str, Any]:
+        parser = argparse.ArgumentParser(description='Process build rules')
+        parser.add_argument('rules', nargs='*', type=str, help='A list of build rules')
+        parser.add_argument('--root', type=str, default=r'.\\', help='Root directory where build is to be triggered')
+        parser.add_argument('--watch', action='store_true', help='Status to build again on file modification')
+        parser.add_argument('--poll', type=float, default=1, help='Time interval for checking the file modification status')
+        parser.add_argument('--ignore', nargs='*', type=str, help='List of files not to be watched')
+        return vars(parser.parse_args(arguments))
```
kaustubh-karkare: What is the benefit of using `vars` here?

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,191 @@
...
+    """Build scans a given directory and stores all the possible build rules """
+
+    BUILD_FILENAME = "build.json"
+
+    def __init__(self, arguments: List[str]):
+        self.logger = logging.getLogger()
+        self.build_rules: Dict[str, Rule] = dict()
+        self.build_files: Dict[str, File] = dict()
+        self.build_arguments: Dict[str, Any] = self._parse(arguments)
+        self.build_arguments['root'] = os.path.realpath(self.build_arguments['root'])
```
kaustubh-karkare: Instead of doing this in a separate phase, you can try:
```
parser.add_argument('--root', type=lambda path: os.path.realpath(path), ...
```

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def __init__(self, message: str):
+        self.message: str = 'Error: ' + message
+        super(BuildAutomationError, self).__init__(self.message)
+
+
+class File(object):
+    """docstring for File"""
+
+    def __init__(self, filename: str, is_BUILD_FILE: bool=False):
```
kaustubh-karkare: Confusing use of capital letters here ... why not just use "is_build_file"?

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,191 @@
...
+                raise BuildAutomationError(f'Invalid rule "{rulename}" provided')
+            dependent_rulenames = self.get_dependent_rulenames_for_executing_rule(rulename)
+            for dependent_rulename in dependent_rulenames:
+                output += self.build_rules[dependent_rulename].execute()
+        return output
+
+    def set_ignored_files(self) -> None:
+        self.ignored_files: List[str] = list()
+        if self.build_arguments['ignore']:
+            for file in self.build_arguments['ignore']:
```
kaustubh-karkare: s/file/filepath/ is a better variable name

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,191 @@
...
+        for rulename in self.build_arguments['rules']:
+            if rulename not in self.build_rules:
+                raise BuildAutomationError(f'Invalid rule "{rulename}" provided')
+            dependent_rulenames = self.get_dependent_rulenames_for_executing_rule(rulename)
+            for dependent_rulename in dependent_rulenames:
+                output += self.build_rules[dependent_rulename].execute()
+        return output
+
+    def set_ignored_files(self) -> None:
+        self.ignored_files: List[str] = list()
```
kaustubh-karkare: In general, use sets when ordering does not matter.

solutions/assignment-2/graph.py
```diff
@@ -0,0 +1,35 @@
...
+        self.message: str = message
+
+
+class Graph(object):
+    """Graph deals with nodes and edges and provides methods like depth first search to resolve the dependencies"""
+
+    def __init__(self, adjacency_list: Dict[Any, List[Any]]):
+        self.adjacency_list = adjacency_list
+
+    def draft_first_search(self, source: Any) -> List[Any]:
```
kaustubh-karkare: This is not a search, it is just a traversal starting from this "source" node.

solutions/assignment-2/graph.py
```diff
@@ -0,0 +1,35 @@
...
+        processed: Dict[Any, int] = {key: -1 for key in self.adjacency_list}
+        dfs_list: List[Any] = list()
+        self._dfs(source, processed, dfs_list)
+        dfs_list.reverse()
```
kaustubh-karkare: You reversed this list here, and also in the other file where you called this method. Unnecessary work?

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,191 @@
...
+        except GraphError as e:
+            raise BuildAutomationError(e.message)
+        return all_dependencies
+
+    def get_dependent_rulenames_for_executing_changes_in_file(self, filename: str) -> List[str]:
+        """Generating one supporter list for all the rules and files"""
+        supporter_list: Dict[str, List[str]] = {key: list() for key in self.build_rules}
+        for file_name in self.build_files:
+            if not self.build_files[file_name].is_BUILD_FILE:
+                supporter_list[file_name] = list()
```
kaustubh-karkare: Wait, so the keys of `supporter_list` include both rule names and file names? Mixing things like that is probably not a good idea, and it looks like this only matters in the first step of your traversal (filename -> rule name), so you can do that manually.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,191 @@
...
+        return all_supporters
+
+    def execute(self) -> str:
+        output = ''
+        for rulename in self.build_arguments['rules']:
+            if rulename not in self.build_rules:
+                raise BuildAutomationError(f'Invalid rule "{rulename}" provided')
+            dependent_rulenames = self.get_dependent_rulenames_for_executing_rule(rulename)
+            for dependent_rulename in dependent_rulenames:
+                output += self.build_rules[dependent_rulename].execute()
```
kaustubh-karkare: See Part2 of updated problem statement.

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,191 @@
...
+    def set_ignored_files(self) -> None:
+        self.ignored_files: List[str] = list()
+        if self.build_arguments['ignore']:
+            for file in self.build_arguments['ignore']:
+                self.ignored_files.append(os.path.join(self.build_arguments['root'], os.path.normpath(file)))
+
+    def check_and_maybe_execute(self) -> str:
+        output = ''
+        for filename in self.build_files:
+            if filename not in self.ignored_files:
```
kaustubh-karkare: Return/continue/break early to reduce indenting.
```
if filename in self.ignored_files:
  continue
```

solutions/assignment-2/build.py
```diff
@@ -0,0 +1,191 @@
...
+        if 'deps' in properties:
+            self.dependent_rulenames = properties['deps']
+        if 'files' in properties:
+            for filename in properties['files']:
+                self.filenames.append(os.path.join(command_path, filename))
+
+    def execute(self) -> str:
+        process = subprocess.Popen(self.command.split(), stdout=subprocess.PIPE, shell=True)
+        (out, err) = process.communicate()
+        return out.decode()
```
kaustubh-karkare: Hmm, I don't think that the stdout of the process is a reliable indicator of success. I'd recommend using return value.

kaustubh-karkare (changes requested):
Looking good!

* There are lots of different things going on in you build.py file. You should break it down into pieces: one file for the classes for rules, another file for executing them, another to watch for changes.
* Don't separate your file structure from your tests, since they're conceptually linked. As a reader, it is ideal if all the information relevant to a single test is in one location. Feel free to separate the tests across multiple files to keep them easily readable and use a test_utils.py module to share code where needed.
* I updated the problem statement to add another requirement: parallel execution of rules without multithreading. Lets do that too :)

solutions/assignment-4.1/Object.py
```diff
@@ -0,0 +1,173 @@
...
+        headfile.write(commit_hash)
+
+
+# class branch(object):
+
+#     def __init__(self):
+#         pass
+
+
+class commit(object):
```
kaustubh-karkare: https://www.python.org/dev/peps/pep-0008/#class-names

solutions/assignment-4.1/Object.py
```diff
@@ -0,0 +1,173 @@
...
+        self.message: str = ''
+        self.timestamp: str = ''
+        if commit_hash:
+            self.set_hash(commit_hash)
+
+    def save_object(self) -> None:
+        with open(os.path.join(vcs_files.OBJECTS, self.get_hash()), 'w') as commit_file:
+            commit_file.write(self.get_content())
+
+    def set_hash(self, hash: str) -> None:
```
kaustubh-karkare: This method does everything except what the name says.

solutions/assignment-4.1/test.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def clean_directory(self):
+        for filename in os.listdir():
+            if os.path.isdir(filename):
+                shutil.rmtree(filename)
+                continue
+            os.remove(filename)
+
+    def setUp(self):
+        test_dir = 'test_dir'
```
kaustubh-karkare: https://docs.python.org/3.7/library/tempfile.html

solutions/assignment-4.1/test.py
```diff
@@ -0,0 +1,254 @@
...
+        if os.path.exists(test_dir):
+            shutil.rmtree(test_dir)
+        os.mkdir(test_dir)
+        os.chdir(test_dir)
+
+    def test_init(self):
+        result = Init.init()
```
kaustubh-karkare: Why not use your vcs class here?

solutions/assignment-4.1/Init.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __init__(self):
+        self.status = None
+        if os.path.exists(vcs_files.ROOT):
+            self.status = 'Reinitializing vcs in this directory'
+        else:
+            self.status = 'Initializing vcs in this directory'
+        self.check_and_create_files()
+
+    def check_and_create_files(self):
```
kaustubh-karkare: This code works in case an incomplete set of those folders/files already exists.
I'm not sure if that's a good idea. Ie - if you find something partial, I would imagine it is safer to raise an exception, and ask the user to delete the things.

solutions/assignment-4.1/Commit.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def __init__(self, message=''):
+        self.status = None
+        if Object.get_HEAD() is not None:
+            if Status.status().any_modification_exists() is False:
+                self.status = 'Clean working directory\nNothing to commit'
+                return
+        commit_object = Object.commit()
+        commit_object.set_parent(Object.get_HEAD())
+        tree_object = self._save_object(os.getcwd())
```
kaustubh-karkare: What if the repo contains multiple nested folders?
Will the commit only capture the current one?

solutions/assignment-4.1/Diff.py
```diff
@@ -0,0 +1,122 @@
...
+        self.common_line_list: List[str] = self._lcs()
+        self._difference_details()
+
+    def _set_file_lines(self, new_file, old_file):
+        with open(new_file, 'r') as read_new_file:
+            self.new_file_lines = read_new_file.read().split('\n')
+        with open(old_file, 'r') as read_old_file:
+            self.old_file_lines = read_old_file.read().split('\n')
+
+    def _lcs(self):
```
kaustubh-karkare: This must have been fun to implement :)

solutions/assignment-4.1/test.py
```diff
@@ -0,0 +1,254 @@
...
+import Init
+import Commit
+import Diff
+import Status
+import Reset
+import Checkout
+import unittest
+
+
+class TestVCS(unittest.TestCase):
```
kaustubh-karkare: This is doing too many things IMO.
For the sake of readability, I would recommend multiple smaller test files, so that you can test your utility methods like LCS specifically too, not just the commands.

PR #60, opened 2019-09-09
Assignment 2.1: Build Automation Tool

ankurdubey521:
So midway through I decided I wanted to do this assignment in Python, mainly because 
1) my skills in python are quite rusty and based on a lot of googling
2) was having quite a lot of trouble with setting up execSync in node
I therefore request you to be stricter with your analysis of the code


solutions/ankurdubey521/Build Automation Tool/.idea/Build Automation Tool.iml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kaustubh-karkare: All these xml files should not be part of your PR. Add them to your .gitignore file, since they are relevant only for you.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_buildconfig.py
```diff
@@ -0,0 +1,15 @@
+import unittest
+from Builder.lib.buildconfig import BuildConfig
+
+
+class TestJsonParser(unittest.TestCase):
+    def test_sample_json_parser_correctly(self):
+        json_path = 'json/test1.json'
```
kaustubh-karkare: The input to your BuildConfig class should be a directory, not a specific file.
The file name "build.json" is fixed, so "test1.json" violates that pattern.

You can add the "build.json" suffix in the BuildConfig class, instead of relying on the caller to add it.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_buildconfig.py
```diff
@@ -0,0 +1,15 @@
...
+        config = BuildConfig(json_path)
+        self.assertEqual(config.command_names(), ['clean', 'test'])
+        self.assertEqual(config.deps('clean'), ["algorithms/clean"])
+        self.assertEqual(config.command('clean'), "rm -f test.o && rm -f test.exe")
```
kaustubh-karkare: Interesting, these utility methods can definitely be useful, but I'd recommend better naming here.
In this case, adding a "get_" prefix to all of them will make it clear that these are not "set_" or "has_" or "check_" or whatever else methods.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_builder.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+import subprocess
+from Builder.main import execute
+
+
+class TestBuilder(unittest.TestCase):
+    def test_compilation(self):
+        path = '/home/ankurdubey/Code/Github/project-omega/solutions' \
```
kaustubh-karkare: This path is non-transferrable. If I downloaded this code in my machine, this test would fail, which is bad.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,41 @@
+import json
+
+
+class BuildConfig:
+    """Parses and Stores build.config"""
+    def __init__(self, json_path):
+        # Parse JSON
+        json_file = open(json_path)
+        self._raw_json = json.load(json_file)
+        json_file.close()
```
kaustubh-karkare: You should use something like:
```
with open(json_path) as file_handle:
  _ = json.load(file_handle)
```

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,41 @@
...
+
+        # Validate the config file
+        self.validate()
+
+        self._command = {}
```
kaustubh-karkare: Proper naming is important:
* _command suggests a single item.
* _commands suggests a list of items.
* I would use `self.name_to_command` because it looks like a map.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,41 @@
...
+        """Return List of Dependencies"""
+        return self._command[command_name]['deps']
+
+    def command(self, command_name):
+        """Returns Command String"""
+        return self._command[command_name]['command']
+
+    def files(self, command_name):
+        """Returns List of Associated Files"""
+        return self._command[command_name]['files']
```
kaustubh-karkare: Create a `Command` class, and then move these getters there.
That will leave you with a single getter here: "get_command(name)"

solutions/ankurdubey521/Build Automation Tool/Builder/lib/commandrunner.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def run(command_string, cwd, print_command):
+    """Run Command and Return Exit Code. Optionally Print the command itself"""
+    if print_command:
+        print(command_string)
+    return subprocess.run(command_string, shell=True, cwd=cwd).returncode
+
+
+if __name__ == '__main__':
```
kaustubh-karkare: This section does nothing. Remove it.

solutions/ankurdubey521/Build Automation Tool/Builder/main.py
```diff
@@ -0,0 +1,42 @@
...
+
+def execute(command, containing_folder_path):
+    """Parses JSON, Resolves Dependencies and Executes Command"""
+
+    # Parse build.json in current directory
+    config = BuildConfig(containing_folder_path + "/build.json")
+
+    # Parse Dependencies First
+    try:
+        deps = config.deps(command)
```
kaustubh-karkare: What if I wanted to run "build algorithms/sort_bubble"?
Since that is not in the current directory, this code would not find it (even though you have the relative path).

solutions/ankurdubey521/Build Automation Tool/Builder/main.py
```diff
@@ -0,0 +1,42 @@
...
+        dep_count = len(deps)
+        print("Executing {} dependencies for {} in {}...".format(dep_count, command, containing_folder_path))
+        for dep in deps:
+            try:
+                # Command in Child Folder
+                dep_containing_folder_path, dep_command = dep.rsplit('/', 1)
+                dep_containing_folder_path = containing_folder_path + "/" + dep_containing_folder_path
+                execute(dep_command, dep_containing_folder_path)
+            except ValueError:
```
kaustubh-karkare: ValueError can be generated by so many things that I have no idea when this will happen. Maybe you want to use a more specific exception?

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_commandrunner.py
```diff
@@ -0,0 +1,18 @@
...
+        self.assertEqual(0, exit_code)
+
+    def test_nonzero_exit_code(self):
+        command = "exit 1"
+        exit_code = run(command, print_command=False, cwd='/')
+        self.assertEqual(1, exit_code)
+
+
+if __name__ == '__main__':
+    unittest.main()
```
kaustubh-karkare: You have tests for the individual components, but not for the overall system (ie - an "integration test").

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_commandrunner.py
```diff
@@ -0,0 +1,18 @@
...
+
+class TestCommandRunner(unittest.TestCase):
+    def test_basic_command(self):
+        command = "echo 'Hello World!'"
+        exit_code = run(command, print_command=False, cwd='/')
+        self.assertEqual(0, exit_code)
+
+    def test_nonzero_exit_code(self):
+        command = "exit 1"
+        exit_code = run(command, print_command=False, cwd='/')
```
kaustubh-karkare: > print_command=False

This should be a default argument value.

solutions/ankurdubey521/Build Automation Tool/Builder/main.py
```diff
@@ -0,0 +1,42 @@
...
+    try:
+        deps = config.deps(command)
+        dep_count = len(deps)
+        print("Executing {} dependencies for {} in {}...".format(dep_count, command, containing_folder_path))
+        for dep in deps:
+            try:
+                # Command in Child Folder
+                dep_containing_folder_path, dep_command = dep.rsplit('/', 1)
+                dep_containing_folder_path = containing_folder_path + "/" + dep_containing_folder_path
+                execute(dep_command, dep_containing_folder_path)
```
kaustubh-karkare: You need to add checks for circular dependencies.

kaustubh-karkare (changes requested):

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,75 @@
...
+            entry = self._name_to_command[command_name]
+            name = entry['name']
+            command_string = entry['command']
+            deps = None
+            if 'deps' in entry:
+                deps = entry['deps']
+            files = None
+            if 'files' in entry:
+                files = entry['files']
+            return Command(name=name, command_string=command_string, deps=deps, files=files)
```
kaustubh-karkare: Instead of creating a Command object on demand, why not do all this in your __init__ method, so that it only needs to be done once?

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,75 @@
+import json
+
+
+class Command:
+    """Stores Command Information"""
+    def __init__(self, *, name, command_string, deps=None, files=None):
+        self._name = name
+        self._command_string = command_string
+        if files is None:
+            self._hasFiles = False
```
kaustubh-karkare: Do you really a separate variable for _hasFiles? Why not just check if _files is None? In languages like C++, where a list variable cannot be null, this makes more sense, but not sure it is needed in python.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,75 @@
...
+            return self._files
+        raise Command.NoFilesException("No Files for {}".format(self._name))
+
+    def get_command_string(self):
+        return self._command_string
+
+    def get_dependencies(self):
+        if self._hasDependencies:
+            return self._dependencies
+        raise Command.NoDependenciesException("No Dependencies for {}".format(self._name))
```
kaustubh-karkare: I don't even think you need to raise exceptions like this. Ie - if deps aren't specified, the default meaning is that there are no deps, so you can just return an empty list.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,75 @@
...
+            command_string = entry['command']
+            deps = None
+            if 'deps' in entry:
+                deps = entry['deps']
+            files = None
+            if 'files' in entry:
+                files = entry['files']
+            return Command(name=name, command_string=command_string, deps=deps, files=files)
+        else:
+            raise BuildConfig.UnknownCommandException('No such command "{}" found.'.format(command_name))
```
kaustubh-karkare, line 87: I liked this model of associating exceptions with class. Hadn't seen it before somehow.

solutions/ankurdubey521/Build Automation Tool/Builder/main.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Parse build.json in current directory
+        config = BuildConfig(command_dir_abs)
+        command = config.get_command(command_name)
+
+        # Process Dependencies
+        try:
+            deps = command.get_dependencies()
+            if not dry_run:
+                print("Executing {} dependencies for {} in {}...".format(len(deps), command_name, command_dir_abs))
```
kaustubh-karkare: This doesn't look too useful, since it doesn't include dependencies.
Can we make the dry_run mode print out all the rules/commands it would have executed?
Output like:
```
> build run --dry-run
[algorithms/sort_bubble] g++ -c sort_bubble.cpp
[algorithms/sort_merge] g++ -c sort_merge.cpp
[algorithms/sort_quick] g++ -c sort_quick.cpp
[test] g++ -std=c++11 -c test.cpp
[run] g++ algorithms/sort_bubble.o algorithms/sort_merge.o algorithms/sort_quick.o test.o -o test.exe && ./test.exe
```

solutions/ankurdubey521/Build Automation Tool/Builder/main.py
```diff
@@ -0,0 +1,91 @@
...
+                    dep_name = dep
+                    dep_dir_abs = command_dir_abs
+
+                # Check for circular dependencies
+                if (dep_name, dep_dir_abs) in self.unresolved_commands:
+                    raise self.CircularDependencyException(
+                        "Detected a Circular Dependency between {}:{} and {}:{}"
+                        .format(dep_dir_abs, dep_name, command_dir_abs, command_name))
+
+                self._build_rule_handler(dep_name, dep_dir_abs, root_dir_abs, dry_run)
```
kaustubh-karkare: Given that your `root_dir_abs` is passed around unchanged, why not save it as a class variable?
Also, you can make it configurable: `build run --root-dir ~/project/`

solutions/ankurdubey521/Build Automation Tool/Builder/main.py
```diff
@@ -0,0 +1,91 @@
...
+        self._build_rule_handler(command_name, command_dir_abs, root_dir_abs, dry_run=True)
+
+        # If everything is fine do an actual run
+        self._build_rule_handler(command_name, command_dir_abs, root_dir_abs)
+
+    class CircularDependencyException(Exception):
+        pass
+
+
+if __name__ == '__main__':
```
kaustubh-karkare, line 8: It might still be useful to keep this __main__ application logic in a separate file from your library.
I'd also recommend catching your exceptions and only showing the message, instead of the whole stack trace.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/commandrunner.py
```diff
@@ -0,0 +1,9 @@
+import subprocess
+
+
+def run(command_string, cwd, print_command=False):
```
kaustubh-karkare: I'd just make this a method in your Builder class below.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_buildconfig.py
```diff
@@ -0,0 +1,37 @@
+import unittest
+from Builder.lib.buildconfig import Command, BuildConfig
+
+
+class TestJsonParser(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
```
kaustubh-karkare: Alternatively,
```
def setUp(self):
  self.config = BuildConfig('json')
```
avoids dealing with class level stuff.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_builder.py
```diff
@@ -0,0 +1,48 @@
...
+import subprocess
+import os
+from Builder.main import Builder
+
+
+class TestBuilder(unittest.TestCase):
+
+    def test_compilation_basic(self):
+        builder = Builder()
+        path = os.getcwd() + '/test_builder_files/test_compilation_basic'
```
kaustubh-karkare: This assumes that you're in a very specific directory.
Maybe something like https://stackoverflow.com/questions/5137497/find-current-directory-and-files-directory makes it possible to run a test from any location?

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_builder.py
```diff
@@ -0,0 +1,48 @@
...
+        builder.execute_build_rule('run', path, path)
+        exec_path = '"' + path + '/test.out' + '"'
+        result = subprocess.run(exec_path, shell=True, capture_output=True, text=True)
+        self.assertEqual('1 2 3 4 5 \n1 2 3 4 5 \n1 2 3 4 5 \n', result.stdout)
+        builder.execute_build_rule('clean', path, os.getcwd())
```
kaustubh-karkare: Verify that test.out is deleted.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_builder.py
```diff
@@ -0,0 +1,48 @@
...
+            result = file_handle.readable()
+        builder.execute_build_rule('clean', path, os.getcwd() + '/test_builder_files/test_commands_referenced_from_root')
+        self.assertEqual(True, result)
+
+    def test_dry_run_does_not_write_files(self):
+        builder = Builder()
+        path = os.getcwd() + '/test_builder_files/test_commands_referenced_from_root'
+        builder._build_rule_handler('run', path, path, dry_run=True)
+        output_file_path = path + '/output'
+        self.assertEqual(False, os.path.isfile(output_file_path))
```
kaustubh-karkare: self.assertFalse(

kaustubh-karkare (changes requested):
Nice!

solutions/ankurdubey521/Build Automation Tool/.gitignore
```diff
@@ -0,0 +1,2 @@
+/venv/
+.idea
```
kaustubh-karkare: btw, it is possible to add .gitignore here to exclude this file too.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,68 @@
+import json
+from typing import List
+
+
+class Command:
+    """Stores Command Information"""
+    def __init__(self, *, name: str, command_string: str, deps: List[str] = None, files: List[str] = None) -> None:
```
kaustubh-karkare: Given the non-intuitive distinction between `Command` and `command_string`, call the class `BuildRule`?

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
kaustubh-karkare: ^ What do you think of this?

solutions/ankurdubey521/Build Automation Tool/Builder/lib/builder.py
```diff
@@ -0,0 +1,89 @@
...
+        self.unresolved_commands = set()
+
+    def _build_rule_handler(self, command_name: str, command_dir_abs: str, dry_run: bool = False) -> None:
+        """Parses JSON, Resolves Dependencies and Executes Command/Do a dry run"""
+
+        # Remove Trailing '/' from paths if present
+        if command_dir_abs.endswith('/'):
+            command_dir_abs = command_dir_abs[:-1]
+        if self._root_dir_abs.endswith('/'):
+            self._root_dir_abs = self._root_dir_abs[:-1]
```
kaustubh-karkare: I usually prefer to do this validation once at the beginning, instead of during every call to this method.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/file_watcher.py
```diff
@@ -0,0 +1,30 @@
...
+from time import sleep
+
+
+class FileWatcher:
+    FILE_WATCH_INTERVAL_SECONDS = 1
+
+    def _get_file_edit_times(self, file_list: List[str]) -> Dict[str, int]:
+        """Populate a Dict of last edit times of files and return it"""
+        file_edit_times = {}
+        for file in file_list:
```
kaustubh-karkare: Slightly more descriptive variable name:
```
for file_path in file_list:
```

solutions/ankurdubey521/Build Automation Tool/Builder/lib/file_watcher.py
```diff
@@ -0,0 +1,30 @@
...
+        print("Listening for changes on {}...".format(file_list))
+        file_edit_times = self._get_file_edit_times(file_list)
+        print("Executing command for the first time...")
+        function(*args)
+        while True:
+            sleep(FileWatcher.FILE_WATCH_INTERVAL_SECONDS)
+            new_file_edit_times = self._get_file_edit_times(file_list)
+            if new_file_edit_times != file_edit_times:
+                file_edit_times = new_file_edit_times
+                print("Detected change of file. Executing command...")
```
kaustubh-karkare: Hmm. How does this method know what we're doing ("Executing command ...") if the function is defined as generic? Access to that information broke an abstraction boundary, right?

solutions/ankurdubey521/Build Automation Tool/Builder/main.py
```diff
@@ -0,0 +1,48 @@
...
+    # Parse Command
+    command = args.build_rule
+    root_dir = ""
+    if args.root_dir is not None:
+        root_dir = args.root_dir
+    else:
+        root_dir = os.getcwd()
+
+    if root_dir.endswith('/'):
+        root_dir = root_dir[:-1]
```
kaustubh-karkare: Huh, if you already did this check here, why did you need to do it in the other file?

solutions/ankurdubey521/Build Automation Tool/Builder/main.py
```diff
@@ -0,0 +1,48 @@
...
+        relative_path, command = command.rsplit('/', 1)
+    path = root_dir + "/" + relative_path
+
+    try:
+        if not args.watch:
+            builder.execute_build_rule(command, path, root_dir)
+        else:
+            # Watch files and run build rule on changes
+            file_watcher = FileWatcher()
+            file_list = BuildConfig(path).get_command(command).get_files()
```
kaustubh-karkare: So now you know why the example json included information about files :)
However, should we be including the files for dependencies of the specified rule too?

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_builder.py
kaustubh-karkare: Interesting, you addressed this concern, but in a way that I didn't expect.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_file_watcher.py
```diff
@@ -0,0 +1,52 @@
...
+        os.chdir('Builder/tests')
+
+    def test_copy_file_on_file_change(self):
+        # Initialize Files
+        file_path = os.getcwd() + "/test_file_watcher_files/test_copy_file_on_file_change/"
+        input_file = file_path + "input.txt"
+        output_file = file_path + "output.txt"
+        with open(input_file, 'w') as file:
+            file.write("Hello World 1.0")
+        with open(output_file, 'w') as file:
```
kaustubh-karkare: s/file/file_handle/

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_file_watcher.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class MyTestCase(unittest.TestCase):
+    # The tests should work for any path inside the project
+    def setUp(self):
+        while os.path.basename(os.getcwd()) != 'Build Automation Tool':
+            os.chdir('..')
+        os.chdir('Builder/tests')
+
+    def test_copy_file_on_file_change(self):
```
kaustubh-karkare, line 14: Interesting, this proves that your file watcher works.
I'm wondering if there is value in demonstrating that actual build commands also work as expected.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_file_watcher.py
```diff
@@ -0,0 +1,52 @@
...
+        input_file = file_path + "input.txt"
+        output_file = file_path + "output.txt"
+        with open(input_file, 'w') as file:
+            file.write("Hello World 1.0")
+        with open(output_file, 'w') as file:
+            file.write("Hello World 1.0")
+
+        # Activate Watcher and change tracked file
+        file_watcher = FileWatcher()
+        process = Process(target=file_watcher.watch_and_execute, args=([input_file], copy_file, input_file, output_file))
```
kaustubh-karkare: Interesting, you're using a separate process here.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/builder.py
```diff
@@ -0,0 +1,89 @@
+from Builder.lib.buildconfig import BuildConfig, Command
+import subprocess
+
+
+class Builder:
+    def __init__(self):
```
kaustubh-karkare: The way you're using this, the constructor must be followed by a call to execute_build_rule. That kind of expectation is unexpected, and while in theory, you could explain it in documentation, there is a better way:

Make execute_build_rule a @classmethod, which creates the object, and then calls _build_rule_handler too. That way, you only need to call one method from outside this file.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/file_watcher.py
```diff
@@ -0,0 +1,30 @@
...
+        """Populate a Dict of last edit times of files and return it"""
+        file_edit_times = {}
+        for file in file_list:
+            file_edit_times[file] = stat(file).st_mtime
+        return file_edit_times
+
+    def watch_and_execute(self, file_list: List[str], function: Callable[[any], any], *args: any) -> None:
+        """Execute a function whenever a file from file_list changes"""
+        print("Listening for changes on {}...".format(file_list))
+        file_edit_times = self._get_file_edit_times(file_list)
```
kaustubh-karkare: You are making an assumption here that the file list is fixed, which IMO we can avoid.
Ie - if you re-explore the build-rule-graph each time, you can detect and appropriately handle changes to the build.json files too. Yes, it is costlier (so I guess you can make it configurable) but might be more useful too, right?

kaustubh-karkare (changes requested):
Excellent progress!

Now for the final requirement: Use multithreading to execute dependency build rules in parallel when possible. And demonstrate the performance benefit via a test, possibly using sleep commands in the command string.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/algorithms.py
```diff
@@ -0,0 +1,59 @@
...
+        toposort = []
+        visited = {}
+        queue = Queue()
+        in_degree = {}
+
+        # Calculate  In-Degree of Each Vertex
+        for source_node in graph_adj_list:
+            visited[source_node] = False
+            for dest_node in graph_adj_list[source_node]:
+                visited[dest_node] = False
```
kaustubh-karkare, line 30: So a node might appear on the right side (of the dict, ie - inside the list), but not the left? Hmm.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/algorithms.py
```diff
@@ -0,0 +1,59 @@
...
+                for dest_node in graph_adj_list[front]:
+                    if not visited[dest_node]:
+                        in_degree[dest_node] = in_degree[dest_node] - 1
+                        if in_degree[dest_node] == 0:
+                            queue.put(dest_node)
+
+        # If any vertex exists whose In-Degree has not been exhausted, then graph must contain a cycle
+        for node in in_degree:
+            if in_degree[node] != 0:
+                raise TopologicalSort.GraphContainsCycleException()
```
kaustubh-karkare, line 56: It would be helpful for this exception to also contain the names of the specific build rules that form the cycle.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,84 @@
...
+import json
+from typing import List
+from Builder.global_constants import GlobalConstants
+
+
+class BuildRule:
+    """Stores Command Information"""
+    def __init__(self, *, name: str, command_string: str, deps: List[str] = [], files: List[str] = []) -> None:
+        self._name = name
+        self._command_string = command_string
```
kaustubh-karkare: Now you can just call this `self._command` :)

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,84 @@
...
+        pass
+
+
+class BuildConfig:
+    """Parses and Stores build.config in the form of BuildRule objects"""
+    def __init__(self, json_containing_folder: str) -> None:
+        # Parse JSON
+        json_path = json_containing_folder + "/" + GlobalConstants.CONFIG_FILE_NAME
+        with open(json_path) as file_handle:
+            self._raw_json = json.load(file_handle)
```
kaustubh-karkare: I don't think you need to actually save this on the object.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,84 @@
...
+
+        self._name_to_command_object = {}
```
kaustubh-karkare: self._name_to_build_rule

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,84 @@
...
+        for command_entry in self._raw_json:
+            name = command_entry['name']
+            command_string = command_entry['command']
+            if 'deps' in command_entry:
+                deps = command_entry['deps']
```
kaustubh-karkare: raw_build_rule.get('deps', [])

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,84 @@
...
+            else:
+                deps = []
+            if 'files' in command_entry:
+                files = command_entry['files']
+            else:
+                files = []
+            self._name_to_command_object[name] =\
+                BuildRule(name=name, command_string=command_string, deps=deps, files=files)
+
+    def get_command(self, command_name: str) -> BuildRule:
```
kaustubh-karkare: get_build_rule

solutions/ankurdubey521/Build Automation Tool/Builder/lib/file_watcher.py
```diff
@@ -0,0 +1,33 @@
...
+    @staticmethod
+    def _get_file_edit_times(file_list: List[str]) -> Dict[str, int]:
+        """Populate a Dict of last edit times of files and return it"""
+        file_edit_times = {}
+        for file_path in file_list:
+            file_edit_times[file_path] = stat(file_path).st_mtime
+        return file_edit_times
+
+    @staticmethod
+    def watch_and_execute(file_list: List[str], function: Callable[[any], any], *args: any) -> None:
```
kaustubh-karkare: Btw, `Any` typehints are generally not considered useful, and can cause you to miss issues.
In this case, you could declare that this callable takes no arguments (because this method does not provide any) and returns nothing (because this method does not look at the return value) and force the caller to wrap the method in such a function, so that it is clear on the other side too, and the arguments/return-value have no significance here.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/file_watcher.py
```diff
@@ -0,0 +1,33 @@
...
+        """Execute a function whenever a file from file_list changes"""
+        print("Listening for changes on {}...".format(file_list))
```
kaustubh-karkare: Instead of printing to stdout directly, you can send these messages to a [logger](https://docs.python.org/3.6/library/logging.html) object, provided in the constructor, and shared across your codebase. That way, if I import your code as a library, I can just point the logger object to write to a file or something.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,220 @@
...
+        """ Constructor
+        :param root_dir_abs: Absolute path of directory which is to b considered as root (or '//')
+        :param max_threads: Max no. of threads to spawn while processing dependencies
+        """
+        # Directed graph, (u, v) => v depends on u. u, v are pairs of (rule_name, rule_dir_abs)
+        # Used for generating Topological Sort
+        self._dependency_graph = {}
+        self._dependency_topological_sort = []
+
+        # List of (dependency_name, dependency_dir_abs) for each build rule
```
kaustubh-karkare, line 44: A bit nitpicky (so feel free to ignore), but I like to put the more general item first, and then increase specificity.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,220 @@
...
+                                        by the command or it's dependencies. File changes result in the build rule
+                                        being re-run
+        """
+        # Create Dependency Graph
+        print("\nExploring Dependencies...")
+        self._explore_and_build_dependency_graph(command_name, command_dir_abs)
+        print("\nDone exploring dependencies")
+
+        # Generate Topological Sort
+        self._dependency_topological_sort = TopologicalSort.sort(self._dependency_graph)
```
kaustubh-karkare: self._topologically_sorted_build_rule_names

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,220 @@
...
+        # Handle case when no dependency exist (empty dependency graph => empty toposort)
+        if len(self._dependency_topological_sort) == 0:
+            self._dependency_topological_sort.append((command_name, command_dir_abs))
+
+        if not watch_for_file_changes:
```
kaustubh-karkare: Negative conditions are a bit harder to read / reason about, especially if you're trying to debug something in a stressful situation, and you end up encountering double/triple negatives. And so `if watch_for_file_changes: ...`

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,220 @@
...
+        print("\nExploring Dependencies...")
+        self._unresolved_commands = set()
+        self._explore_and_build_dependency_graph(command_name, command_dir_abs)
+        print("\nDone exploring dependencies")
+
+        # Generate Topological Sort
+        self._dependency_topological_sort = TopologicalSort.sort(self._dependency_graph)
+        # Handle case when no dependency exist (empty dependency graph => empty toposort)
+        if len(self._dependency_topological_sort) == 0:
+            self._dependency_topological_sort.append((command_name, command_dir_abs))
```
kaustubh-karkare: I'm generally uncomfortable with special casing, which is what this is.
I would modify the code appropriately so that this situation never happens.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,220 @@
...
+        # Mark the command's dependencies as unresolved
+        self._unresolved_commands.add((command_name, command_dir_abs))
+
+        # Process Dependencies
+        deps = command.get_dependencies()
+        for dep in deps:
+            if dep.startswith('//'):
+                # Command path referenced from root directory
+                dep_dir_abs = self._root_dir_abs + '/' + dep[2:]
+                dep_dir_abs, dep_name = dep_dir_abs.rsplit('/', 1)
```
kaustubh-karkare, line 92: - I guess this is why buck's build rules are like "//path/to/some:rule_name", so that you can do full_rule_name.partition(":"), and there's no ambiguity between the directory path and rule name.
- While I have no strong objections to using tuples like you are, I probably would have gone with strings since they (subjective opinion warning) "feel more elegant".

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,220 @@
...
+        # Handle case when no dependency exist (empty dependency graph => empty toposort)
+        if len(self._dependency_topological_sort) == 0:
+            self._dependency_topological_sort.append((command_name, command_dir_abs))
+
+        # Dict[Tuple[name, abs_dir]: Futures]
+        rule_to_futures = {}
+
+        # Execute the Build Rules. starting from the deepest dependency
+        print("\nExecuting Build Rules...")
+        with ThreadPoolExecutor(max_workers=self._max_threads) as executor:
```
kaustubh-karkare: I see you ended up using multithreading, because subprocess.Popen.wait is blocking. As a learning exercise, would you want to use  to see if you can avoid using multithreading entirely? I imagine you have a list of Popen objects, and you just loop over all of them, using the `poll` method to see if a given one is completed.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,220 @@
...
+        :param command_dir_abs:  Directory which contains build.config
+        :param dependency_futures: List of Future objects of spawned threads for running dependencies. Used to wait
+                                    until deps have finished executing
+        :return: return value of process
+        """
+        for dependency_future in dependency_futures:
+            return_value = dependency_future.result()
+            # Stop Execution if Command Fails
+            if return_value != 0:
+                exit(-1)
```
kaustubh-karkare: Does the `exit` here terminate the thread or the process?
Is it better to just return from here, and allow existing parallel commands to complete properly (but not start any new ones), instead of aborting them?

solutions/ankurdubey521/Build Automation Tool/Builder/main.py
```diff
@@ -0,0 +1,35 @@
...
+    parser.add_argument('--max-threads', type=int)
+    args = parser.parse_args(sys.argv[1:])
+
+    # Parse Command
+    command = args.build_rule
+    root_dir = ""
+    if args.root_dir is not None:
+        root_dir = args.root_dir
+    else:
+        root_dir = os.getcwd()
```
kaustubh-karkare, line 25: Alternatively,
```
current_path = os.getcwd()
root_dir = None
while current_path != root:
  if os.path.exists(current_path + os.path.sep + ".buildconfig"):
    root_dir = current_path
  current_path = parent_of(current_path)
if root_dir is None:
  raise Exception("could not find project root")
```

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_buildconfig.py
```diff
@@ -0,0 +1,34 @@
+import unittest
+import tempfile
+from Builder.lib.buildconfig import BuildConfig
+from Builder.global_constants import GlobalConstants
+
+
+class TestJsonParser(unittest.TestCase):
+    def setUp(self):
+        json_string = '[{"name":"clean","deps":["algorithms/clean"],"command":"rm -f test.o && rm -f test.exe"}' + \
+                     ',{"name":"test","files":["test.cpp"],"command":"g++ -std=c++11 -c test.cpp"}]'
```
kaustubh-karkare: You want your tests to be easily readable too. These long strings are not.
I would do something like:
```
json_string = json.dumps({
  # well formatted data
})

File: solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_buildconfig.py [hunk 56]
```diff
@@ -0,0 +1,34 @@
...
+        with tempfile.TemporaryDirectory() as tmpdir:
+            json_file_path = tmpdir + "/" + GlobalConstants.CONFIG_FILE_NAME
+            with open(json_file_path, 'w') as json_handle:
+                json_handle.write(json_string)
+            self.config = BuildConfig(tmpdir)
```
kaustubh-karkare: I guess this reveals the weakness of this API: you're forced to create a temporary file to create these objects.
Could be avoided if your constructor takes in the raw json string, and then you have a @classmethod that loads it from the file and returns an instance, if that is how you want to create it.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_file_watcher.py
```diff
@@ -0,0 +1,46 @@
...
+            input_file = file_path + "/input.txt"
+            output_file = file_path + "/output.txt"
+            with open(input_file, 'w') as file_handle:
+                file_handle.write("Hello World 1.0")
+            with open(output_file, 'w') as file_handle:
+                file_handle.write("Hello World 1.0")
+
+            # Activate Watcher and change tracked file
+            file_watcher = FileWatcher()
+            process = Process(target=file_watcher.watch_and_execute,
```
kaustubh-karkare: Any reason to prefer multiprocessing over multithreading?

My theory about the current implementation is: your watcher class doesn't have a stop_watching() method, and so `process.terminate()` is convenient.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_file_watcher.py
```diff
@@ -0,0 +1,46 @@
...
+            with open(input_file, 'w') as file_handle:
+                file_handle.write("Hello World 2.0")
+            sleep(2)
+            with open(output_file, 'r') as file_handle:
+                output_file_content = file_handle.read()
+            self.assertEqual("Hello World 2.0", output_file_content)
+
+            # Cleanup
+            with open(input_file, 'w') as file_handle:
+                file_handle.write("Hello World 1.0")
```
kaustubh-karkare: Why do you have to reset this?

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_parallel_builder.py
```diff
@@ -0,0 +1,135 @@
...
+            parallel_builder = ParallelBuilder(path, MAX_THREAD_COUNT)
+            parallel_builder._explore_and_build_dependency_graph('Z', path)
+            dependency_graph = {}
+            for node_tuple in parallel_builder._dependency_graph:
+                dependency_graph[node_tuple[0]] = []
+                for item in parallel_builder._dependency_graph[node_tuple]:
+                    dependency_graph[node_tuple[0]].append(item[0])
+            pprint.pprint(dependency_graph, indent=4)
+            correct_dependency_graph = {'X': ['Z'], 'XX': ['X'], 'XY': ['X'], 'Y': ['Z'], 'YX': ['Y'], 'YY': ['Y']}
+            self.assertEqual(correct_dependency_graph, dependency_graph)
```
kaustubh-karkare, line 133: Yeah, there's no way for me to understand this test without looking at multiple files, which is inconvenient. And because I'm lazy, I'm not motivated to go look at all the file contents, and validate them against the contents of this test here, which means that only 1 human has verified this stuff.

If I have access to an explorer-style view of your codebase, I guess that is easy, so I assume that is why you prefer it. But the PR page that I'm reviewing it in, it isn't easy to figure out. As a result, I have much stronger preference for having all the information in one place.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_parallel_builder.py
```diff
@@ -0,0 +1,135 @@
...
+            path = tmpdir + "/test"
+            shutil.copytree(local_path, path)
+            parallel_builder = ParallelBuilder(path, MAX_THREAD_COUNT)
+            parallel_builder._explore_and_build_dependency_graph('Z', path)
+            dependency_graph = {}
+            for node_tuple in parallel_builder._dependency_graph:
+                dependency_graph[node_tuple[0]] = []
+                for item in parallel_builder._dependency_graph[node_tuple]:
+                    dependency_graph[node_tuple[0]].append(item[0])
+            pprint.pprint(dependency_graph, indent=4)
```
kaustubh-karkare: Fine for debugging, but you wouldn't want this if you're running thousands of tests in parallel, which is what happens in large codebases. All the outputs will interfere with each other, and most people running the tests wont even understand the output.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_parallel_builder.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def test_parallel_sleep_commands(self):
+        local_path = os.getcwd() + '/test_builder_files/test_parallel_sleep_commands'
+        with tempfile.TemporaryDirectory() as tmpdir:
+            path = tmpdir + "/test"
+            shutil.copytree(local_path, path)
+            parallel_builder = ParallelBuilder(path, MAX_THREAD_COUNT)
+            process = Process(target=parallel_builder.execute,
+                              args=('Z', path))
+            sleep(16)
```
kaustubh-karkare: This is the only example that I checked the file contents for, because without that, I couldn't see where the number 16 came from.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/buildconfig.py
```diff
@@ -0,0 +1,84 @@
+"""" Sample build.json
```
kaustubh-karkare, line 1: This is an fairly detailed and specific example. Better suited for a readme file?

kaustubh-karkare (changes requested):
Cool, I'm pretty happy with the overall state of things, we're almost done.

My comments are now primarily to clean up inconsistencies, improve readability (stuff that was affecting the overall score) and a couple of minor extensions for learning purposes.

Btw, I tend to unashamedly ask for naming changes that span multiple files. Because if you have good tests, even large refactors should be cheap and safe.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/algorithms.py
ankurdubey521: I intended this to be a generic method.
I have implemented the requested functionality in parallel_builder.py, line 88
` # Check for circular dependencies
            if dependency_tuple in self._unresolved_commands:
                raise self.CircularDependencyException(
                    "Detected a Circular Dependency between {}:{} and {}:{}"
                    .format(dep_dir_abs, dep_name, command_dir_abs, command_name))`

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_file_watcher.py
ankurdubey521: That's true

solutions/ankurdubey521/Build Automation Tool/Builder/lib/algorithms.py
kaustubh-karkare: Yeah, that's reasonable enough.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/algorithms.py
```diff
@@ -0,0 +1,59 @@
...
+                queue.put(source_node)
+
+        # BFS
+        while not queue.empty():
+            front = queue.get()
+            visited[front] = True
+            toposort.append(front)
+            if front in graph_adj_list:
+                for dest_node in graph_adj_list[front]:
+                    if not visited[dest_node]:
```
kaustubh-karkare, line 48: I was thinking about this line for a while ... trying to figure out when it will be false. Can you help me out?

solutions/ankurdubey521/Build Automation Tool/Builder/lib/file_watcher.py
```diff
@@ -0,0 +1,39 @@
+import logging
+from os import stat
+from time import sleep
+from typing import Callable, Dict, List
+
+from Builder.global_constants import GlobalConstants
+
+# Logging Configuration
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
```
kaustubh-karkare: Instead of creating separate logger objects for each file, I think you should be passing around a single one (provided to each class as part of its constructor). That way, you only need to configure it once from your "application section" of the code.

The current setup is not thread-safe: ie - I'm trying to run multiple instances of your ParallelBuilder class for whatever reason in parallel, and I want their output to go to different destinations, that isn't possible.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,272 @@
...
+    def _explore_and_build_dependency_graph(self, command_name: str, command_dir_abs: str) -> None:
+        """ Populates self._dependency_graph and self._dependency_list and checks for CircularDependencies
+        :param command_name: .
+        :param command_dir_abs: Directory which contains build.config
+        """
+        # Remove Trailing '/' from paths if present
+        if command_dir_abs.endswith('/'):
+            command_dir_abs = command_dir_abs[:-1]
+        if self._root_dir_abs.endswith('/'):
+            self._root_dir_abs = self._root_dir_abs[:-1]
```
kaustubh-karkare: Can we move this to the constructor now?

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # Parse build.json in current directory
+        config = BuildConfig.load_from_build_directory(command_dir_abs)
+        command = config.get_build_rule(command_name)
```
kaustubh-karkare: Based on the usage, s/command_name/build_rule_name/

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,272 @@
...
+
+            # Update _dependency_graph
+            if dependency_tuple not in self._dependency_graph:
+                self._dependency_graph[dependency_tuple] = []
+            self._dependency_graph[dependency_tuple].append(dependent_tuple)
+
+            # Update _dependency_list
+            if dependent_tuple not in self._dependency_list:
+                self._dependency_list[dependent_tuple] = []
+            self._dependency_list[dependent_tuple].append(dependency_tuple)
```
kaustubh-karkare: These names "_dependency_graph" and "_dependency_list" are not intuitive: there's no way I'd figure out what is contained in them based on purely the name and, it is very easy to confuse between them.

Something like "_rule_name_to_dependency_rule_names" and "_rule_name_to_dependent_rule_names" makes it easier to figure out, although "rule_name" isn't right since you're using a tuple. I couldn't think of a better name while reviewing, but you get the point.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,272 @@
...
+    def _execute_build_rule_and_dependencies(self, command_name: str, command_dir_abs: str) -> bool:
+        """ Main execution logic
+        :param command_name: .
+        :param command_dir_abs: Directory which contains build.config
+        :return: boolean indicating build success of build rule
+        """
+        # Create Dependency Graph
+        logger.info("Exploring Dependencies...")
+        self._unresolved_commands = set()
+        self._explore_and_build_dependency_graph(command_name, command_dir_abs)
```
kaustubh-karkare: Wait, so in your "execute" method, you called "_explore_and_build_dependency_graph" and then "_execute_build_rule_and_dependencies". So why do you need to call this again?

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,272 @@
...
+        logger.info("Done exploring dependencies")
+
+        # Generate Topological Sort
+        self._topologically_sorted_build_rule_names = TopologicalSort.sort(self._dependency_graph)
+        # Handle case when no dependency exist (empty dependency graph => empty toposort)
+        if len(self._topologically_sorted_build_rule_names) == 0:
+            self._topologically_sorted_build_rule_names.append((command_name, command_dir_abs))
```
kaustubh-karkare: When does this happen?

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
kaustubh-karkare: ^ Afaict, this is still pending.
If you write your own polling mechanism, I expect you shouldn't need to use the multithreading module or this ThreadPoolExecutor at all.

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,272 @@
...
+        build_failed_for_dependency = False
+
+        # Execute the Build Rules. starting from the deepest dependency
+        logger.info("Executing Build Rules...")
+        with ThreadPoolExecutor(max_workers=self._max_threads) as executor:
+            for build_rule_tuple in self._topologically_sorted_build_rule_names:
+                if not build_failed_for_dependency:
+                    (rule_name, rule_dir_abs) = build_rule_tuple
+                    rule_command_string = \
+                        BuildConfig.load_from_build_directory(rule_dir_abs).get_build_rule(rule_name).get_command()
```
kaustubh-karkare: Hmm, as part of the exploration phase, you've already loaded data from files (which is a relatively slow operation I would think), so you should avoid loading the same data again here. Maybe maintain a "self._name_to_build_rule" mapping, that is populated once, and then used here.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_buildconfig.py
```diff
@@ -0,0 +1,46 @@
...
+                {
+                    "name": "test",
+                    "files": [
+                        "test.cpp"
+                    ],
+                    "command": "g++ -std=c++11 -c test.cpp"
+                }
+            ]
+        )
+        self.config = BuildConfig(json_string)
```
kaustubh-karkare: If you make BuildConfig take in a JSON object, you don't need to serialize it here :)

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_file_watcher.py
```diff
@@ -0,0 +1,54 @@
...
+logger = logging.getLogger()
+logger.setLevel(logging.DEBUG)
+debug_handler = logging.StreamHandler()
+debug_handler.setLevel(logging.DEBUG)
+debug_handler.setFormatter(logging.Formatter(fmt='%(asctime)s - %(message)s', datefmt="%H:%M:%S"))
+logger.addHandler(debug_handler)
+error_handler = logging.StreamHandler()
+error_handler.setLevel(logging.WARNING)
+error_handler.setFormatter(logging.Formatter(fmt='%(asctime)s - %(levelname)s - %(message)s', datefmt="%H:%M:%S"))
+logger.addHandler(error_handler)
```
kaustubh-karkare: Yeah, this is ... non-ideal.

If this is reasonable default behavior, how about moving this to your main code?
```
parallel_builder = ParallelBuilder(path, MAX_THREAD_COUNT)
# either this ...
parallel_builder.setLogger(logger)
# or this ...
parallel_builder.setDefaultLogger()
```
That way, you avoid this duplication in your tests.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_parallel_builder.py
```diff
@@ -0,0 +1,678 @@
...
+        dir_path = os.path.dirname(file_path)
+        Path(dir_path).mkdir(exist_ok=True, parents=True)
+        with open(file_path, "w") as file_handle:
+            file_handle.write(file_dict["content"])
+
+
+class TestParallelBuilder(unittest.TestCase):
+    def test_basic_shell_command(self):
+        command = "echo 'Hello World!'"
+        exit_code = ParallelBuilder._run_shell(command, cwd=os.getcwd()).wait()
```
kaustubh-karkare: Calling out how the method name "_run_shell" does not indicate that you're returning a Popen object, so without reading the implementation, I have no idea what you're calling `wait` on.

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_parallel_builder.py
```diff
@@ -0,0 +1,678 @@
...
+
+    def test_nonzero_exit_code_for_shell_command(self):
+        command = "exit 1"
+        exit_code = ParallelBuilder._run_shell(command, cwd=os.getcwd()).wait()
+        self.assertEqual(1, exit_code)
+
+    def test_dependency_graph_creation(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            path = os.path.join(tmpdir, "test")
+            write_test_files(self._testMethodName, path)
```
kaustubh-karkare: Okay, so you've moved the file structure into the test file, which is progress.
But in the interest of readability, why not keep that test structure right here too?
That way, I don't have to scroll up to find out what this file structure is.

I know this kind of thing is annoying, but the point I'm trying to make is that you always want to optimize for readability, not easy of writing (since people are going to read your code many many more times than you write it).

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_parallel_builder.py
```diff
@@ -0,0 +1,678 @@
...
+            parallel_builder = ParallelBuilder(path, MAX_THREAD_COUNT)
+            parallel_builder._explore_and_build_dependency_graph('Z', path)
+            dependency_graph = {}
+            for node_tuple in parallel_builder._dependency_graph:
+                dependency_graph[node_tuple[0]] = []
+                for item in parallel_builder._dependency_graph[node_tuple]:
+                    dependency_graph[node_tuple[0]].append(item[0])
```
kaustubh-karkare: You cant just use `parallel_builder._dependency_graph` ?

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_parallel_builder.py
```diff
@@ -0,0 +1,678 @@
...
+            self.assertEqual(correct_dependency_graph, dependency_graph)
+
+    def test_topological_sort_creation(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            path = os.path.join(tmpdir, "test")
+            write_test_files(self._testMethodName, path)
+            parallel_builder = ParallelBuilder(path, MAX_THREAD_COUNT)
+            parallel_builder.execute('Z', path)
+            toposort = [item[0] for item in parallel_builder._topologically_sorted_build_rule_names]
+            print(toposort)
```
kaustubh-karkare: Once you're done testing, remove print statements :)

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
```diff
@@ -0,0 +1,272 @@
...
+        self._dependency_list = {}
+
+        # Space for rough work :P
+        self._unresolved_commands = set()
+
+    class CircularDependencyException(Exception):
+        pass
+
+    def get_last_build_pass_status(self) -> bool:
+        return self._last_build_passed
```
kaustubh-karkare, line 70: Is this primarily for testing?

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_parallel_builder.py
```diff
@@ -0,0 +1,678 @@
...
+                for item in parallel_builder._dependency_graph[node_tuple]:
+                    dependency_graph[node_tuple[0]].append(item[0])
+            correct_dependency_graph = {'X': ['Z'], 'XX': ['X'], 'XY': ['X'], 'Y': ['Z'], 'YX': ['Y'], 'YY': ['Y']}
+            self.assertEqual(correct_dependency_graph, dependency_graph)
+
+    def test_topological_sort_creation(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            path = os.path.join(tmpdir, "test")
+            write_test_files(self._testMethodName, path)
+            parallel_builder = ParallelBuilder(path, MAX_THREAD_COUNT)
```
kaustubh-karkare: Looks like these 4 lines are common across all your tests. How about using the `setUp` and `tearDown` methods to avoid the duplication?

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_parallel_builder.py
```diff
@@ -0,0 +1,678 @@
...
+            parallel_builder.execute('run', path)
+            output_file_path = os.path.join(path, 'output')
+            with open(output_file_path) as file_handle:
+                result = file_handle.readable()
+            self.assertEqual(True, result)
+            # CLEANUP
+            parallel_builder = ParallelBuilder(path, MAX_THREAD_COUNT)
```
kaustubh-karkare: You cant reuse the existing ParallelBuilder object?

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_parallel_builder.py
```diff
@@ -0,0 +1,678 @@
...
+            end_time = time.time()
+            self.assertTrue(end_time - start_time < 16)
+            self.assertTrue(parallel_builder.get_last_build_pass_status())
+
+    def test_files_list_generation_adds_files_of_dependencies(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            path = os.path.join(tmpdir, "test")
+            write_test_files(self._testMethodName, path)
+            parallel_builder = ParallelBuilder(path, MAX_THREAD_COUNT)
+            parallel_builder._explore_and_build_dependency_graph('Z', path)
```
kaustubh-karkare, line 625: Alternative **optional** design suggestion: instead of having to call this manually, how about doing it from the ParallelBuilder constructor? Instead of exploring dependencies for a single rule, it creates the graph for everything under that directory (not specific to any rule, because that isn't actually needed), and then you can choose to execute a single rule (for which the information is now already cached).

kaustubh-karkare (changes requested):

solutions/ankurdubey521/Build Automation Tool/Builder/tests/test_parallel_builder.py
ankurdubey521: fixed

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
ankurdubey521: yes

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
ankurdubey521: fixed

solutions/ankurdubey521/Build Automation Tool/Builder/lib/parallel_builder.py
ankurdubey521: when a rule has no dependency, the dependency graph will contain no edge, thus it's adjacency list would be empty

kaustubh-karkare (approved):
Cool, this is sufficiently close to the end goal I had in mind. As always, there is more that can be done, but I'm going to approve, since I understand that it can be boring to keep working on the same problem for so long.

PR #61, opened 2019-09-15
assignment-1 solution

drs-11:


solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,103 @@
...
+
+class Parser(object):
+    """
+    Class to handle parsing
+    """
+
+    def __init__(self):
+        self.args = {}          #storing arguments as objects as values and their names as keys
+
+    def add_argument(self, arg_name, arg_type, cant_be_used_with=[], required=False):
```
kaustubh-karkare: Instead of duplicating all the argument names, you could just use `def add_argument(self, *args, **kwargs):`

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,103 @@
...
+        Initializes Argument class
+        """
+        self.args[arg_name] = Argument(arg_name, arg_type, cant_be_used_with, required)
+
+
+    def parse_arguments(self):
+        """
+        Parses argument and assignes value to argument dictionary
+        """
+        arguments = sys.argv[1:]
```
kaustubh-karkare: Take this as an argument, so that you can provide whatever value you want as part of tests.
And if not provided, then you can default to this.

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,103 @@
...
+
+        for argument in arguments:
+            try:
+                arg, value = argument.split('=')
```
kaustubh-karkare: * What if there are more/less parts?
* Check out the partition method on strings too.
* A slightly better set of variable names is "name, value", because args refers to the objects and not strings, right?

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,103 @@
...
+                value = int(value) if value.isdigit() else value        #typecasting integer value if string
```
kaustubh-karkare: Only cast to int if the type is int. Because sometimes people might want to put only numbers in a string field too.

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,103 @@
...
+            arg = arg[2:]           #removing the dashes
+
+            if value:
+                if not arg in self.args:
+                    raise NoSuchArgError(arg)
+                if not isinstance(value, self.args[arg].arg_type):
+                    raise WrongTypeError(type(value), self.args[arg].arg_type, arg)
+            else:
+                if not arg in self.args:
+                    raise NoSuchArgError(arg)
```
kaustubh-karkare: This part looks duplicated.

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,103 @@
...
+    Class to define argument definitions and characteristics
+    """
+
+    def __init__(self, arg_name, arg_type, cant_be_used_with=[], required=False):
+        self.arg_name = arg_name
+        self.arg_type = arg_type
+        self.value = None
+        self.required = required
+        self.used = False           #flag to identify if the argument is used or not
+        self.cant_be_used_with = cant_be_used_with
```
kaustubh-karkare: What if I wanted to enforce that exactly 1 of 5 different arguments is used?
Ie - in this model, I would have to mention the other 4 arguments in each case, which gets tedious.
Also, how do I enforce that exactly one is present?

solutions/assignment-1/test_runs.md
```diff
@@ -0,0 +1,29 @@
+# Test runs:
```
kaustubh-karkare: Cool, this is a decent start, but yeah, we need to use a proper framework for testing.
https://docs.python.org/3/library/unittest.html

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,103 @@
...
+
+class Argument(object):
+    """
+    Class to define argument definitions and characteristics
+    """
+
+    def __init__(self, arg_name, arg_type, cant_be_used_with=[], required=False):
+        self.arg_name = arg_name
+        self.arg_type = arg_type
+        self.value = None
```
kaustubh-karkare: You shouldn't store the values in here.
Create a dict in your `parse_arguments` method and return that.

kaustubh-karkare (changes requested):

drs-11:
added solution for assignment-2

kaustubh-karkare (changes requested):
The solution for assignment 2 should be a separate branch/pull-request. I cant review it here, since I don't want to mix the history of comments for 2 problems. I will take a proper look once you separate them :)

kaustubh-karkare:
This is still pending.

drs-11:
will solve them ASAP

solutions/assignment-1/CLIparser.py
drs-11: If I use a dictionary for storing values in that method then the class Argument will become useless. So should I remove the class as well then?

drs-11:
changes done

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,97 @@
...
+        self.test = test
+
+    def add_argument(self, arg_name, **kwargs):
+        """
+        Initializes Argument class
+        """
+        self.args[arg_name] = dict()
+        self.args[arg_name]["used"] = False
+        self.args[arg_name]["cant_be_used_with"] = None
+        self.args[arg_name]["required"] = False
```
kaustubh-karkare: Why should these 3 things be special-cased and then everything is also valid?
Instead, you should explicitly define which properties are valid:

```
def add_argument(self, arg_name, *, required=True, <name>=<default>):
```

That way, you only allow users to specify valid things.

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,97 @@
...
+		
+        for name, value in kwargs.items():
+            self.args[arg_name][name] = value
+            
+
+    def parse_arguments(self, *args):
+        """
+        Parses argument and assignes value to argument dictionary
+        """
+        if self.test:
```
kaustubh-karkare: This is special casing: the Parser class should not be aware of whether it is used in a test or in real.
Instead, you can change the signature to `def parse_arguments(self, argv=None)`, and if argv = None, fallback to sys.argv

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,97 @@
...
+            arguments = sys.argv[1:]
+
+        for argument in arguments:
+            name, sign, value = argument.partition('=')
+            name = name.split("-")[-1]
+            try:
+                value = int(value) if (self.args[name]["type"] == int) else value        #typecasting
+                value = float(value) if (self.args[name]["type"] == float) else value	   
+            except ValueError:
+                pass
```
kaustubh-karkare: This is more special casing: we want to be able to support arbitrary types, but here, you only support these 2.
Instead, you can do: `value = self.args[name]["type"](value)`
where the user provides a `type` method that validates whatever the type is, and the value does not comply, raises a ValueError.

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,97 @@
...
+            if value:                
+                if not isinstance(value, self.args[name]["type"]):
+                    raise WrongTypeError(type(value), self.args[name]["type"], name)
+            if not name in self.args:
+                raise NoSuchArgError(name)
+
+            self.args[name]["value"] = value
+            self.args[name]["used"] = True
```
kaustubh-karkare: Don't save this stuff on the class. It is only relevant in this method, and so save it as a local variable, and then return it.

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,97 @@
...
+
+        for arg in self.args:       #checking if an argument can be used with the others
+            if self.args[arg]["used"]:
+                if self.args[arg]["cant_be_used_with"]:
+                    for other_arg in self.args[arg]["cant_be_used_with"]:
```
kaustubh-karkare: This model has a problem: fields that cannot be used together are called "mutually exclusive". If you have 2 such fields, each field name is mentioned twice (so 4 times in total). If you have 5 such field names (ie - cant_be_used_with contains 4 items), each field name is mentioned 5 times (and so 25 times in total).
This is not good, it makes it hard to manage large groups of mutually exclusive fields.

solutions/assignment-1/test.py
```diff
@@ -0,0 +1,17 @@
+import CLIparser
+import json
+
+parser = CLIparser.Parser()
+parser.add_argument('key', int, required=True)
+parser.add_argument('name', str)
```
kaustubh-karkare: Use keyword arguments when possible:
```
parser.add_argument('name', type=str)
```

solutions/assignment-1/test.py
```diff
@@ -0,0 +1,17 @@
...
+parser.add_argument('local', None, cant_be_used_with=['remote'])
+parser.add_argument('remote', None, cant_be_used_with=['local'])
+parser.parse_arguments()
+args = parser.args
```
kaustubh-karkare: The user of the library should deal with a very simple API, doing as little work as possible.
```
args = parser.parse_arguments()
print(json.dumps(args))
```

solutions/assignment-1/test_runs.md
kaustubh-karkare: ^ This comment still applies.

kaustubh-karkare (changes requested):
We need to fix a bunch of special casing in this code.

solutions/assignment-1/unittest_CLIparser.py
```diff
@@ -0,0 +1,35 @@
+import unittest
+import v3
+
+class TestParser(unittest.TestCase):
+
+    def test_basic_usage(self):
+        parser = v3.Parser()
+        parser.add_argument('key', type=int)
+        parser.add_argument('name', type=str)
+        self.assertEqual(parser.parse_arguments(["--key=12345", "--name=drs"]), (eval('{"key": 12345, "name":"drs"}')))
```
kaustubh-karkare: You don't need to eval the string.
```
eval('{"key": 12345, "name":"drs"}') == {"key": 12345, "name":"drs"}  # true
```

solutions/assignment-1/unittest_CLIparser.py
```diff
@@ -0,0 +1,35 @@
+import unittest
+import v3
```
kaustubh-karkare: What is v3?

solutions/assignment-1/unittest_CLIparser.py
```diff
@@ -0,0 +1,35 @@
...
+        parser = v3.Parser()
+        parser.add_argument('key', type=int, required=True)
+        parser.add_argument('name', type=str)
+        with self.assertRaises(v3.ReqArgError):
+            parser.parse_arguments()
+
+    def test_simultaneous_usage(self):
+        parser = v3.Parser()
+        parser.add_argument('local', type=None, mutually_exclusive_arg=True)
+        parser.add_argument('remote', type=None, mutually_exclusive_arg=True)
```
kaustubh-karkare: You should support multiple groups of mutually exclusive arguments :)
So --local and --remote are one group, and additionally --ip-address and --domain-name could be another group.

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,99 @@
...
+        self.args = dict()
+        self.mutually_exclusive_args = list()
+
+    def add_argument(self, arg_name, *, required=False, used=False, mutually_exclusive_arg=False, **kwargs):
+
+        self.args[arg_name] = dict()
+        self.args[arg_name]["required"] = required
+        self.args[arg_name]["used"] = used
+        for name, value in kwargs.items():
+            self.args[arg_name][name] = value
```
kaustubh-karkare: What is the value of this?

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,99 @@
...
+
+        for argument in argv:
+            name, sign, value = argument.partition('=')
+            name = name.split("-")[-1]
+            try:
+                value = self.args[name]["type"](value)
+            except TypeError:
+                pass
+            except ValueError:
+                pass
```
kaustubh-karkare: What are these used for?

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,99 @@
...
+            if value:
+                if not isinstance(value, self.args[name]["type"]):
```
kaustubh-karkare, line 47: What happens if `self.args[name]["type"]` is a custom function?

solutions/assignment-1/CLIparser.py
kaustubh-karkare: ^ Pending. This is the main reason for RC :)

kaustubh-karkare (changes requested):
Good progress, almost done!

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,99 @@
...
+    def __init__(self):
+        self.args = dict()
+        self.mutually_exclusive_args = list()
+
+    def add_argument(self, arg_name, *, required=False, **kwargs):
+
+        self.args[arg_name] = dict()
+        self.args[arg_name]["required"] = required
+        for attribute_name, attribute_value in kwargs.items():
+            self.args[arg_name][attribute_name] = attribute_value
```
kaustubh-karkare, line 21: Given that you will never use these additional attributes, why even allow them?

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,99 @@
...
+
+        if not argv:
+            argv = sys.argv[1:]
+
+        for argument in argv:
+            name, sign, value = argument.partition('=')
+            name = name.split("-")[-1]
+            try:
+                value = self.args[name]["type"](value)
+            except TypeError:        #to catch None calling as a function error which should be ignored
```
kaustubh-karkare, line 41: Instead of this, you could just do:
```
if self.args[name].get("type") is None:
  pass
else:
  # call the method
```

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,99 @@
...
+                pass
+            except ValueError:          #the typecasting is being done to convert str to respective types but sometimes the entered value is a wrong type which is later raised by the isinstance function
```
kaustubh-karkare, line 43: Yeah, you have 2 different steps responsible for doing type validation, which IMO is inelegant.
Ie - you could drop the next step entirely and just rely on this one, and create the contract that if the "type" method  does not raise a ValueError, then it means that the library user is fine with the returned value.

However, we discussed this, and agreed that it is fine for now, so no-op.

solutions/assignment-1/unittest_CLIparser.py
```diff
@@ -0,0 +1,36 @@
...
+        parser.add_argument('key', type=int, required=True)
+        parser.add_argument('name', type=str)
+        with self.assertRaises(v3.ReqArgError):
+            parser.parse_arguments()
+
+    def test_simultaneous_usage(self):
+        parser = v3.Parser()
+        parser.add_argument('local', type=None)
+        parser.add_argument('remote', type=None)
+        parser.add_mutually_exclusive_args('local', 'remote')
```
kaustubh-karkare, line 31: Much better, so you specify the relationship exactly once.

solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,99 @@
...
+                raise NoSuchArgError(name)
+            
+            arg_parsed[name] = value
+
+        for arg in arg_parsed:       #checking if an argument can be used with the others
+            for group in self.mutually_exclusive_args:
+                if arg in group:
+                    for other_arg in group:
+                        if other_arg in arg_parsed:
+                            raise SimultaneousUsageError(arg, group)
```
kaustubh-karkare, line 59: Wait, this is O(n^3) complexity? I think it can be reduced to O(n) time + O(n) space.

kaustubh-karkare (approved):

PR #62, opened 2019-10-11
Added Solution for Assignment 2.1

pranjalrai:
Added solution for Build Automation Tool along with unit tests.


solutions/Pranjal Rai/src/Assignment 2/BuildAutomationTool.py
```diff
@@ -0,0 +1,45 @@
...
+import json
+
+
+class BuildAutomationTool():
+
+	def execute_command(self, current_working_directory, previous_working_directory, process_command, current_script):
+
+		if process_command != "build":
+			raise Exception(process_command + ' is not a recognized process command')
+			return
```
kaustubh-karkare, line 12: Hmm, this "process_command" thing shouldn't be necessary, if there is only 1 allowed value for this.

solutions/Pranjal Rai/src/Assignment 2/BuildAutomationTool.py
```diff
@@ -0,0 +1,45 @@
...
+
+		absolute_path = os.path.abspath(current_working_directory)
+		os.chdir(absolute_path)
+		script_found = False
+
+
+		with open(os.path.join("build.json")) as json_file:
+			scripts = json.load(json_file)
+
+			for script in scripts:
```
kaustubh-karkare, line 23: Lets call these actions, or build-rules, or something.

I recommend creating a separate class for an Action, and maintaining a proper graph, because it will be useful in part3 of the problem.

solutions/Pranjal Rai/src/Assignment 2/BuildAutomationTool.py
```diff
@@ -0,0 +1,45 @@
...
+				if script['name'] == current_script:
+					script_found = True
+					if 'deps' in script:
+						for dependecy in script['deps']:
+							path = str(dependecy).split('/')
+							new_script = path.pop()
+							new_path = ""
+							for path_segment in path:
+								new_path += path_segment
+							self.execute_command(new_path, current_working_directory, process_command, new_script)
```
kaustubh-karkare, line 33: You need to detect circular dependencies and show appropriate error message.

solutions/Pranjal Rai/src/Assignment 2/BuildAutomationTool.py
```diff
@@ -0,0 +1,45 @@
...
+					current_command = script['command']
+					os.system(current_command)
```
kaustubh-karkare, line 35: This vaguely satisfied part1 of the problem statement, lets do part2 now.

solutions/Pranjal Rai/src/Assignment 2/BuildAutomationTool.py
```diff
@@ -0,0 +1,45 @@
...
+		absolute_path = os.path.abspath(current_working_directory)
+		os.chdir(absolute_path)
+		script_found = False
+
+
+		with open(os.path.join("build.json")) as json_file:
+			scripts = json.load(json_file)
+
+			for script in scripts:
+				if script['name'] == current_script:
```
kaustubh-karkare, line 24: What if I provide current_script = "algorithms/sort_bubble" ? The name in build.json will only be "sort_bubble", and you're not looking in the "algorithms" directory.

solutions/Pranjal Rai/src/Assignment 2/BuildAutomationToolTest.py
```diff
@@ -0,0 +1,53 @@
...
+		previous_location = absolute_location
+		absolute_location += "/algorithms"
+		os.chdir(absolute_location)
+		self.assertEqual(path.exists('sort_bubble.o'), False)
+		self.assertEqual(path.exists('sort_quick.o'), False)
+		self.assertEqual(path.exists('sort_merge.o'), False)
+		os.chdir(previous_location)
+		
+
+	def test_build_run(self):
```
kaustubh-karkare, line 25: Instead of creating new files in the current directory, use tempfile.TemporaryDirectory so that you can run multiple copies of the test in parallel.

solutions/Pranjal Rai/src/Assignment 2/BuildAutomationToolTest.py
```diff
@@ -0,0 +1,53 @@
...
+		build_automation_tool_object = Build.BuildAutomationTool()
+		build_automation_tool_object.execute_command(os.getcwd(), os.getcwd(), 'build', 'run')
```
kaustubh-karkare, line 27: This API is suboptimal.
The fact that you need to maintain previous_working_directory is an implementation detail, that is an internal problem for the library, and should not affect the API.
Can you do something like:
```
BuildAutomationTool(root=".").execute("algorithms/sort_bubble")

kaustubh-karkare (changes requested):
The code for CommandLineParser should not appear here. Use a separate branch for separate problem.

PR #64, opened 2019-10-13
Added build and build_test.

humanoid2795:
A rudimentary build tool that takes in a *name* as an argument and then builds the dependencies, ensures all files are present and executes the build command.

### DIFF
--- a/solutions/build-automation-tool/build.py
+++ b/solutions/build-automation-tool/build.py
@@ -1,0 +1,120 @@
+import enum
+import json
+import os
+import subprocess
+import sys
+
+
+class Build(object):
+    """
+    The class builds the dependencies, checks for the existence of required
+    files and executes the build command for the specified name.
+    """
+
+    class BuildSpecs(enum.Enum):
+        BUILD = 'build.json'
+        NAME = 'name'
+        DEPENDENCIES = 'dependencies'
+        FILES = 'files'
+        COMMAND = 'command'
+
+    def __init__(self, build_file=os.path.join(os.getcwd(), 'build.json')):
+        """
+        :param build_file: Path of the build file.
+        :type build_file: str
+        """
+
+        self.build_instructions = None
+        with open(build_file, 'r') as build_file:
+            self.build_instructions = json.load(build_file)
+
+    def _getBuildInstruction(self, name):
+        """
+        Returns the build instruction available for name.
+
+        :param name: The name of the build instruction.
+        :type name: str
+        """
+
+        for build_instruction in self.build_instructions:
+            if build_instruction[self.BuildSpecs.NAME.value] == name:
+                return build_instruction
+        return None
+
+    def _buildDependencies(self, dependencies):
+        """
+        Builds the dependencies for the current build instruction.
+
+        :param dependencies: The dependecnies to build.
+        :type depenedecies: list
+        """
+
+        for dependency in dependencies:
+            build_directory, name = os.path.split(dependency)
+            if build_directory:
+                current_directory = os.getcwd()
+                try:
+                    os.chdir(build_directory)
+                    build_dependency = Build(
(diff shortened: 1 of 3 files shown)


solutions/build-automation-tool/build.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+
+    class BuildSpecs(enum.Enum):
+        BUILD = 'build.json'
+        NAME = 'name'
+        DEPENDENCIES = 'dependencies'
+        FILES = 'files'
+        COMMAND = 'command'
+
+    def __init__(self, build_file=os.path.join(os.getcwd(), 'build.json')):
```
kaustubh-karkare: You don't really need a default value here, you can just provide this path from the main section.
Create the entire graph here, check for cyclic dependencies and stuff.

solutions/build-automation-tool/build.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def _getBuildInstruction(self, name):
+        """
+        Returns the build instruction available for name.
+
+        :param name: The name of the build instruction.
+        :type name: str
+        """
+
+        for build_instruction in self.build_instructions:
```
kaustubh-karkare: Instead of O(n) search, you can create a Map<name, config>

solutions/build-automation-tool/build.py
```diff
@@ -0,0 +1,120 @@
+import enum
+import json
+import os
+import subprocess
+import sys
+
+
+class Build(object):
```
kaustubh-karkare: * Looks like this class represents a single build.json file, and so can be called "BuildConfig". The individual items inside that are called "BuildRule"s.
* If you create a map<full_rule_name, BuildRule>, I don't think you need a BuildConfig class.

solutions/build-automation-tool/build.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    The class builds the dependencies, checks for the existence of required
+    files and executes the build command for the specified name.
+    """
+
+    class BuildSpecs(enum.Enum):
+        BUILD = 'build.json'
```
kaustubh-karkare: This is unexpected. The rest of the items are attributes of the BuildRule structure, but this one is the file name, and so inconsistent.

solutions/build-automation-tool/build.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def _ensureFiles(self, files):
+        """
+        Asserts if all the files are exists or not
+
+        :param files: Files to check for existence
+        :type files: list
+        """
+
+        for file in files:
```
kaustubh-karkare: The name "file" is not useful, since it isn't clear whether it is a file_name or file_handle or whatever else.

solutions/build-automation-tool/build_test.py
```diff
@@ -0,0 +1,106 @@
...
+import os
+import tempfile
+import textwrap
+import unittest
+
+from build import Build
+
+
+@contextlib.contextmanager
+def create_and_change_to_tmpdir(working_directory):
```
kaustubh-karkare: Why does the caller need to provide working_directory? You can just calculate it here.

solutions/build-automation-tool/build.py
```diff
@@ -0,0 +1,120 @@
...
+        :param name: The name of the build instruction.
+        :type name: str
+        """
+
+        for build_instruction in self.build_instructions:
+            if build_instruction[self.BuildSpecs.NAME.value] == name:
+                return build_instruction
+        return None
+
+    def _buildDependencies(self, dependencies):
```
kaustubh-karkare: Assumption: A depends on B & C, and both B & C depend on D.
* Right now, you will build D twice. Can you avoid that?
* Can you build B & C in parallel? Use subprocess.Popen, but you shouldn't need to use any multithreading.

PR #65, opened 2019-10-24
solution for assignment 2

drs-11:


solutions/d-r-siddhartha/assignment-2/build.json
```diff
@@ -0,0 +1,17 @@
...
+    "command": "rm -f test.o && rm -f test.exe"
+  },
+  {
+    "name": "test",
+    "files": ["test.cpp"],
+    "command": "g++ -std=c++11 -c test.cpp"
+  },
+  {
+    "name": "run",
+    "deps": ["test", "algorithms/sort_bubble", "algorithms/sort_merge", "algorithms/sort_quick"],
```
kaustubh-karkare: See the updated problem statement.
In order to do part3 properly, we want to only link one sorting algorithm at a time with the test.cpp file.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,144 @@
...
+    class encapsulation for build tool
+    """
+
+    def __init__(self):
+        self.output = list()
+        self.addr_log = dict()
+        self.dir_struct = self.__generate_dir_struct(os.getcwd())
+        if len(sys.argv) > 1:
+            self.rule = sys.argv[1]
+            self.watch_run = bool((len(sys.argv) > 2) and (sys.argv[2] == "--watch"))
```
kaustubh-karkare: Don't do this manually. Use the argparse library.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,144 @@
...
+    def __generate_dir_struct(path):
+        """
+        generates a dictionary of file name(key) and timestamp of last modification(value)
+        """
+        base_addr = os.getcwd()
+        dir_struct = dict()
+        for root, dirs, files in os.walk(path):
+            for file_name in files:
+                if "pyc" in file_name: #weak statement, needs better method to add only those files required in the build rule file
+                    continue
```
kaustubh-karkare: Quoting problem statement:

> Feel free to add support for new attributes in the action data structure.

Because a build automation tool should not make assumptions like this.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,144 @@
...
+                os.chdir(root)
```
kaustubh-karkare: Pretty sure you don't need to change directory if you use the full path in os.stat

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,144 @@
...
+
+        self.addr_log[rule] = addr
+        os.chdir(addr)
+
+        try:
+            build_file = open('build.json', 'r')
+        except FileNotFoundError:
+            print("Error: Build json file not found!")
+            return None
+        build_data = json.load(build_file)
```
kaustubh-karkare: Use something like:
```
with open(path + 'build.json') as file_handle:
  json.load(file_handle)
```

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,144 @@
...
+                            dep_addr = addr
+                            dep_rule = dep
+
+                        self.rule_exe(dep_rule, dep_addr)
+                        os.chdir(self.addr_log[rule])
+
+                output_file = os.popen(obj["command"])
+                output_data = output_file.read()
+                if output_data:
+                    print(output_data)
```
kaustubh-karkare: Never print to stdout directly unless you're debugging, and remember to clean up afterwards.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,144 @@
...
+        os.chdir(addr)
+
+        try:
+            build_file = open('build.json', 'r')
+        except FileNotFoundError:
+            print("Error: Build json file not found!")
+            return None
+        build_data = json.load(build_file)
+
+        for obj in build_data:
```
kaustubh-karkare: Check out part2 of the problem.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        function to implement watch feature which recompiles recently modified files
+        """
+        while True:
+            os.chdir(addr)
+            temp_struct = self.__generate_dir_struct(os.getcwd())
+            time.sleep(10)
+            for file_name in temp_struct:
+                try:
+                    if temp_struct[file_name] != self.dir_struct[file_name]:
```
kaustubh-karkare: Names like `temp_struct` and `dir_struct` aren't useful to figure out what the value is.

See coding instruction 3 [here](https://github.com/kaustubh-karkare/project-omega/blob/master/assignments/README.md).

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,144 @@
...
+                self.watch_feature(os.getcwd())
+
+
+    def rule_exe(self, rule, addr):
+        """
+        function to execute rule commands and of its dependencies recursively
+        """
+
+        self.addr_log[rule] = addr
+        os.chdir(addr)
```
kaustubh-karkare: After going through your code, I don't think you need to be using os.chdir as much as you are using.
In fact, I'd challenge you to modify your code to avoid using it entirely.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,144 @@
...
+    @staticmethod
+    def check_parent_rule(file_name, base_addr):
+        """
+        static method that checks the parent of a rule if it is a dependency
+        """
+        for walk in os.walk(base_addr):
+            os.chdir(base_addr)
+            os.chdir(walk[0])
+            try:
+                build_file = open('build.json', 'r')
```
kaustubh-karkare: If you've already read this information once from the files, no need to read it again.

You should store all the information in memory after the first reading.

Create classes to represent actions, and you should also figure out how to represent the graph.

solutions/d-r-siddhartha/assignment-2/build_test.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+import os
+from build import Builder
+
+class BuildToolTest(unittest.TestCase):
+
+    tool = Builder()
+
+    def test_cpp_example(self):
+        self.assertEqual(self.tool.rule_exe("run", os.getcwd())[-1], "1 2 3 4 5 \n1 2 3 4 5 \n1 2 3 4 5 \n")
```
kaustubh-karkare: * Since this operation will create files in this location, if you're running the "clean" action in parallel, it will interfere. You should make this impossible. Run your tests in a temporary directory if they will modify file structure (see the tempfile module).

kaustubh-karkare (changes requested):
Decent start, but the code is kinda focussed on only the usecase you were trying to solve, and isn't really flexible. You can fix that as you work on implementing part2 of the problem.

drs-11:
on it

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+import os
+import sys
+import json
+import time
+
+class Graph(object):
+
+    def __init__(self):
+        self.graph_loc = {"base_addr" : os.getcwd()}
+        self.graph_dict = dict()
```
kaustubh-karkare: graph_dict? It doesn't tell me much about  the contents :(

Based on usage below, lets call this "rule_name_to_rule" ?

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def create_graph(self):
+        '''
+        links rules and their dependencies forming a graph from a rule json file
+        '''
+        for root, dirs, files in os.walk(os.getcwd()):
+            if 'build.json' in files:
```
kaustubh-karkare: Store this "build.json" string in some constant and reuse it.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+                with open(root + '/build.json') as build_file:
```
kaustubh-karkare: os.path.join(current_directory, constant)

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+                    rule_json = json.load(build_file)
+                    for obj in rule_json:
+                        rule_name = root.partition(self.graph_loc["base_addr"])[-1][1:] + "/" + obj["name"] if root != self.graph_loc["base_addr"] else obj["name"]
```
kaustubh-karkare: This line is hard to understand.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def __init__(self):
+        self.graph_loc = {"base_addr" : os.getcwd()}
+        self.graph_dict = dict()
+
+    def create_graph(self):
+        '''
+        links rules and their dependencies forming a graph from a rule json file
+        '''
+        for root, dirs, files in os.walk(os.getcwd()):
```
kaustubh-karkare: s/root/current_location/

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+                            self.graph_dict[rule_name] = Rule(obj["name"])
+                        self.graph_loc[rule_name] = root
+                        if "deps" in obj:
+                            for dep in obj["deps"]:
+                                self.graph_dict[rule_name].addDeps(dep)
+                                if dep not in self.graph_dict:
+                                    self.graph_dict[dep] = Rule(dep)
+                                self.graph_dict[dep].parents.append(rule_name)
+                        if "command" in obj:
+                            self.graph_dict[rule_name].command = obj["command"]
```
kaustubh-karkare: You can move this logic to the Rule class, which is responsible for extracting information from a single JSON object. That way, the Graph doesn't need to worry about Rule-level logic.

rule = Rule.create(obj)
self.rule_name_to_rule[rule.name] = rule

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+
```
kaustubh-karkare: In a separate method, you can detect if there are circular dependencies.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+        for dep in self.executor.graph[rule].unresolved_deps:
+            self.map_file_command(dep)
+
+
+class Rule(object):
+
+    def __init__(self, name):
+        self.name = name
+        self.resolved_deps = list()
+        self.unresolved_deps = list()
```
kaustubh-karkare: Hmm, I don't think we should store the concept of resolved/unresolved at the rule level.
It matters only for executors, so that is the class which can store this information.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+    def __init__(self, graph, location_map):
+        self.graph = graph
+        self.location_map = location_map
+
+    def exe(self, rule):
+        '''
+        executes rules and their dependencies in parallel reducing execution time
+        '''
+        process_to_execute = self.get_independent_deps(rule)
+        name_to_process = {dep: subprocess.Popen((self.graph[dep].command), cwd=self.location_map[dep], shell=True) for dep in process_to_execute if self.graph[dep].command}
```
kaustubh-karkare: - I thought the linter would complain about this line being too long to read properly.
- It looks like `process_to_execute` and `name_to_process.keys()` are basically the same, right? So why have both. Instead of maintaining both structures, just maintain one of them.
- The fact that a rule might not have a command requires special handling here. Can you modify your rule class to hide that detail, so that executors do not have to know about it?

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+        while True:
+            for dep in list(name_to_process):
+                if name_to_process[dep].poll() is None:
+                    continue
+                else:
+                    dep_obj = self.graph[dep]
+                    del name_to_process[dep]
+                    process_to_execute.remove(dep)
+                    for parent in dep_obj.parents:
+                        self.graph[parent].remDeps(dep)
```
kaustubh-karkare: Yeah, the executor shouldn't be able to modify the graph/rules objects that it depends on, since this state is relevant only to executor.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+                                self.graph_dict[dep].parents.append(rule_name)
+                        if "command" in obj:
+                            self.graph_dict[rule_name].command = obj["command"]
+
+
+class ParallelExe(object):
+
+    def __init__(self, graph, location_map):
+        self.graph = graph
+        self.location_map = location_map
```
kaustubh-karkare: IMO the location of the rule should be saved in the rule object, not in a separate map outside the graph.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def __init__(self, executor):
+        self.executor = executor
+        self.file_deps = dict()
+        self.file_watch = dict()
+
+    def watch_init(self):
+        if not self.file_deps: return
+        while True:
+            for root, dirs, files in os.walk(os.getcwd()):
```
kaustubh-karkare: Why os.getcwd? I would think that the graph.base_addr is the relevant thing here.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+                                self.executor.exe(self.file_deps[file])
+                                for parent in self.executor.graph[self.file_deps[file]].parents:
+                                    self.executor.exe(parent)
+                                self.file_watch[file] = os.stat(root+'/'+file).st_mtime
+                        else:
+                            print(root+'/'+file)
+                            self.file_watch[file] = os.stat(root+'/'+file).st_mtime
+            time.sleep(1)
+
+    def map_file_command(self, rule):
```
kaustubh-karkare, line 216: Who is calling these methods? This codebase is incomplete.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,157 @@
...
+    def __init__(self, executor):
+        self.executor = executor
+        self.file_deps = dict()
+        self.file_watch = dict()
+
+    def watch_init(self):
+        if not self.file_deps: return
+        while True:
+            for root, dirs, files in os.walk(os.getcwd()):
+                for file in files:
```
kaustubh-karkare: I thought you were going to only look at the files explicitly listed in the rule, not all of them.

kaustubh-karkare (changes requested):

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+        if rule in self.file_deps.values():
+            return
+        files = input("Enter depended files for %s(Press Enter if none): " % (rule)).split()
+        for file in files:
+            self.file_deps[file] = rule
+        for dep in self.executor.graph[rule].deps:
+            self.map_file_command(dep)
+
+
+class Rule():
```
kaustubh-karkare: Minor preference: put this class at the top, so that I can see how Rule is defined before the Graph that uses it.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+        self.deps = list()
+        self.parents = list()
+        self.command = None
+        self.address = None
+
+    def addDeps(self, dep):
+        self.deps.append(dep)
+
+    def remDeps(self, dep):
+        self.deps.remove(dep)
```
kaustubh-karkare: You should ideally avoid modifying the rule objects based on executor.
Ie - if executor needs to remove dependencies, it can keep track of that information in its own class, it shouldn't mutate dependencies.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class Rule():
+
+    def __init__(self, name):
+        self.name = name
+        self.deps = list()
+        self.parents = list()
+        self.command = None
+        self.address = None
```
kaustubh-karkare: Why not just call this `self.cwd` for consistency? address is new terminology.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+    def remDeps(self, dep):
+        self.deps.remove(dep)
+
+    def exe_rule(self):
+        if self.command:
+            return subprocess.Popen((self.command), cwd=self.address, shell=True)
+        else:
+            return None
+
+    def create(self, obj, rule_name, rule_name_to_rule, addr):
```
kaustubh-karkare:
```
@classmethod
def create(cls, ...):
    # do whatever you need to
    return cls(...)

# since rule_name_to_rule is a feature of the Graph, not an individual rule,
# the above create method should not be responsible for populating it
rule = Rule.create(json.load(...))
self.rule_name_to_rule[rule.name] = rule

# once all the rules are populated, then add dependents
```

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+        for dep in self.executor.graph[rule].deps:
+            self.map_file_command(dep)
+
+
+class Rule():
+
+    def __init__(self, name):
+        self.name = name
+        self.deps = list()
+        self.parents = list()
```
kaustubh-karkare: Lets call these `dependencies` and `dependents`?
A name like `parent` implies children, and the mapping with the above isn't clear.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+import os
+import json
+import time
+
+class Graph():
+
+    def __init__(self):
+        self.BASE_ADDR = os.getcwd()
+        self.rule_name_to_rule = dict()
+        self.RULE_FILE = 'build.json'
```
kaustubh-karkare: If this is a constant (ie - because of the all-caps name), it should be outside the constructor.
```
class Graph:
    RULE_FILE = "build.json"

File: solutions/d-r-siddhartha/assignment-2/build.py [hunk 32]
```diff
@@ -0,0 +1,186 @@
+import subprocess
+import os
+import json
+import time
+
+class Graph():
+
+    def __init__(self):
+        self.BASE_ADDR = os.getcwd()
```
kaustubh-karkare: This can be an argument to the graph.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+
+        if addr != self.BASE_ADDR:
+            name = os.path.join(addr.partition(self.BASE_ADDR)[-1][1:], rule)
+        else:
+            name = rule
+        return name
+
+    def detect_circular_dependency(self, rule, resolved=[], seen=[]):
+        seen.append(rule)
+        rule_obj = self.rule_name_to_rule[rule]
```
kaustubh-karkare, line 81: To improve the naming here: s/rule/rule_name/ , which allows you to s/rule_obj/rule/

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+        for dep in rule_obj.deps:
+            if dep not in resolved:
+                if dep in seen:
+                    raise Exception("Circular dependency detected!")
+                self.detect_circular_dependency(dep, resolved, seen)
```
kaustubh-karkare, line 86: Can you do this without recursion?

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+            if self.RULE_FILE in files:
+                with open(os.path.join(root, self.RULE_FILE)) as build_file:
+                    rule_json = json.load(build_file)
+                    for obj in rule_json:
+                        rule_name = self.generate_rule_name(root, obj["name"])
+                        if rule_name not in self.rule_name_to_rule:
+                            self.rule_name_to_rule[rule_name] = Rule(obj["name"])
+                        self.rule_name_to_rule[rule_name].create(obj, rule_name, self.rule_name_to_rule, root)
+
+    def generate_rule_name(self, addr, rule):
```
kaustubh-karkare: This logic also is specific to a rule, not the entire graph.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+                    visited.add(dep)
+        return independent_deps
+
+
+class SerialExe():
+
+    def __init__(self, graph):
+        self.graph = graph
+
+    def exe(self, rule, resolved=[], unresolved=[]):
```
kaustubh-karkare: * See [this issue](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). Instead of passing it around as an argument, you can just save it on the class: `self.resolved`
* Doesn't look like you're using `unresolved` anyway.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class ParallelExe():
+
+    def __init__(self, graph):
+        self.graph = graph
+        self.unresolved_deps = {rule: self.graph[rule].deps for rule in self.graph}
+        self.resolved_deps = {rule: list() for rule in self.graph}
+
+    def exe(self, rule):
```
kaustubh-karkare: s/rule/rule_name/

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+        returns a set of independent dependencies to form the base of parallel execution
+        '''
+        visited, stack = set(), [rule]
+        independent_deps = set()
+        visited.add(rule)
+        while stack:
+            vertex = stack.pop()
+            for dep in self.graph[vertex].deps:
+                if dep not in visited:
+                    stack.append(dep)
```
kaustubh-karkare, line 142: You're using it as a queue, not a stack.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+        '''
+        executes rules and their dependencies in parallel reducing execution time
+        '''
+        process_to_execute = self.get_independent_deps(rule)
+        name_to_process = {dep: self.graph[dep].exe_rule() for dep in process_to_execute}
+        while True:
+            for dep in list(name_to_process):
+                if name_to_process[dep].poll() is None:
+                    continue
+                else:
```
kaustubh-karkare: Don't need the `else`. You can avoid one layer of indenting for the following code.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+        self.resolved_deps = {rule: list() for rule in self.graph}
+
+    def exe(self, rule):
+        '''
+        executes rules and their dependencies in parallel reducing execution time
+        '''
+        process_to_execute = self.get_independent_deps(rule)
+        name_to_process = {dep: self.graph[dep].exe_rule() for dep in process_to_execute}
+        while True:
+            for dep in list(name_to_process):
```
kaustubh-karkare: s/dep/rule_name/

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+                if name_to_process[dep].poll() is None:
```
kaustubh-karkare: You probably also want to verify that the return value of poll() is 0 if not None.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+                else:
+                    dep_obj = self.graph[dep]
+                    del name_to_process[dep]
+                    process_to_execute.remove(dep)
+                    for parent in dep_obj.parents:
+                        self.unresolved_deps[parent].remove(dep)
+                        self.resolved_deps[parent].append(dep)
+                        if not self.unresolved_deps[parent]:
+                            process_to_execute.add(parent)
+                            if self.graph[parent].command:
```
kaustubh-karkare, line 121: What if this command is None? Shouldn't you trigger more dependents of that rule?

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+                            for parent in self.executor.graph[self.file_deps[file]].parents:
+                                self.executor.exe(parent)
+                            self.file_watch[file] = file_edit_time
+                    else:
+                        self.file_watch[file] = file_edit_time
+            time.sleep(1)
+
+    def map_file_command(self, rule):
+        '''maps files to commands to assist in implementing the watch feature'''
+        if rule in self.file_deps.values():
```
kaustubh-karkare: s/file_deps/filepath_to_rule_name/

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+            return
+        files = input("Enter depended files for %s(Press Enter if none): " % (rule)).split()
```
kaustubh-karkare, line 220: instead of taking this as input, this information should be part of the build.json file.
ie - each rule can be associated with the files it compiles

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def __init__(self, executor):
+        self.executor = executor
+        self.file_deps = dict()
+        self.file_watch = dict()
+
+    def watch_init(self):
+        if not self.file_deps: return
+        while True:
+            for root, _, files in os.walk(os.getcwd()): #base addr inaccesible from graph as the dictionary is being taken instead of the whole object, thus os.getcwd is used again
```
kaustubh-karkare: This isn't right. You shouldn't have more than one place in code making assumptions like this, because they can get out of sync.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+class WatchChanges():
+
+    def __init__(self, executor):
+        self.executor = executor
+        self.file_deps = dict()
+        self.file_watch = dict()
+
+    def watch_init(self):
+        if not self.file_deps: return
+        while True:
```
kaustubh-karkare: If this is infinite loop that is detecting changes, how can it be in an "init" method?
This logic to set the original modification-times for files is the "init" part of it, yes, but to compare it to new values of mtime is not "init". You should separate the logic into 2 parts.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,186 @@
...
+            for root, _, files in os.walk(os.getcwd()): #base addr inaccesible from graph as the dictionary is being taken instead of the whole object, thus os.getcwd is used again
+                for file in set(self.file_deps).intersection(set(files)):
+                    file_edit_time = os.stat(os.path.join(root, file)).st_mtime
+                    if file in self.file_watch:
+                        if self.file_watch[file] < file_edit_time:
+                            print(file, "changed...")
+                            print("recompiling files...")
+                            self.executor.exe(self.file_deps[file])
+                            for parent in self.executor.graph[self.file_deps[file]].parents:
+                                self.executor.exe(parent)
```
kaustubh-karkare: What if the parents also have parents that need to be recompiled?
Assuming only 2 levels isn't right here.

solutions/d-r-siddhartha/assignment-2/run_build.py
```diff
@@ -0,0 +1,25 @@
...
+    parser.add_argument('--parallel', action='store_true')
+    parser.add_argument("--watch", action='store_true')
+    args = parser.parse_args()
+
+    BuildGraph = build.Graph()
+    BuildGraph.create_graph()
+    BuildGraph.detect_circular_dependency(args.rule)
+
+    if args.parallel:
+        Executor = build.ParallelExe(BuildGraph.rule_name_to_rule)
```
kaustubh-karkare: s/Executor/executor

The python style guide says that names starting with capital letters are for classes, not class instances.

kaustubh-karkare (changes requested):

solutions/d-r-siddhartha/assignment-2/build.py
drs-11: but wouldn't having a map between file_name and associated rule be convenient and faster for executing dependent actions?

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+    def addDeps(self, dep):
+        self.dependencies.append(dep)
+
+    def exe_rule(self):
+        if self.command:
+            return subprocess.Popen((self.command), cwd=self.cwd, shell=True)
+        else:
+            return None
+
+    def generate_rule_name(self, base_addr):
```
kaustubh-karkare, line 25: This feels like it should be a private method.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+
+    @classmethod
+    def create(cls, obj, addr, base_addr):
+        rule_obj = cls(obj["name"])
+        rule_obj.cwd = addr
+        if "deps" in obj:
+            for dep in obj["deps"]:
+                rule_obj.addDeps(dep)
+        if "command" in obj:
+            rule_obj.command = obj["command"]
```
kaustubh-karkare, line 42: This is simpler:
```
rule_obj.dependencies.extend(obj.get("deps", []))
rule_obj.command = obj.get("command")
```
Check out the difference between accessing a key using [] notation vs "get"

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+        self.dependencies.append(dep)
+
+    def exe_rule(self):
+        if self.command:
+            return subprocess.Popen((self.command), cwd=self.cwd, shell=True)
+        else:
+            return None
+
+    def generate_rule_name(self, base_addr):
+        '''generates the name of the rule with a relative path'''
```
kaustubh-karkare, line 26: Instead of the relative path, how about always using absolute paths?

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+        for rule_name in self.rule_name_to_rule:
+            for dep_name in self.rule_name_to_rule[rule_name].dependencies:
+                self.rule_name_to_rule[dep_name].dependents.append(rule_name)
+
+
+    def detect_circular_dependency(self, rule, resolved=None, seen=None):
+        if type(resolved) != list:
+            resolved = list()
+        if type(seen) != list:
+            seen = list()
```
kaustubh-karkare, line 79: Since the default value is None, you can check for that explicitly.
```
if resolved is None:

File: solutions/d-r-siddhartha/assignment-2/build.py [hunk 53]
```diff
@@ -0,0 +1,234 @@
...
+        seen.append(rule)
+        rule_obj = self.rule_name_to_rule[rule]
+        for dep in rule_obj.dependencies:
+            if dep not in resolved:
+                if dep in seen:
+                    raise Exception("Circular dependency detected!")
+                self.detect_circular_dependency(dep, resolved, seen)
+        resolved.append(rule)
+        return False
```
kaustubh-karkare, line 88: You don't need to return anything here.
The contract is that if there is a problem, then an exception is raised, otherwise the function completes without problems.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+
+class ParallelExe():
+
+    def __init__(self, graph_obj):
+        self.graph_obj = graph_obj
+        self.graph = graph_obj.rule_name_to_rule
```
kaustubh-karkare, line 96: Try to use the same name as much as possible.
```
self.rule_name_to_rule = graph_obj.rule_name_to_rule
```
Alternatively, you could even do:
```
self.graph = graph
# later
something(self.graph.rule_name_to_rule)

File: solutions/d-r-siddhartha/assignment-2/build.py [hunk 55]
```diff
@@ -0,0 +1,234 @@
...
+        visited, stack = set(), [rule]
+        independent_deps = set()
+        visited.add(rule)
+        while stack:
+            vertex = stack.pop()
+            for dep in self.graph[vertex].dependencies:
+                if dep not in visited:
+                    stack.append(dep)
+                    if not self.graph[dep].dependencies:
+                        independent_deps.add(dep)
```
kaustubh-karkare, line 144: What if there is only 1 rule in the project?
If the vertex does not have dependencies, you should add it to independent_deps too.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+        self.graph_obj = graph_obj
+        self.graph = graph_obj.rule_name_to_rule
+        self.unresolved_deps = {rule: self.graph[rule].dependencies for rule in self.graph}
+        self.resolved_deps = {rule: list() for rule in self.graph}
+
+    def exe(self, rule_name):
+        '''
+        executes rules and their dependencies in parallel reducing execution time
+        '''
+        process_to_execute = self.get_independent_deps(rule_name)
```
kaustubh-karkare, line 104: Inaccurate name.

s/process_to_execute/independent_deps/

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+
+        while True:
+
+            for rule_name in list(name_to_process):
+                if name_to_process[rule_name].poll() is not None:
+                    if name_to_process[rule_name].poll() is not 0:
+                        raise subprocess.CalledProcessError
+                    rule = self.graph[rule_name]
+                    del name_to_process[rule_name]
+                    process_to_execute.remove(rule_name)
```
kaustubh-karkare, line 115: These 2 lines do the same thing?

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+                    for parent in rule.dependents:
+                        self.unresolved_deps[parent].remove(rule_name)
+                        self.resolved_deps[parent].append(rule_name)
```
kaustubh-karkare, line 118: You don't actually need to maintain the list of resolved_deps, since you never use it.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+                        if not self.unresolved_deps[parent]:
+                            process_to_execute.add(parent)
+                            if self.graph[parent].command:
+                                name_to_process[parent] = self.graph[parent].exe_rule()
+                            else:
+                                for dependent in self.graph[parent].dependents:
+                                    process_to_execute.add(dependent)
```
kaustubh-karkare, line 125: This is never used in the loop.

solutions/d-r-siddhartha/assignment-2/build.py
kaustubh-karkare: The current code still does not work.
How about something like:
```
name_to_process[parent] = self.graph[parent].exe_rule()
# even if rule.command is None, since that is handled by the Rule class
```
and on line 110, 
```
if name_to_process[rule_name] is None or name_to_process[rule_name].poll() is not None:
```
to handle both cases

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+class WatchChanges():
+
+    def __init__(self, executor):
+        self.executor = executor
+        self.filepath_to_rule_name = dict()
+        self.file_watch = dict()
+
+    def watch_changes(self):
+            while True:
+                for root, _, files in os.walk(self.executor.graph_obj.base_addr):
```
kaustubh-karkare, line 181: Typically the term "root" describes the root of the filesystem or the project or something, and generally is fixed.
In this case, "current_directory" is more accurate than root.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+            return
+        files = input("Enter depended files for %s(Press Enter if none): " % (rule)).split()
+
+        with open(os.path.join(self.executor.graph[rule].cwd, self.executor.graph_obj.RULE_FILE), 'r') as build_file:
+            rule_json = json.load(build_file)
+            for obj_index in range(len(rule_json)):
+                if rule_json[obj_index]["name"] == self.executor.graph[rule].name:
+                    rule_json[obj_index]["files"] = files
+                    for file in files:
+                        self.filepath_to_rule_name[file] = rule
```
kaustubh-karkare, line 228: Instead of re-opening the file, why not save the list of files originally while creating the graph?

solutions/d-r-siddhartha/assignment-2/build.py
kaustubh-karkare: > wouldn't having a map between file_name and associated rule be convenient and faster for executing dependent actions?

It will be, and that is how I did it too, which is fine.

I'm actually saying an unrelated thing: since the source of this information is build.json, we don't need to ask the user for it via this input method. You shouldn't modify that file either, since it is created and therefore owned by the user.

solutions/d-r-siddhartha/assignment-2/build.py
```diff
@@ -0,0 +1,234 @@
...
+    def __init__(self, executor):
+        self.executor = executor
+        self.filepath_to_rule_name = dict()
+        self.file_watch = dict()
+
+    def watch_changes(self):
+            while True:
+                for root, _, files in os.walk(self.executor.graph_obj.base_addr):
+                    if self.executor.graph_obj.RULE_FILE in files:
+                        with open(os.path.join(root, self.executor.graph_obj.RULE_FILE)) as build_file:
```
kaustubh-karkare, line 183: There are 2 options here:
* What you've done is to re-fetch all build.json files in each loop, and then look at each file.
* Alternatively, explore `build.json` files only once, since the build rules for a project change very infrequently compared to the rest of the code, and then only look at the files.
Both are fine I guess.

solutions/d-r-siddhartha/assignment-2/unittest_build.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def test_serial_build(self):
+        '''
+        Testing basic build functionality. The build tool should complete the actions of the dependencies
+        before executing the root rule.
+        '''
+
+        RULE = "test_sort_merge"
+
+        with tempfile.TemporaryDirectory() as TempDir:
```
kaustubh-karkare, line 20: Python style says that variable names are lowercase with underscores, while class names are camel-case.
```
with tempfile.TemporaryDirectory() as temp_dir_name:

File: solutions/d-r-siddhartha/assignment-2/unittest_build.py [hunk 64]
```diff
@@ -0,0 +1,58 @@
...
+            copy_tree(self.BASE_ADDR, TempDir)
+            os.chdir(TempDir)
+            build_tool_graph = build.Graph()
```
kaustubh-karkare, line 23: This should work too:
```
graph = build.Graph(temp_dir_name)
```
That way, you don't ever need to call `os.chdir`

solutions/d-r-siddhartha/assignment-2/unittest_build.py
```diff
@@ -0,0 +1,58 @@
...
+        RULE = "test_all"
+
+        with tempfile.TemporaryDirectory() as TempDir:
+            copy_tree(self.BASE_ADDR, TempDir)
+            os.chdir(TempDir)
+            build_tool_graph = build.Graph()
+            build_tool_graph.create_graph()
+            executor = build.ParallelExe(build_tool_graph)
+            executor.exe(RULE)
+            files = ["test.o","algorithms/sort_merge.o", "algorithms/sort_bubble.o", "algorithms/sort_quick.o", "test_sort_bubble.exe", "test_sort_quick.exe", "test_sort_merge.exe"]
```
kaustubh-karkare, line 43: Why are the rest of these things being built if the only rule being executed is `test_sort_merge`?

solutions/d-r-siddhartha/assignment-2/unittest_build.py
```diff
@@ -0,0 +1,58 @@
...
+            for file in files:
```
kaustubh-karkare, line 44: Instead of a long line, this is more readable:
```
for file_name in [
  "test.o",
  ...
]
```
Also note the better name. "file" can refer to the file handle or the name, so always prefer being more specific.

solutions/d-r-siddhartha/assignment-2/unittest_build.py
```diff
@@ -0,0 +1,58 @@
...
+            build_tool_graph = build.Graph()
+            build_tool_graph.create_graph()
+            executor = build.SerialExe(build_tool_graph)
+            executor.exe(RULE)
+            files = ["test.o", "algorithms/sort_merge.o", "test_sort_merge.exe"]
+            for file in files:
+                self.assertTrue(os.path.exists(os.path.join(TempDir, file)))
+            os.chdir(self.BASE_ADDR)
+
+    def test_parallel_build(self):
```
kaustubh-karkare, line 32: This test doesn't really prove that things happen in parallel, right?
Ie - if you used the SerialExecutor was used, it would still work.
How can you modify it so that SerialExecutor would fail, but ParallelExecutor would pass?

kaustubh-karkare (approved):
I think you've now reached a reasonable point on this problem.
You can continue working on it, or we could move on now.

PR #66, opened 2019-10-28
Solution for Assignment2.

Abhilasha06:


solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,117 @@
+import os
+import json
+import subprocess
+import threading
+import time
+
+class Action():
```
kaustubh-karkare: You have these classes called Action and ActionGraph, but I still don't quite understand the difference in responsibilities between them. Can you update the code so that:
* Action represents a single parsing a single json object / running a single command. While writing that code, you cannot make any assumptions about there being multiple actions.
* ActionGraph contains ALL actions across ALL directories, and deals with operations that span multiple actions. It uses the Action class to perform underlying operation.

You should be able to do things like:
```
action = Action(raw_data)
action.get_dependencies()
action.start_command()
# the above methods are only invoked from inside ActionGraph

graph = ActionGraph(root_dir)  # recursively explores root_dir to load all build.json files
and it creates a single map<name, Action>
graph.execute(action_name)  # execute dependencies and then execute specified action
# the above methods are invoked by your main section
```

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,117 @@
...
+        
+     def execute_commands(self):
+       
+        thread_list = []
+        count = 0
+        for command, work_dir in zip(self.commands, self.working_directories):
+         
+            thread = threading.Thread(target = self.process, args = (command, work_dir))
+            thread_list.append(thread)
+            time.sleep(2)
```
kaustubh-karkare: Why do you have to sleep?

solutions/Abhilasha/Assignment2/btest.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class TestBuildAutomationTool(unittest.TestCase):
+    
+    
+	def test_build_test_all(self):
+
+		BAT_obj = BAT.Action()
+
+		BAT_obj.get_command(os.getcwd(), 'build', 'test_all')
```
kaustubh-karkare: If I run 2 copies of this test in parallel, they will interfere with each other.
Can you use the `tempfile` module to create a new unique directory inside /tmp, copy the relevant code in there, and then run the commands using that new directory as the root?

kaustubh-karkare (changes requested):
Cool, good start!
Next step is to re-organize your code a bit to simplify the design, without any change in functionality, so that it becomes easy to add the next feature while respecting "modularity".

solutions/Abhilasha/Assignment2/btest.py
```diff
@@ -0,0 +1,127 @@
...
+
+class TestBuildAutomationTool(unittest.TestCase):
+    
+    
+    
+    def test_build_test_all(self):
+        
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            
+            shutil.copytree("C://Users//Abhilasha//code",tmpdirname+"//code")
```
kaustubh-karkare: * This path is hardcoded to your machine, and so is not reusable. You probably want to put the code directory inside this repository, and then use relative paths, so that anyone can clone the repo and run the test.
* Also, the "//" is true for windows, but not linux. Use `os.path.sep` instead.
* Since this tempfile stuff is common across tests, you could put it in the setUp method. [Relevant documentation](https://docs.python.org/3.7/library/unittest.html#unittest.TestCase.setUp).

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
kaustubh-karkare: I think this is still pending. Based on the above, I was hoping to see something like the following in your tests:
```
graph = ActionGraph("wherever/code")
# the above creates all the Action objects internally, so the user never deal with them directly

graph.execute("build")
graph.execute("test_all")
```

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,174 @@
...
+         
+        self.commands = []
+
+        self.working_directories = []
+        
+        self.orders = []
+        
+  
+
+     def plot_graph(self, text, cwd, order):
```
kaustubh-karkare: This name suggests that you're still thinking of mathematical graphs with x/y-axes, etc.
In case of CS graphs, create/process/execute are acceptable terms too, depending on what this is doing.
If you want be even more precise in this case, you can use a name like "def compute_execution_order" or something. As usual, the goal is to figure out exactly what the method does, based on only the name.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,174 @@
...
+
+        if 'deps' in text:
+
+            for dependency in text['deps']:
+
+                if len(str(dependency).split('/')) == 2:
+
+                    dependency = str(dependency).split('/')[1]
+
+                    with open(os.path.join("algorithms//build.json")) as json_file:
```
kaustubh-karkare: This path is hard-coded, and therefore not allowed.
The assumption you're allowed to make is that the file name containing the actions is called "build.json". You can save that in some constant if you wish, but then you have to compute the rest of the location based on the context.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,174 @@
...
+     def process(self, command, work_dir):
+         
+        os.chdir(work_dir)
+
+        os.system(command)
+
+        return
+
+            
+     def execute_commands(self):
```
kaustubh-karkare: Ohh, interesting.
You're doing this in batches of actions with the same order value!
This is definitely better than the original approach where everything was serial.

The case where it doesn't work is this:
```
A (1s) -> B (1s) -> C (1s) -> D (1s) -> E (1s) -> Final <- G (5s)
```
So while G was running, we could have run A-E also, since they take 5s overall. 
I think in your current scheme, E & G will happen in parallel, but before that, A-D will happen separately.
Can you figure out a way to support that?

kaustubh-karkare (changes requested):

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,144 @@
...
+    def __init__(self, root_dir):
+        
+        self.root_dir = root_dir
+        self.all_action_status = {}
+        self.all_action_commands = {}
+        self.all_action_dependencies = {}
+        
+    def create_action_map(self):
+        
+        for filename in Path(self.root_dir).rglob('*.json'):
```
kaustubh-karkare: * Oh, nice! I didn't know about this way to do it.
* You care specifically about "build.json" files, right? Not all json files though

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,144 @@
...
+            rel_location = relpath(filename,self.root_dir)
+            if len(str(rel_location).split(os.path.sep)) == 1:
+                rel_location = ''
+            else:
+                rel_location = str(rel_location).split(os.path.sep)[0]
```
kaustubh-karkare: Instead of splitting the same thing twice, you can just share the result.
```
path_elements = str(rel_location).split(os.path.sep)
```
Also, you don't just want the first element, you want everything except the last element: path_elements[:-1]

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,144 @@
...
+            with open(filename) as json_file:
+                all_data = json.load(json_file)
```
kaustubh-karkare, line 42: This is the only line that needs to be inside the "with" block.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,144 @@
...
+                for data in all_data:
+                    if rel_location != '':
+                        action = rel_location+'/'+data['name']
+                    else:
+                        action = data['name']
+                    self.all_action_status[action] = 'not started'
+                    if 'command' in data:
+                        self.all_action_commands[action] = data['command']
+                    else:
+                        self.all_action_commands[action] = 'none'
```
kaustubh-karkare:
```
self.all_action_commands[action] = data.get('command')
```
That way, you're not dealing with a "none" string, but a proper None object.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,144 @@
...
+                        self.all_action_dependencies[action] = data['deps']
+                    else:
+                        self.all_action_dependencies[action] = 'none'
+
+            
+    def get_command_to_be_executed(self, build, action):
+        
+        if action not in self.all_action_status:
+            raise Exception('Command not recognized.')
+            return
```
kaustubh-karkare: Will never reach this line due to `raise` on the previous one.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,144 @@
...
+                        self.all_action_commands[action] = data['command']
+                    else:
+                        self.all_action_commands[action] = 'none'                
+                    if 'deps' in data:
+                        self.all_action_dependencies[action] = data['deps']
+                    else:
+                        self.all_action_dependencies[action] = 'none'
+
+            
+    def get_command_to_be_executed(self, build, action):
```
kaustubh-karkare: What is this build argument used for?

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,144 @@
...
+                     loc = name[:name.rindex('/')]                     
+                 else:                     
+                     loc = ''   
+                                      
+                 if self.current_action_commands[name] != 'none':
+                     if loc != '':
+                         os.chdir(cwd+os.path.sep+loc)
+                     command = str(self.current_action_commands[name])
+                     p = subprocess.Popen(command, shell=True)
+                     os.chdir(cwd)
```
kaustubh-karkare: You can avoid the manual changing of directories this way:
```
p = subprocess.Popen(command, shell=True, cwd=cwd)
```

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,162 @@
...
+                     continue
+                     
+             if any_subprocess_completed == True:
+                 self.ongoing_subprocesses.remove(p)
+                 self.action_name_for_ongoing_subprocess.remove(action)
+                 self.current_action_status[action] = 'done'
+                 for name in self.current_action_status:
+                     if(self.current_action_status[name] == 'not started'):     
+                         self.update_dependencies_status(name)                             
+                 self.execute_commands()
```
kaustubh-karkare: Okay, so this line confused me. I suppose it works, but it is non-intuitive.
You want to start the next action (ie - run lines 116 to 139), which is why I think you're calling the method recursively. However, that method will contain its own copy of this while loop, which is unnecessary.
You could make this easier to read by reorganizing the code to something like:
```
while there are pending actions:
    start any actions that are ready
    while True:
        if any ongoing action is completed:
            break
        time.sleep(1)
    mark new actions as ready
```

solutions/Abhilasha/Assignment2/btest.py
```diff
@@ -0,0 +1,93 @@
...
+        
+        with self.test_dir as tmpdirname:
+            new_root_dir = tmpdirname+os.path.sep+"code"
+            shutil.copytree(self.root+os.path.sep+"code", new_root_dir )
+            os.chdir(new_root_dir)
+            BAT_obj = BAT.ActionGraph(new_root_dir)
+            BAT_obj.create_action_map()
+            BAT_obj.get_command_to_be_executed('build', 'test_sort_merge')       
+            self.assertEqual(path.exists('test.o'), True)
+            self.assertEqual(path.exists('test_sort_merge.exe'), True)
```
kaustubh-karkare, line 58: You can verify that other exes are not generated.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,162 @@
...
+                 else:
+                     continue
+                     
+             if any_subprocess_completed == True:
+                 self.ongoing_subprocesses.remove(p)
+                 self.action_name_for_ongoing_subprocess.remove(action)
+                 self.current_action_status[action] = 'done'
+                 for name in self.current_action_status:
+                     if(self.current_action_status[name] == 'not started'):     
+                         self.update_dependencies_status(name)
```
kaustubh-karkare: This seems somewhat inefficient. Ie - update_dependencies_status already contains recursion, so will update multiple items anyway. Looping over all of them results in duplicate work.

I think the proper fix is to also keep track of "dependents" (ie - the other direction of dependencies). That way, you just have to update the dependents of the completed action via recursion, and can avoid the loop.

solutions/Abhilasha/Assignment2/btest.py
```diff
@@ -0,0 +1,93 @@
...
+    def TearDown(self):
+        
+        shutil.rmtree(self.test_dir)
+    
+    
+    def test_build_test_all(self):
+        
+        with self.test_dir as tmpdirname:
+            new_root_dir = tmpdirname+os.path.sep+"code"
+            shutil.copytree(self.root+os.path.sep+"code", new_root_dir )
```
kaustubh-karkare: spacing here is inconsistent, did you use a linter?

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,162 @@
...
+            return        
+        else:
+            Action_obj = Action(self.all_action_status, self.all_action_commands, self.all_action_dependencies)
+            Action_obj.get_dependencies(action)
+            Action_obj.update_dependencies_status(action)
+            Action_obj.execute_commands()
+            return 
+                    
+              
+class Action():
```
kaustubh-karkare, line 53: This isn't the correct name.
An action is a specific thing you can execute.
This class is responsible for executing multiple actions.
So "ActionExecutor" might be a better name.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,162 @@
...
+    
+
+     def __init__(self, all_action_status, all_action_commands, all_action_dependencies):
+     
+        self.all_action_status = all_action_status
+        self.all_action_commands = all_action_commands
+        self.all_action_dependencies = all_action_dependencies
```
kaustubh-karkare: Alternatively, you could just store a single reference to the action_graph, and then do `self.graph.action_commands`

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,202 @@
...
+        self.root_dir = root_dir
+        self.all_action_status = {}
+        self.all_action_commands = {}
+        self.all_action_dependencies = {}
+        
+    def create_action_map(self):
+        
+        for filename in Path(self.root_dir).rglob('*.json'):
+            
+            if str(filename).endswith('\\build.json'):
```
kaustubh-karkare: Why not just glob for build.json files directly?

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,202 @@
...
+                path_elements = str(relpath(filename, self.root_dir)).split(os.path.sep)          
+                rel_location = os.path.sep.join(path_elements[:-1])
```
kaustubh-karkare: You can also use `os.path.split` and `os.path.join`, which already assume that the separator is `os.path.sep`.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,202 @@
...
+                with open(filename) as json_file:
+                    all_data = json.load(json_file)
+                for data in all_data:
+                    if rel_location != '':
+                        action = rel_location+'/'+data['name']
+                    else:
+                        action = data['name']
+                    self.all_action_status[action] = 'not started'
```
kaustubh-karkare: Technically, the concept of `status` is not relevant to a graph. It only matters to the executor and so should not be a property of this object.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,202 @@
...
+    def get_command_to_be_executed(self, build, action):
+        
+        if build != 'build':
+            raise Exception('Command not recognized.')
+            
+        if action not in self.all_action_status:
+            raise Exception('Command not recognized.')
+                  
+        
+        Action_obj = ActionExecutor(self.all_action_status, self.all_action_commands, self.all_action_dependencies)
```
kaustubh-karkare: This is a method of the ActionGraph object depends on the ActionExecutor.
But ActionExecutor technically depends on the data in the ActionGraph?
Circular dependencies like this should be avoided.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
kaustubh-karkare: ^

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,202 @@
...
+        Action_obj.get_dependencies(action)
+        Action_obj.store_dependents_for_all_action()
+        Action_obj.initialize_dependencies_status(action)
+        Action_obj.initialize_pending_actions()
+        Action_obj.execute_commands()
```
kaustubh-karkare: One of the goals for better design is to make class objects easy to use, without having to understand how it works internally. My concern here is that the knowing that we must call this sequence of 5 methods is not intuitive. It is more useful to see something like:
```
executor = ActionExecutor(graph)
executor.execute(action_name)
```

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,202 @@
...
+            self.get_dependents(action)
+        return
+                       
+            
+    def get_dependents(self, action):
+         
+         if not self.all_action_dependencies[action] is None:                 
+            for dep in self.all_action_dependencies[action]:
+                if self.action_name not in self.dependents[dep]:
+                    self.dependents[dep].append(self.action_name)
```
kaustubh-karkare: I thought you were going to store dependent information in the Action class.
Event if you don't have a specific Action class, the dependents is a feature of the graph, not the executor, since it does not change during the execution.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,202 @@
...
+                 no_of_dependencies_done += 1
+         if no_of_dependencies_done == total_no_of_dependencies:
+             self.current_action_status[action] = 'ready'
+         return
+     
+     
+    def execute_commands(self):
+           
+        while len(self.pending_actions) != 0:
+            action_has_no_command = False
```
kaustubh-karkare: To be a little more clear `any_action_has_no_command`

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,202 @@
...
+            for name in self.current_action_status:
+                 if self.current_action_status[name] == 'ready':
+                     self.current_action_status[name] = 'processing'
```
kaustubh-karkare: Also, instead of using strings for this, you should use an enumeration, which prevents any typos in the string values.

https://docs.python.org/3/library/enum.html

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,202 @@
...
+                 for p, action in zip(self.ongoing_subprocesses, self.action_name_for_ongoing_subprocess):
+                     if not p.poll() is None:
+                         any_subprocess_completed = True
+                         break     
+                 if any_subprocess_completed:
+                     break
+                 time.sleep(1)
+                            
+            self.ongoing_subprocesses.remove(p)
+            self.action_name_for_ongoing_subprocess.remove(action)
```
kaustubh-karkare: Hmm, you're relying on the values of p & action from above. I guess this works.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,202 @@
...
+                         p = subprocess.Popen(command, shell=True, cwd=cwd)
+                         self.ongoing_subprocesses.append(p)
+                         self.action_name_for_ongoing_subprocess.append(name)
+                         
+            if action_has_no_command:
+                continue
+               
+            while True:                
+                 any_subprocess_completed = False                
+                 for p, action in zip(self.ongoing_subprocesses, self.action_name_for_ongoing_subprocess):
```
kaustubh-karkare: Instead of `p`, use `process`. Always go for useful names.

solutions/Abhilasha/Assignment2/btest.py
```diff
@@ -0,0 +1,97 @@
...
+            self.assertEqual(path.exists('test_sort_quick.exe'), True)
+            self.assertEqual(path.exists('test_sort_merge.exe'), True)
+            os.chdir(os.path.join("algorithms"))
+            self.assertEqual(path.exists('sort_bubble.o'), True)
+            self.assertEqual(path.exists('sort_quick.o'), True)
+            self.assertEqual(path.exists('sort_merge.o'), True)
+            os.chdir(self.root)
+            
+            
+    def test_clean(self):
```
kaustubh-karkare: I think you might want to merge this test with the one above.
Ie - when you copied the code directly, I don't think it contains the .o or .exe file anyway, so executing the clean command does nothing, which doesn't provide that much value, right?

solutions/Abhilasha/Assignment2/btest.py
```diff
@@ -0,0 +1,97 @@
...
+        shutil.rmtree(self.test_dir)
+    
+    
+    def test_build_test_all(self):
+        
+        with self.test_dir as tmpdirname:
+            new_root_dir = tmpdirname+os.path.sep+"code"
+            shutil.copytree(self.root+os.path.sep+"code", new_root_dir) 
+            os.chdir(new_root_dir)
+            BAT_obj = BAT.ActionGraph(new_root_dir)
```
kaustubh-karkare: Instead of `BAT_obj`, why not just call this "action_graph" ?

kaustubh-karkare (changes requested):

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,237 @@
...
+                self.all_action_commands[action] = data.get('command')
+                self.all_action_dependencies[action] = data.get('deps')  
+        return self.all_action_commands, self.all_action_dependencies
+
+            
+class Action():
+    
+    def __init__(self):
+
+        self.current_action_status = {}
```
kaustubh-karkare, line 57: The action status should be stored in the ActionExecutor, not here, since that is the class that cares about it.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,237 @@
...
+        self.current_action_commands = {}
+        self.current_action_dependencies = {}
+        self.dependents = {}
+        
+    
+    def action_sequence(self, root, action):
+            
+        graph = ActionGraph()
```
kaustubh-karkare, line 65: Action = single item
ActionGraph = a collection of actions.

ActionGraph can use Action, but Action cannot use ActionGraph like this.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,237 @@
...
+                if self.action_name not in self.dependents[dep]:
+                    self.dependents[dep].append(self.action_name)
+                self.get_dependents(dep)
+
+         return
+
+     
+    def initialize_dependencies_status(self, action):      
+
+         ''' initialize the status of the actions with no deps as ready '''
```
kaustubh-karkare, line 120: As mentioned above, the status is relevant only for ActionExecutor.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,237 @@
...
+        self.ongoing_subprocesses = []
+        self.action_name_for_ongoing_subprocess = []
+        self.action_name = ''
+        self.pending_actions = []
+
+        
+    def execute(self, build, action):
+
+        if build != 'build':
+            raise Exception('Command not recognized.')
```
kaustubh-karkare, line 146: Once again, why do you need this at all?

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,237 @@
...
+                         any_action_has_no_command = True
+                         self.current_action_status[name] = Status.done
+                         self.pending_actions.remove(name)
+                         if len(self.dependents[name]) != 0:
+                             for action in self.dependents[name]:
+                                 if self.current_action_status[action] == Status.not_started:
+                                     self.update_dependencies_status(action)
+                         break
+                     else:
+                         process = subprocess.Popen(command, shell=True, cwd=cwd)
```
kaustubh-karkare, line 208: There's a lot of custom logic here, that the executor shouldn't care about. I would suggest adding a "start" method to your Action class.
Eg - what if you have another type of executor class (which does things serially, instead of in parallel). You don't want to duplicate logic in that one.

solutions/Abhilasha/Assignment2/BuildAutomationTool.py
```diff
@@ -0,0 +1,237 @@
...
+                         self.ongoing_subprocesses.append(process)
+                         self.action_name_for_ongoing_subprocess.append(name)
```
kaustubh-karkare, line 210: Instead of separate lists, you can just use a single dict.
```
self.action_name_to_ongoing_process[action_name] = process
```

solutions/Abhilasha/Assignment2/btest.py
```diff
@@ -0,0 +1,81 @@
...
+        self.test_dir = tempfile.TemporaryDirectory()
+
+    def tearDown(self):
+
+        shutil.rmtree(self.test_dir)
+
+    def test_build_test_all_and_test_clean(self):
+
+        with self.test_dir as tmpdirname:
+            new_root_dir = tmpdirname+os.path.sep+"code"
```
kaustubh-karkare, line 23: Use `os.path.join`, instead to directly concatenating.

kaustubh-karkare (changes requested):
I think there is still some confusion about the responsibilities of each class.

PR #67, opened 2020-01-15
Assignment 3

drs-11:
not ready part 1 solution


solutions/assignment-1/CLIparser.py
```diff
@@ -0,0 +1,99 @@
+"""
```
kaustubh-karkare: Can remove these files.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,71 @@
...
+    def download(self, url):
+        self.send(url)
+        self.receive()
+        self.save_file()
+        self.stop()
+
+    @staticmethod
+    def find_server_and_resouce(url):
+        server_addr, slash, resource_addr = url.partition('//')[-1].partition('/')
+        return resource_addr, server_addr
```
kaustubh-karkare: Why change the ordering?

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def download(self, url):
+        self.send(url)
+        self.receive()
+        self.save_file()
+        self.stop()
+
+    @staticmethod
+    def find_server_and_resouce(url):
+        server_addr, slash, resource_addr = url.partition('//')[-1].partition('/')
```
kaustubh-karkare: * Assert that the part before "://" is "http" ?
* I'd actually recommend using a regex here.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def connect(self, host, port=80):
+        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.sock.connect((host, self.port))
+
+    def stop(self):
+        self.sock.close()
+
+    def generate_request(self, resource_addr, server_addr):
+        GET_request = ("GET /%s HTTP/1.1\r\nHost: %s\r\nAccept: */*\r\n\r\n" % (resource_addr, server_addr)).encode()
```
kaustubh-karkare: Use multiline string here maybe. Or even better:
```
"".join(["%s: %s\r\n" % (key, value) for key, value in headers.items()])
```

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,71 @@
...
+
+class Client():
+
+    def __init__(self, host=None, port=80):
+        self.host = host
+        self.port = 80
+
+    def connect(self, host, port=80):
+        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.sock.connect((host, self.port))
```
kaustubh-karkare: The port could be part of the URL too.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,71 @@
...
+    def send(self, url):
+        resource_addr, server_addr = self.find_server_and_resouce(url)
+        self.connect(server_addr)
+        request = self.generate_request(resource_addr, server_addr)
+        self.sock.sendall(request)
+
+    def receive(self    ):
+        header_data = self.get_response_headers()
+        content_length = self.get_content_length(header_data)
+        payload = self.get_payload(content_length)
```
kaustubh-karkare: Instead of downloading everything into memory (what if the file is massive?), write to file as you read from socket.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,79 @@
...
+    def _generate_request(self, resource, server_addr):
+        GET_request = ("GET %s HTTP/1.1\r\n"
+                       "Host: %s\r\n"
+                       "Accept: */*\r\n\r\n" % (resource, server_addr)).encode()
+        return GET_request
+
+
+    @staticmethod
+    def _find_server_and_resouce(url):
+        regex_pattern = '^(?:http://)([^:\/?]+)(?:[\:]?)([0-9]*)(?:\/?)([\S]*)$'
```
kaustubh-karkare: Hmm, not sure if this would always work. I was expecting something more like:
`^((\w+)://)?([^:\/?]+)(?:(\d+))([^?]*)(.*)`
where group 1 = protocol, 2 = hostname, 3 = port, 4 = path, 5 = query string

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,79 @@
...
+class Downloader():
+
+    def __init__(self, url, save_as):
+        self.host, self.port, self.resource = self._find_server_and_resouce(url)
+        if not self.port:
+            self.port = 80
+        else:
+            self.port = int(self.port)
+        if not self.resource:
+            self.resource = '/'
```
kaustubh-karkare: You should do this stuff in _find_server_and_resouce too.
Ie - let that me the only one that understands urls and defaults.
This method should not need to know the details.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,79 @@
...
+        self.url = url
+        self.save_as = save_as
+
+    def start(self):
+        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.sock.connect((self.host, self.port))
+        self._send()
+        self._receive()
```
kaustubh-karkare: This isn't a "start" method, it is a "do everything" method :)
Create a separate thread that starts downloading, and allow the stop method to cancel that ongoing download.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,79 @@
...
+        response_msg = ''
+        msg_buffer = self.sock.recv(1)
+        response_msg += msg_buffer.decode()
+        #print(response_msg)
+        while response_msg[-4:] != '\r\n\r\n':
+            msg_buffer = self.sock.recv(1)
+            response_msg += msg_buffer.decode()
+        return response_msg
+
+    def _get_content_length(self, header_data):
```
kaustubh-karkare: Instead of a specialized method for one particular header, why not load all headers into a dict, and just do a key lookup?

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, interface, port):
+        self.interface = interface
+        self.port = port
+        self.homepage = 'http://' + interface + ':' + str(port)
+
+    def start(self):
+        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        self.sock.bind((self.interface, self.port))
+        self._handle_request()
```
kaustubh-karkare: Like before, do this in a separate thread :)

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def stop(self):
+        print("Exiting...")
```
kaustubh-karkare: Instead of `print`, can you use the logging module, and only print if the CLI has a --verbose/-v option?

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,117 @@
+import socket
+import os
+import re
+
+class Server():
```
kaustubh-karkare: Technically, this is a FileServer, just like that other module was a FileDownloader.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,117 @@
...
+        headers = request_data.strip().split('\r\n')
+        get, self.resource,  http_version = headers.pop(0).split()
+        assert (http_version == 'HTTP/1.1')
+        self.headers = dict()
+        for header_line in headers:
+            header, data = header_line.split(': ')
+            self.headers[header] = data
+
+    def _process_request(self):
+        path = os.path.join(os.getcwd(), self.resource.strip('/'))
```
kaustubh-karkare: Instead of os.getcwd(), take the location as CLI input, and use this only as a default if that argument is not provided.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,117 @@
...
+            else:
+                print("generating directory listing and sending...")
+                send_data = self._generate_index_of(self.resource).encode()
+                resp_status_msg = '200 OK'
+
+        elif os.path.isfile(self.resource.strip('/')):
+            if os.path.exists(path):
+                with open(path, 'rb') as upload_file:
+                    print("sending file...")
+                    send_data = (upload_file.read())
```
kaustubh-karkare: Lets avoid reading the entire file.
Define a constant at the top of your file, like:
```
BLOCK_SIZE = 256 * 1000 # 256kb
```
and then you can only read/write/send that much at a time.
How will you restructure the code to support that?

It is fine to assume that headers will fit inside that, but the rest of the file will not.

-----

Using this, can you implement a per-connection speed limit?
The integration test that I'm hoping to see is:
* Create a dummy file is 4mb data.
* Start a FileServer(speed_limit_bps=256 * 1000) which is 256kb/second, so normally takes 16 seconds to download it.
* Start a FileDownloader(threads=4), which does a HEAD request to get the overall file size, figures out the ranges, and then does 4 range requests to complete the download in 4 seconds.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,117 @@
...
+                         '<ul>\n'
+                         '</ul>\n'
+                         '<hr>\n'
+                         '</body>\n'
+                         '</html>' % (self.resource))
+
+        for content in os.listdir('.' + location):
+            link = os.path.join(self.homepage, self.resource, content)
+            listing = '<li><a href=%s>%s</a></li>' % (link, content)
+            seek = pattern_obj.search(body_template).start()
```
kaustubh-karkare: This usage of regex is kinda inefficient.
Why not put all of your list items in an array, and then later:
`body_template = "<ul>%s</ul>" % "".join(list_items)`

kaustubh-karkare (changes requested):
Cool, this is a good start!

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,156 @@
...
+        self.homepage = 'http://' + interface + ':' + str(port)
+        self.server_location = server_location
+        self.logger = logger
+
+    def start(self):
+        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        self.sock.bind((self.interface, self.port))
+
+    def run(self):
```
kaustubh-karkare: It feels weird that this is a separate method.
You can start a new thread from your `start` method, directly invokes `_handle_request` below.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,156 @@
...
+        try:
+            self._handle_request()
+        except KeyboardInterrupt:
```
kaustubh-karkare: Do not assume that KeyboardInterrupts are the only reason we want to stop.
Here's what I like to do:
```
signal.signal(signal.SIGINT, file_server.stop)
```
which allows me to trigger it in a more general way.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,156 @@
...
+            self.sock.close()
+
+    def stop(self):
+        self.logger.info("Exiting...")
+        self.sock.shutdown(socket.SHUT_RDWR)
+        self.sock.close()
+
+    def _handle_request(self):
+        self.sock.listen(1)
+        while True:
```
kaustubh-karkare: If you do `while some_flag:`, you could make your stop method set that flag to False.
I think you'll need to use a [non-blocking accept](https://stackoverflow.com/questions/5308080/python-socket-accept-nonblocking) in that case though.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,156 @@
...
+                self.stop()
+                break
+            request_received = self._recvall()
+            #print("Response received..")
+            self._parse_request(request_received)
+            self.logger.info('%s requested...' % (self.resource))
+            if "Range" in self.headers:
+                self._process_range_request()
+            else:
+                self._process_request()
```
kaustubh-karkare: Instead of doing this in the same method, start a different thread here, so that you can immediately accept the next new connection.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def _parse_request(self, request_data):
+        headers = request_data.strip().split('\r\n')
+        get, self.resource,  http_version = headers.pop(0).split()
+        assert (http_version == 'HTTP/1.1')
+        self.headers = dict()
```
kaustubh-karkare: This is a server object, which should be handle multiple clients.
The headers dict is specific to one particular client, and so should not be on the server, since it does not allow multithreading. You can create a separate class to deal with client specific details.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,156 @@
...
+        for header_line in headers:
+            header, data = header_line.split(': ')
+            self.headers[header] = data
+
+    def _process_request(self):
+        is_big_file = False
+
+        path = os.path.join(self.server_location, self.resource.strip('/'))
```
kaustubh-karkare: Why do you `.strip('/')` ?
If you're consistently doing it, you should only do it once at the beginning.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,156 @@
...
+                is_big_file = True
+                content_length = os.stat(path).st_size
+            else:
+                self.logger.info("file not found, sending error response...")
+                send_data = b''
+                resp_status_msg = '404 Not Found'
+        else:
+            self.logger.info("Bad Request")
+            send_data = b''
+            resp_status_msg = '400 Bad Request'
```
kaustubh-karkare: I think this is also a 404 case, right?

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,156 @@
...
+            if not data_chunk:
+                break
+            self.conn_obj.sendall(data_chunk)
+            time.sleep(1)
+        file_handler.close()
+
+    def _generate_response_header(self, content_length, resp_status_code):
+        header = ('HTTP/1.1 %s\r\n'
+                  'Server: server.py\r\n'
+                  'Content-Length: %s\r\n\r\n' % (resp_status_code, content_length))
```
kaustubh-karkare: This should be constructed like `"".join("%s: %s\r\n" % (key, value) for key, value in headers.items())` too.

solutions/assignment-3/server.py
drs-11: what if I put the signal.signal() call here instead putting in the main function?

kaustubh-karkare: Can you help me understand how that would work?

drs-11: The signal.signal(signal.SIGINT, self.stop) call will be inside the while loop. The self.stop method will shutdown the socket and set the loop flag to False thereby breaking the while loop.

kaustubh-karkare: You only need to call signal.signal once. You're telling the library that whenever that event happens, invoke this method, and you don't need to keep telling it repeatedly, otherwise the library might call it repeatedly when once was enough.

drs-11: okay

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,132 @@
...
+    response_msg += msg_buffer.decode()
+    while response_msg[-4:] != '\r\n\r\n':
+        msg_buffer = sock.recv(1)
+        response_msg += msg_buffer.decode()
+    return response_msg
+
+def parse_response(response_data, expected_resp_code):
+    headers = response_data.strip().split('\r\n')
+    http_version, resp_code, resp_status_msg = headers.pop(0).split(" ", 2)
+    assert (resp_code == expected_resp_code), 'Expected %s HTTP response code, got %s.\nResponse Status Message: %s\n' % (expected_resp_code, resp_code, resp_status_msg)
```
kaustubh-karkare: This method is responsible for parsing stuff, it should not be doing other things like assertions.
You can modify this method to `return status_code, headers`, and do the assertion at the calling location.

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,132 @@
...
+    assert (http_version == 'HTTP/1.1')
+    parsed_headers = dict()
+    for header_line in headers:
+        header, data = header_line.split(': ')
```
kaustubh-karkare: `.split(": ", maxsplit=1)`

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,132 @@
...
+
+def parse_request(request_data):
+    headers = request_data.strip().split('\r\n')
+    req_method, resource,  http_version = headers.pop(0).split()
+    assert (http_version == 'HTTP/1.1')
+    recv_headers = dict()
+    for header_line in headers:
+        header, data = header_line.split(': ')
+        recv_headers[header] = data
+    return (req_method, resource, http_version, recv_headers)
```
kaustubh-karkare: This part is duplicated, hmm.

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,132 @@
...
+    assert (resp_code == expected_resp_code), 'Expected %s HTTP response code, got %s.\nResponse Status Message: %s\n' % (expected_resp_code, resp_code, resp_status_msg)
+    assert (http_version == 'HTTP/1.1')
+    parsed_headers = dict()
+    for header_line in headers:
+        header, data = header_line.split(': ')
+        parsed_headers[header] = data
+    return parsed_headers
+
+def parse_request(request_data):
+    headers = request_data.strip().split('\r\n')
```
kaustubh-karkare: Since this '\r\n' appears multiple times in your code, you should define it once as a constant at the top and reuse it whenever possible. Having strings like this inside of code is generally considered bad practice.

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,132 @@
...
+    '''
+    Calculates ranges to download the resource according to
+    number of parts specified
+    '''
+    if "Accept-Ranges" not in headers:
+        raise KeyError("Server does not accept range requests")
+
+    data_ranges = list()
+    start = 0
+    file_size = int(headers["Content-Length"])
```
kaustubh-karkare: Instead of passing in the entire headers object, you can try to limit the information to only what is exactlu necessary, right?
```
if not headers.get("Accept-Ranges"):
  compute_request_ranges(headers["Content-Length"], parts)
```

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,132 @@
...
+
+    regex_pattern -
+        non-capturing group 1 : http protocol
+        capturing group 1: host
+        non-capturing group 2: colon(:)
+        capturing group 2: port number
+        non-capturing group 3: slash(/) - unnecessary will rectify it later
+        capturing group 3: path to resource
+    '''
+    regex_pattern = '^(?:http:\/\/)([^:\/?]+)(?:[\:]?)([0-9]*)(?:\/?)(.*)$'
```
kaustubh-karkare: * Since you're already validating the protocol below (with a better error message), you can relax the constraint here: instead of "http" in the regex, you could just use "\w+".
* if the port number is specified, then the colon has to be specified, right? So `(?:[\:]([0-9]*))?`

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,135 @@
...
+                    copyfileobj(temp_file, main_file)
+                os.remove(download_file)
+
+    def _download_multi(self):
+        threads = list()
+        download_objs = dict()
+        for index in range(len(self.data_ranges)):
+            filename = self.save_as + ("%d" % (index))
+            download_objs[filename] = DownloadHandler(self.host, self.port, self.resource, filename, self.logger, self.data_ranges[index])
+            thread_obj = threading.Thread(target=download_objs[filename].download())
```
kaustubh-karkare: I don't think you're supposed to call the function here. Just give it to the Thread constructor, and that will call it.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def _analyze_header_data(self, header_data):
+        self.logger.info("Analyzing header data...")
+        headers = utility.parse_response(header_data, "200")
+        self.data_ranges = utility.compute_request_ranges(headers, self.threads)
+
+        self.logger.debug("Byte ranges created: %s" % (self.data_ranges))
+        self.logger.info("Script ready to download!")
+
+    def _collect_header_data(self):
```
kaustubh-karkare: Should use your DownloadHandler for this too.

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,132 @@
...
+
+def get_headers(sock):
+    '''
+    Receive header data from socket
+    '''
+    response_msg = ''
+    msg_buffer = sock.recv(1)
+    response_msg += msg_buffer.decode()
+    while response_msg[-4:] != '\r\n\r\n':
+        msg_buffer = sock.recv(1)
```
kaustubh-karkare: Loading one character at a time is pretty inefficient. You can create a wrapper object over the socket, which reads an entire block, but only gives you what you want, saving the rest for later use.

solutions/assignment-3/file_downloader_cli.py
```diff
@@ -0,0 +1,27 @@
...
+    else:
+        log_level = logging.WARNING
+
+    logger = logging.getLogger('file_downloader_logger')
+    logger.setLevel(log_level)
+    console_handler = logging.StreamHandler()
+    console_handler.setLevel(log_level)
+    formatter = logging.Formatter('%(asctime)s - %(message)s')
+    console_handler.setFormatter(formatter)
+    logger.addHandler(console_handler)
```
kaustubh-karkare, line 24: This looks familiar :)

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,173 @@
...
+    :param port: port number to host server as
+    :param logger: logger object used for debugging and other running information
+    :param server_location: system path to run server at
+    :param speed_limit_bps: speed limit in bytes per second to send data
+    """
+
+    def __init__(self, interface, port, logger, server_location, speed_limit_bps):
+        self.interface = interface
+        self.port = port
+        self.homepage = 'http://' + interface + ':' + str(port)
```
kaustubh-karkare: Looks like this is only used for logging, and so doesn't really need to be class property.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,173 @@
...
+        self.server_location = server_location
+        os.chdir(self.server_location)
```
kaustubh-karkare: Try to avoid using this function in general. Provide server location directly to subsequent methods that rely on it.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def start(self):
+        """
+        Create socket instance and bind ip address and host
+        """
+        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        self.sock.bind((self.interface, self.port))
+        self.logger.info("Server started")
+        self.run = True
```
kaustubh-karkare: `self.is_running` is a better name, since `run` could mean "start the run now"

solutions/assignment-3/server.py
kaustubh-karkare: ^ Still pending :)

So this is what we call a "blocking api". Ie - start does not return until stop is called.
In contrast, `select.select` makes you socket reads non-blocking. Ie - they return immediately.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,173 @@
...
+        self.sock.listen(10)
+        read_list = [self.sock]
+
+        while self.run:
+            read_ready, write_ready, execption_list = select.select(read_list, [], [])
+            for sock in read_ready:
+                if sock is self.sock:
+                    self.logger.info('Listening at %s' % (self.homepage))
+                    conn_obj, client_addr = self.sock.accept()
+                    read_list.append(conn_obj)
```
kaustubh-karkare: You only need the server socket to be non-blocking.
The client socket (ie - the conn_obj here) can immediately be handed off to its own thread, where it is okay to using blocking api.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,173 @@
...
+        self.conn_obj = conn_obj
+        self.server_info = server_info
+        self.logger = server_info['logger']
+
+    def run(self):
+        self.logger.info("Receiving headers")
+        request_received = utility.get_headers(self.conn_obj)
+        self.logger.debug("Headers received: \n%s" % (request_received))
+        self.req_method, self.resource, http_ver, self.recv_headers = utility.parse_request(request_received)
+        self._process_request()
```
kaustubh-karkare: By saving everything on the class, it is more difficult to figure out which method uses what data.
I would recommend trying to use local variables here as far ass possible.

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,212 @@
...
+
+        port = int(port) if port else 80 #assign default TCP port 80 if none assigned
+        resource = '/' if not resource else resource
+
+        return host, port, resource
+
+class HttpRequest(HttpProtocol):
+
+    def __init__(self, sock=None):
+        super(HttpRequest, self).__init__(sock)
```
kaustubh-karkare: * The default constructor does this too, so you don't need to specify this.
* Even if you did, you can do something like this:
```
def __init__(self, *args, **kwargs):
  super().__init__(*args, **kwargs)
  ...
```
Which basically means "I don't really care what arguments the parent takes, just forward what I have to it."

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,212 @@
...
+import socket
+from time import sleep
+
+class HttpProtocol:
+
+    CRLF = '\r\n'
+
+    def __init__(self, sock):
+        if sock:
+            self.sock = sock
```
kaustubh-karkare: Don't even store this here, since it is now entirely SocketBuffer's responsibility.

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,212 @@
...
+            self.sb = SocketBuffer(sock)
+
+    def parse_headers(self, headers_data):
```
kaustubh-karkare: https://docs.python.org/3/library/typing.html
Can you also start adding types for all function arguments?
```
def parse_headers(self, headers_data: str) -> Dict[str, str]:
```
Raise an exception if you cant parse something, don't be forgiving :)

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,212 @@
...
+
+    WAIT_TIME = 10 #in seconds
+    BLOCK_SIZE = 2048 #in bytes
+
+    def __init__(self, sock):
+        self.sock = sock
+        self.data_recv = b''
+        self.sock.settimeout(self.WAIT_TIME)
+
+    def recv(self, length=0):
```
kaustubh-karkare: This wrapper doesn't seem to be that useful, tbh.
```
headers = {}
while True:
  line = sb.readLine()
  if line == "": break
  key, _, value = line.partition(": ")
  heades[key] = value
```

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,212 @@
...
+    def parse_request(self, request_data):
+        parsed_headers, unparsed_headers = self.parse_headers(request_data)
+        for line in unparsed_headers:
+            try:
+                req_method, resource,  http_version = unparsed_headers.pop().split()
+            except ValueError:
+                pass #need to handle this error, todo later
+        parsed_headers['req_method'] = req_method
+        parsed_headers['resource'] = resource
+        parsed_headers['http-version'] = http_version
```
kaustubh-karkare: Use constants in place of these strings too, so that you can access via the same strings.

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,212 @@
...
+    def generate_request(self, req_method, resource, server_addr, data_range=None):
+        request_line = "%s /%s HTTP/1.1%s" % (req_method, resource, self.CRLF)
+
+        if data_range is not None:
+            data_range = "bytes=" + data_range
+        header_dict = {
+            'User-Agent': 'Python Script',
+            'Host': '%s' % (server_addr),
+            'Accept': '*/*',
+            'Range': data_range
```
kaustubh-karkare, line 84: Same as previous comment.
```
class HttpHeaders(Enum):
  USER_AGENT = 'User-Agent'
  ...
```

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,212 @@
...
+        resource = '/' if not resource else resource
+
+        return host, port, resource
+
+class HttpRequest(HttpProtocol):
+
+    def __init__(self, sock=None):
+        super(HttpRequest, self).__init__(sock)
+
+    def parse_request(self, request_data):
```
kaustubh-karkare: Instead of getting request_data as an argument, you should just read it directly from `self.sb` since that is possible here.

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,212 @@
...
+        if data_range is not None:
+            data_range = "bytes=" + data_range
+        header_dict = {
+            'User-Agent': 'Python Script',
+            'Host': '%s' % (server_addr),
+            'Accept': '*/*',
+            'Range': data_range
+        }
+        request = (request_line + "".join("%s: %s%s" % (key, value, self.CRLF) for key, value in header_dict.items() if value is not None) + self.CRLF)
+        return request
```
kaustubh-karkare, line 87: Instead of returning, just write it to self.sb directory.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+    def _analyze_header_data(self):
+        download_obj = DownloadHandler(self.host, self.port, self.resource, self.logger)
+        header_data = download_obj.headers_request()
+        self.logger.info("Analyzing header data...")
+        temp_response_util = utility.HttpResponse()
+        headers = temp_response_util.parse_response(header_data)
+        assert (headers['resp_code'] == '200')
+        self.data_ranges = utility.HttpProtocol.compute_request_ranges(int(headers['Content-Length']), self.threads)
```
kaustubh-karkare: The logic in compute_request_ranges is actually not protocol specific.
It is only relevant to this module responsible for parallel downloads.
You should move it here.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,139 @@
...
+
+    :param url: url of resource to download
+    :param save_as: file name to save download as
+    :param logger: logger object used for debugging and other running information
+    :param threads: number of parts to download specified resource in
+    """
+
+    def __init__(self, url, save_as, logger, threads=1):
+        self.url = url
+        self.save_as = save_as
```
kaustubh-karkare: I guess this is more accurate: s/save_as/download_path/

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,139 @@
...
+        threads and downloads file part sizes and then stitches all the files into one
+        main file.
+        """
+        if self.threads > 1:
+            self._analyze_header_data()
+            self.logger.info("Download started")
+            start_time = time.time()
+            self._download_multi()
+            time_taken = time.time() - start_time
+            self._stitch_downloads()
```
kaustubh-karkare: Interesting, this is readable, but my concern here is that it isn't clear what data is being output from one thing and input to the other. Eg - the following is a better alternative:
```
file_size = self._get_file_size(self.host, self.port, self.resource)
download_ranges = self._compute_ranges(file_size, self.threads)
handlers = []
with ThreadPoolExecutor(max_workers=self.threads):
    for download_range in download_ranges:
        handler = DownloadHandler(..., download_range)
        handlers.append(handler)
        executor.submit(handler.download)
# the with-block is only completed when all downloads are complete
self._stitch_downloads(self.download_path, [handler.download_path for handler in handlers])
```
Basically, I'm trying to minimize class properties and using local variables whenever possible, unless there is no choice. In theory, I could even pass `self.logger`, but that might be too annoying.
This way, the output of the function is purely dependent on the input. https://en.wikipedia.org/wiki/Pure_function

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,139 @@
...
+    def headers_request(self):
+        self._send("HEAD")
+        headers = self.request_util.get_headers()
+        self.sock.close()
+        return headers
+
+    def _send(self, req_method):
+        request = self.request_util.generate_request(req_method, self.resource, self.host, self.data_ranges)
+        self.logger.debug("Request generated:\n%s" % (request))
+        self.sock.sendall(request.encode())
```
kaustubh-karkare, line 139: If you're using SocketBuffer, use it consistently. All reads/writes should go through that.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,139 @@
...
+        self.port = port
+        self.resource = resource
+        self.save_as = save_as
+        self.logger = logger
+        self.data_ranges = data_ranges
+
+        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.sock.connect((self.host, self.port))
+        self.request_util = utility.HttpRequest(self.sock)
+        self.response_util = utility.HttpResponse(self.sock)
```
kaustubh-karkare: I think this creates 2 copies of the SocketBuffer object, right?
```
buffer = utility.SocketBuffer(self.socket)
self.request = utility.HttpRequest(buffer)
self.response = utility.HttpResponse(buffer)
```

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,212 @@
...
+                unparsed_headers.append(header_line)
+        return parsed_headers, unparsed_headers
+
+
+    def download_payload(self, save_as, content_length):
+        download_length = 0
+
+        with open(save_as, 'wb') as download_file:
+            while download_length < content_length:
+                self.sb.recv()
```
kaustubh-karkare: Consistently read using BLOCK_SIZE

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,180 @@
...
+        self.logger.info("Exiting")
+        self.is_running = False
+        self.sock.shutdown(socket.SHUT_RDWR)
+        self.sock.close()
+
+    def _handle_request(self):
+        """
+        Handles incoming request in non-blocking way
+        """
+        self.sock.listen(10)
```
kaustubh-karkare, line 55: Use kwargs whenever possible since it helps with readability.
```
self.socket.listen(backlog=10)
```
I original thought this was a timeout.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,180 @@
...
+        homepage = 'http://' + self.interface + ':' + str(self.port)
+        read_list = [self.sock]
```
kaustubh-karkare, line 57: Don't need this variable anymore.

solutions/assignment-3/unittest_socket.py
```diff
@@ -0,0 +1,92 @@
...
+        with open(file_name, 'wb') as dummy_file:
+            dummy_file.write(os.urandom(file_size))
+        return os.stat(file_name).st_size
+
+
+    def test_simple_download(self):
+         URL = 'http://olps.cgtransport.org/Esewa/img/ESEWA2.jpg'
+         resource_size = 24963
+
+         with tempfile.TemporaryDirectory() as TempDir:
```
kaustubh-karkare: Naming convention: s/TempDir/temp_dir/

solutions/assignment-3/unittest_socket.py
```diff
@@ -0,0 +1,92 @@
...
+             os.chdir(TempDir)
```
kaustubh-karkare: Lets avoid using `os.chdir` in general.
Give your FileDownloader class an absolute path (temp_dir + file_name), and it can figure out the directory from that.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def _receive(self):
+        header_data = self.request_util.get_headers()
+        self.logger.debug("Header received:\n%s" % (header_data))
+        # expect 206 Partial Content response if range request issued otherwise expect normal response 200
+        expected_resp_code = "206" if self.data_ranges is not None else "200"
+        self.received_headers = self.response_util.parse_response(header_data)
+        assert (self.received_headers['resp_code'] == expected_resp_code)
+        assert (self.received_headers['http-version'] == 'HTTP/1.1')
+        self.response_util.download_payload(self.save_as, int(self.received_headers['Content-Length']))
```
kaustubh-karkare:
```
socket_buffer.download(to=self.download_location, size=...)
```
and in the server,
```
socket_buffer.upload(from=<whatever>, speed_bps=1024)
```

solutions/assignment-3/unittest_socket.py
```diff
@@ -0,0 +1,92 @@
...
+            server.stop()
+            self.assertAlmostEqual(os.stat("download_file").st_size, file_size, delta=10)
+        os.chdir(self.BASE_ADDR)
+
+    def test_non_blocking_server(self):
+
+        FILE_NAME = 'temp_file'
+        FILE_SIZE = 4 * 1000 * 1000
+        SPEED_LIMIT = 256 * 1000 * 1
+        URL = 'http://0.0.0.0:8000/' + FILE_NAME
```
kaustubh-karkare, line 102:
```
HOST = "0.0.0.0"
PORT = 8000
URL = f"http://{HOST}:{PORT}/{FILE_NAME}"
```
This way, you can use constants below on line 83.

solutions/assignment-3/unittest_socket.py
```diff
@@ -0,0 +1,92 @@
...
+
+        with tempfile.TemporaryDirectory() as TempDir:
+            os.chdir(TempDir)
+            logger = self.create_logger('test_non_blocking_server')
+            file_size = self.create_dummy_file(FILE_SIZE, FILE_NAME)
+            server = FileServer('0.0.0.0', 8000, logger, TempDir, SPEED_LIMIT)
+            server.start()
```
kaustubh-karkare: You can do all this in the setUp method, so that every test will have access to the server.
You can also specify port 0 initially, so that the OS will pick a random one, and construct the URL later, once an actual port has been assigned: https://stackoverflow.com/a/1365284/903585

Screen Shot 2020-02-14 at 7 33 58 PM

See the "address already in use errors" that happen otherwise.

kaustubh-karkare (changes requested):
This is a fairly complex implementation (which is good), which means I ended up making a lot of comments. They're all suggestions that you can follow if you wish (based on the code reviews I see at work), but if you're getting bored of this problem, we can skip them, since they're basically code quality improvements.
The only thing before I can accept is to get the unit test that verifies both the server and client behavior (ie - assertEqual, not assertAlmostEqual) to work reliable.

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,180 @@
...
+        self.logger.info("Server started")
+        self.is_running = True
+
+        self._handle_request_thread = threading.Thread(target=self._handle_request)
+        self._handle_request_thread.start()
+
+    def stop(self, *args):
+        self.logger.info("Exiting")
+        self.is_running = False
+        self.sock.shutdown(socket.SHUT_RDWR)
```
kaustubh-karkare, line 48: While running your code locally, I get errors on this line. Do you see them too?

solutions/assignment-3/server.py
```diff
@@ -0,0 +1,180 @@
...
+            else:
+                self.logger.info("Generating directory listing and sending")
+                upload_file = io.BytesIO()
+                list_page = self._generate_index_of('/' +  resource, recv_headers)
+                content_length = len(list_page)
+                upload_file.write(list_page.encode())
+
+        elif os.path.isfile(path):
+            if os.path.exists(path):
+                upload_file = open(path, 'rb')
```
kaustubh-karkare, line 138: Need to close the file at the end of the method.

solutions/assignment-3/file_downloader.py
```diff
@@ -0,0 +1,139 @@
...
+        for index in range(len(self.data_ranges)):
+            filename = self.save_as + ("%d" % (index))
+            download_objs[filename] = DownloadHandler(self.host, self.port, self.resource, self.logger, data_ranges=self.data_ranges[index], save_as=filename)
+            thread_obj = threading.Thread(target=download_objs[filename].download)
+            threads.append(thread_obj)
+            thread_obj.start()
+        for thread in threads:
+            thread.join()
+
+        self.download_files = download_objs.keys()
```
kaustubh-karkare: BUG! The keys of the dictionary do not have any implicit order, so your different parts could be stitched together in the wrong order.

solutions/assignment-3/utility.py
```diff
@@ -0,0 +1,212 @@
...
+                response_msg += msg_buffer
+                if response_msg[-4:] == '\r\n\r\n':
+                    header_recv = True
+                    break
+        return response_msg
+
+    def send_data_chunks(self, chunk_size, file_handler, seek_start, seek_end):
+        file_handler.seek(seek_start)
+        while file_handler.tell() <= seek_end:
+            data_chunk = file_handler.read(chunk_size)
```
kaustubh-karkare: BUG! This was the line causing problems for you.
```
data_chunk = file_handler.read(min(chunk_size, seek_end - file_handler.tell()))
```
Otherwise you read and return more data than you actually intend to.

solutions/assignment-3/server.py
drs-11: The file was closed in send_data_chunks method. Should I close it in this method itself?

solutions/assignment-3/utility.py
drs-11: Hmm then what should I do with the socket reference I pass to this class? I would need it.

solutions/assignment-3/utility.py
drs-11: Understood, will do.

solutions/assignment-3/utility.py
drs-11: Oh I should create the SocketBuffer instance here shouldn't I?

